PR #83, opened 2022-05-26
Add convenience functions for AppInstance and PluginInstance creation

eob:
In general, we're working to prune the Python Client down to the bare minimum surface area, which I think is fantastic.

But as we begin telling the Steamship story to more people, I do think that there are a few "WOW!" lines of code that will help us **show, rather than tell** the big idea.

One of them is the ability to say:

```python
app = Steamship.app("zendesk-tagger", "my-support-department")
app.load_tickets()
```

Getting to something like the above code would require three things:

1) Automatic loading of steamship credentials
2) A "quick load-or-create AppInstance" method
3) Dynamically creating functions on the AppInstance based on some kind of endpoint manifest

This PR does just step 2: it adds a quick-load function so that we can write:

```python
steamship = Steamship()
app = steamship.app("zendesk-tagger", "my-support-department")
app.post("load_tickets")
```

I realize this is all superfluous from a technical perspective. This PR adds no new functionality.

But that one-liner:

```
app = steamship.app("zendesk-tagger", "my-support-department")
```

I really think is attractive from a narrative perspective.

Thoughts?


EniasCailliau (dismissed):
This level of abstraction makes a lot of sense IMO. The less lines we need to create & run an app, the better. Great thinking!

dkolas (dismissed):
I think it's a good idea to streamline the up-front code as much as possible.  See top comment below - I'd much rather change the name of the verb we're using though.

To your larger points about the ultimate streamlining of the setup process, I'm very on board with #1 and #2, but am not a fan of #3 at all.  I do not like dynamic method names in APIs, to me it makes it much less clear what's part of the API and what's something specific to what you're doing (in this case, the remote app methods).  If other people love it, I'll hold my tongue, but it's not my preference.

src/steamship/client/client.py
```diff
@@ -141,6 +144,30 @@ def tag(
             space=space,
         )
 
+    def app(
```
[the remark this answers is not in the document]
dkolas: I'm happier with both `use` or `import` than `app` for this.

eob:
Update:

Based on the conversation above (and realization we already had something similar for plugins!), I:

- Updated the Engine to support creating AppInstance objects with App handles alone (rather than IDs),
- Implemented it as `Steamship.use`
- Added an additional `Steamship.use_plugin` for loading plugins

## Why `use` rather than `use_app`?

Since we want apps to be the main point of consumption, it felt OK to have it be `use` rather than `use_app`. If this takes off, we'll be seeing a lot of `Steamship.use` so concision is nice.

## Why add a plugin helper at all?

It seems like we still use plugins enough that it feels nice to be able to say:

```
   tagger = Steamship.use_plugin('oneai-tagger', 'default', ONEAI_CONFIG)
```

Rather than the longer `PluginInstance.create(...)` block.

src/steamship/client/client.py
```diff
@@ -156,7 +198,7 @@ def list_comments(
         external_id: str = None,
         external_type: str = None,
         external_group: str = None,
-    ) -> Response[TaskCommentList]:
+    ) -> IResponse[TaskCommentList]:
```
eob, line 201: @EniasCailliau I'm not sure if my branch is out of sync, but Pycharm was complaining here so I altered the response type on this

EniasCailliau: True. I never fix these since we have so many MyPy issues in the codebase. Holding off until we start "the big refactor" once we plan to go public.

src/steamship/client/client.py
```diff
@@ -104,6 +106,46 @@ def embed_and_search(
...
+    def use(
+        self, app_handle: str, handle: str = None, config: Dict[str, Any] = None
+    ) -> AppInstance:
+        """Creates or loads an instance named `handle` of App named `app_handle`."""
+        instance = AppInstance.create(
+            self,
+            app_handle=app_handle,
+            handle=handle,
+            config=config,
+            upsert=True,  # TODO(ted): Rename `upsert` globally to something semantically appropriate
```
EniasCailliau, line 118: Up for debate: Should `upsert` be True or False by default? 

Always creating Apps might pollute the workspace of our customers. 

Side note: Lately I've been using the term `workspace` to remind me what a `space` means. To me it works well as a mental model. A workspace is a place where I can think around with apps, plugins, and data. If I want to start from scratch, I just walk away from my workspace and go to a new one. If not I have to clean my workspace. 

I'm also dreaming that the client would be the mediator for workspaces as a user. This would like as follows: 

```
client = Steamship() 
File.create(...) # Create a file in my default workspace
client.workspace("new-workspace-for-lazy-enias") 
File.create(...) # This creates a file in my new workspace  
```

src/steamship/client/client.py
```diff
@@ -104,6 +106,46 @@ def embed_and_search(
             space=space,
         )
 
+    def use(
+        self, app_handle: str, handle: str = None, config: Dict[str, Any] = None
```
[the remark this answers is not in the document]
EniasCailliau: I would also add the optional version parameter.

src/steamship/data/app_instance.py
```diff
@@ -12,7 +12,9 @@
 class CreateAppInstanceRequest(Request):
     id: str = None
     app_id: str = None
-    appVersionId: str = None
```
EniasCailliau, line 15: Oh nooo! Still some camelCase in our precious codebase. 😱 Good catch!

tests/tests/app/integration/test_configurable_app_instance.py
```diff
@@ -34,3 +34,21 @@ def test_configurable_instance_invoke():
 
         res2 = instance2.post("greet").data
         assert res2 == f"{greeting2}, Person"
+
+        # Test with quick-create
+        greeting3 = "Howdy"
+        instance_config3 = {"greeting": greeting3}
+        instance3 = client.use(app.handle, f"{instance2.handle}-2", config=instance_config3)
```
EniasCailliau, line 41: Great improvement. Can't wait for devs to use this!

src/steamship/client/client.py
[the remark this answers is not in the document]
EniasCailliau: I'm thinking if we can come up with a better naming convention for `app_handle` vs `handle`. Just reading them requires me to read the documentation, which is something we want to avoid. Can't think of a better way right now, will come back to this.

EniasCailliau (commented):
Awesome PR! Can't wait to have a repository full of useful apps people rave about

EniasCailliau (approved):

src/steamship/client/client.py
eob: 👍

src/steamship/client/client.py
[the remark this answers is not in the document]
eob: 1. `reuse` I think is a great general re-phrasing of the (misnamed) `upsert` across the codebase. I think it might make sense to do that as a separate PR in which we globally rename that variable? That will be a breaking change with respect to current plugin & app code.

2. Optional version param - agree. Will add to this PR.

3. app_handle vs handle -- I don't like that pairing either. Let's plan for a future PR, but what do you think of the following proposal:

Option 1: "Allow anything":

   * `app` is always `Union[str, App]`. If `App` we transform it to `app.id`. If `str` it can be either `handle or id`. On the engine-side, we use regex to guess if it's UUID-like and then either do the lookup against `id` or `handle` in the DB.

Option 2: "Always handle":

   * Systems like NPM, Github only ever use the handle in their API. From the customer's perspective that's all that matters. We could adopt a similar posture in which the ID field is basically an internal piece of state; the customer only ever needs to remember the human readable name.

src/steamship/client/client.py
```diff
@@ -104,6 +106,46 @@ def embed_and_search(
...
+            self,
+            app_handle=app_handle,
+            handle=handle,
+            config=config,
+            upsert=True,  # TODO(ted): Rename `upsert` globally to something semantically appropriate
+        )
+
+        if instance.error:
+            raise instance.error
+        if not instance.data:
```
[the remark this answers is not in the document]
eob: Ah, yes! Will change.

src/steamship/client/client.py
eob: Yeah; I think we're due for a round of naming experimentation before we launch.

Similarly, I've started really liking `package` instead of `app` in the way I think/talk about things.

Regarding `client.workspace(...)` I agree that some notion of a global client might be really cool. Taking that idea further, it would support:

```
File.create(...)
```

Without **any** `client = Steamship()` if we wanted the "what to do if `client=None` to mean `client=get_global_client()` which itself noticed the lack of one and created it...

src/steamship/client/client.py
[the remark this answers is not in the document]
eob: (just checked)  -- in the current definition it seems like if `response.status` is None and `response.data` is None, it will happily return `None`. That might be unlikely, but this final check will make sure it will truly never return `None` for this particular function.

I think permitting response to be `None` friendly is actually a feature, not a bug -- so leaving that check here rather than pushing it down into the `.data` accessorΩ

PR #90, opened 2022-06-01
Move .tag function from Client to PluginInstance

EniasCailliau:
@eob As discussed, moving `.tag` function to PluginInstance which: 
1. Feels more natural 
2. Is consistent with the `get_training_param` and `train` calls


dkolas:
Is the intent to move other plugin operations to PluginInstance also?

src/steamship/data/plugin_instance.py
```diff
@@ -19,6 +19,10 @@
 from steamship.plugin.outputs.train_plugin_output import TrainPluginOutput
 from steamship.plugin.outputs.training_parameter_plugin_output import TrainingParameterPluginOutput
 
+from .block import Block
+from .file import File
+from .operations.tagger import TagRequest, TagResponse
+
```
dkolas, line 25: Is this style of import preferred?

PR #91, opened 2022-06-03
Remove tasks from client

EniasCailliau:


EniasCailliau:
@dkolas Good point. Not sure if this was already supported by our code though. 

The current `tasks` property was used as a meta object to list comments through a `list_comments` method. Besides that, the `tasks` object did not have any use. Hence why I chose to remove the level of abstraction.  

Sounds like you would like the client to offer a `list_tasks()` method. Does the engine support this through its API or would the client need to track the tasks?

src/steamship/client/client.py
```diff
@@ -106,6 +104,28 @@ def embed_and_search(
             space=space,
         )
 
+    def tag(
```
eob, line 107: I think this `tag` addition snuck in from not having rebased/merged against `main`?

src/steamship/client/client.py
```diff
@@ -129,3 +149,18 @@ def get_space(self) -> Space:
             )
         logging.info(f"Got space: {space.data.id}")
         return space.data
+
+    def list_comments(
```
eob, line 153: Since `comments` are a sub-item of `Task`, and this requires a `task_id`, wouldn't it make sense to put this on the `Task` object?

eob:
To the comment above, there is a `Task.list` -- https://github.com/nludb/python-client/blob/4759f03c66966034a14fa1bc9759b2875c533c89/src/steamship/base/tasks.py#L80

eob (changes requested):
TaskComment isn't being used anymore so might make sense to chat about our goals for it before making any changes.

That could be anything from removing it, to doing this here, to folding it into the Task object, etc.

FWIW it was used as an easy way for the HelpScout prototype to record ratings of our prediction performance. E.g. Steamship would predict a helpdesk article, that would get shown to the UI, and the UI would use the taskId on the result object to issue comments from the users (thumbs up, thumbs down, notes) that would get aggregated into a report about performance.

eob (approved):
LGTM.

I think this is probably a waypoint toward some eventual refactoring of TaskComments anyway, but glad to be rid of that extra unnecessary class in the interim.

PR #93, opened 2022-06-06
Allow explicit False values in plugin config

dkolas:


dkolas:
Merging on slack comments

PR #94, opened 2022-06-06
Addl training fixes

dkolas:
- Pass space when running PluginInstance.train
- Provide pydantic alias for TrainPluginInput.training_data_url
- Bring back TrainingParameterPluginOutput.from_input that was mistakenly deleted in previous pr


src/steamship/plugin/outputs/training_parameter_plugin_output.py
```diff
@@ -17,6 +18,17 @@ class TrainingParameterPluginOutput(CamelModel):
     inference_params: Dict[str, Any] = None
     export_request: ExportPluginInput = None
 
+    @staticmethod
+    def from_input(input: TrainingParameterPluginInput) -> TrainingParameterPluginOutput:
```
EniasCailliau, line 22: This can also be done using `TrainingParameterPluginOutput.parse_obj(input.__dict__)`. The verbosity helps in this case so I leave it up to you to decide what feels best.

EniasCailliau (approved):

src/steamship/plugin/outputs/training_parameter_plugin_output.py
dkolas: Thanks for the pointer!  I think I'll leave it in this case, because there may be specific transformations here.

PR #95, opened 2022-06-08
Raise exception when data is accessed on an incomplete asynchronous request

EniasCailliau:


dkolas (approved):
Thank you for this!!

src/steamship/base/response.py
```diff
@@ -28,6 +28,15 @@ class Config:
     def data(self):
         if self.error:
             raise self.error
+        if self.task is not None:
+            if self.task.state == TaskState.failed:
+                raise SteamshipError(
```
eob, line 33: here we could also check `task.statusMessage` which likely includes the reason for the failure. I believe there is also a `task.statusSuggestion` which might contain a possible remediation

eob (approved):
LGTM! Comment within on how to further improve the error thrown (there's a message from the server you can include)

PR #97, opened 2022-06-11
Utility library for remapping the output of third party services to Steamship model objects

eob:
Helper classes for mapping third-party API response objects into Steamship model objects.

Many third-party APIs produce data that is isomorphic, but differently structured, than Steamship plugin responses.
These utility classes reduce the required conversion to a declarative specification, enabling faster, simpler, and more
error-robust construction of Steamship Plugins which wrap third-party services.

For example, to reshape the object:

```
       {
          "entities": [
             {
               "entity": {
                  "entityTag": "ORG",
                  "start": 1,
                  "end": 5
               }
            }
          ]
      }
```

    Into an array containing the Steamship Tag.CreateRequest objects, one can write:

```
       reshape_array_of_dicts(
         input_dict: input_dict,
         array_keypath: ["entities"],
         array_required: True,
         mappings: {
            "kind": Mapping(const="NER"),
            "name": Mapping(keypath=["entity", "entityTag"]),
            "start_idx": Mapping(keypath=["entity", "start"]),
            "end_idx": Mapping(keypath=["entity", "start"])
         },
         into_base_model: Tag.CreateRequest
       )
```

    Producing the following output:

```
       [
         Tag.CreateRequest(
            kind="NER",
            name="ORG",
            start_idx=1,
            end_idx=5
         )
       ]
```


src/steamship/utils/dict_mapper.py
```diff
@@ -0,0 +1,179 @@
...
+            if key not in pointer:
+                if required:
+                    raise SteamshipError(
+                        message=f"[get_keypath] After traversing keypath {traversed_keypath} in object, unable to find expected key {key}."
+                    )
+                return None
+            traversed_keypath.append(key)
+            pointer = pointer.get(key, None)
+        elif isinstance(pointer, list):
+            if key < 0 or key >= len(pointer):
```
EniasCailliau, line 48:
```suggestion
            if not 0 <= key < len(pointer):
```
Just fancy Python

src/steamship/utils/dict_mapper.py
```diff
@@ -0,0 +1,179 @@
...
+
+        if isinstance(pointer, dict):
+            if key not in pointer:
+                if required:
+                    raise SteamshipError(
+                        message=f"[get_keypath] After traversing keypath {traversed_keypath} in object, unable to find expected key {key}."
+                    )
+                return None
+            traversed_keypath.append(key)
+            pointer = pointer.get(key, None)
```
EniasCailliau, line 46:
```suggestion
            pointer = pointer[key]
```

We can use pointer[key] here as we checked if key is present before.

src/steamship/utils/dict_mapper.py
```diff
@@ -0,0 +1,179 @@
...
+def reshape_dict(
+    input_dict: Dict,
+    mappings: Dict[str, Mapping],
+    into_base_model: Optional[Type[BaseModel]] = None,
+) -> Union[Dict, BaseModel]:
+    """Produces a new, 1-level deep, dictionary or BaseModel from a set of mappings against `input_dict`
+
+    - If `into_pydantic` is not None, the resulting dictionary will be parsed into the provided Pydantic class.
+    """
+    ret = {key: mappings[key].resolve_against(input_dict) for key in mappings}
```
EniasCailliau, line 115:
```suggestion
    ret = {key: mapping.resolve_against(input_dict) for key, mapping in mappings.items()}
```

src/steamship/utils/dict_mapper.py
```diff
@@ -0,0 +1,179 @@
...
+
+    if into_base_model is not None:
+        return into_base_model(**ret)
+    return ret
```
EniasCailliau, line 119:
```suggestion
    return ret if into_base_model is None else into_base_model(**ret)
```

EniasCailliau (changes requested):
Love the functionality! Discovery of these tools will be key for our users.

Had a few questions regarding the use of the `Mapping` definition. Looking forward to your thoughts.

dkolas (changes requested):
I understand the spirit of what you're after here, because this bit is really the only thing that has to be reimplemented per HF plugin.  However:

1 - Is this declarative way actually better than just writing the few lines of python? Below is the relevant method in the zero shot classifier plugin:
```
 def make_tags_from_response(self, response, block_len) -> List[Tag.CreateRequest]:
        tags = []
        for i, label in enumerate(response['labels']):
            tags.append(Tag.CreateRequest(kind=self.config.tag_kind,
                                          name= label,
                                          startIdx=0,
                                          endIdx=block_len,
                                          value={'score': response['scores'][i]}))
        return tags
```
It's shorter than your example, and there's nothing new to learn.

2 - Should we be trying to add this right now?  It feels like a new (albeit small) layer atop the data model that I'm not sure we can justify quite yet.

eob:
Hrm - yeah, @dkolas I think you might be right. 

The thing that spawned this was that the output of some third-party taggers like OneAI & NLPCloud have a slightly different structure PER model - and each has >10 models. And collapsing the imperative code into declarative code made it clearer what was the same, and what was different, between the responses....

But at the end of the day if it's the same amount of code required.. perhaps it's just needless indirection. Will close this PR without merge and have a go at implementing manually to see how it feels

PR #98, opened 2022-06-13
Fix bug in huggingface when optional additional_parameters is None

EniasCailliau:
Fixed a tiny bug that occurred when the optional `additional_params` was None. 

Also added default list values for our data model fields so that we can avoid: 

```
if file.blocks is None: 
    file.blocks = []
```

### DIFF
--- a/src/steamship/utils/huggingface_helper.py
+++ b/src/steamship/utils/huggingface_helper.py
@@ -4,7 +4,6 @@ It uses asyncio parallelism to make many http requests simultaneously.
 """
 
 import asyncio
-import json
 import logging
 import time
 from http import HTTPStatus
@@ -17,8 +16,9 @@ from steamship import Block, SteamshipError
 
 
 async def _model_call(session, text: str, api_url, headers, additional_params: dict = None) -> list:
-    json_input = dict(inputs=text, wait_for_model=False, parameters=additional_params)
-    data = json.dumps(json_input)
+    additional_params = additional_params or {}
+    json_input = {"inputs": text, "wait_for_model": True, "parameters": additional_params}
+    ok_response, nok_response = None, None
 
     max_error_retries = 3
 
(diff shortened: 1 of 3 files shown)


src/steamship/utils/huggingface_helper.py
```diff
@@ -17,8 +16,9 @@
 
 
 async def _model_call(session, text: str, api_url, headers, additional_params: dict = None) -> list:
-    json_input = dict(inputs=text, wait_for_model=False, parameters=additional_params)
-    data = json.dumps(json_input)
+    additional_params = additional_params or {}
+    json_input = {"inputs": text, "wait_for_model": True, "parameters": additional_params}
```
dkolas, line 20: Sorry I missed this!

src/steamship/utils/huggingface_helper.py
```diff
@@ -27,38 +27,34 @@ async def _model_call(session, text: str, api_url, headers, additional_params: d
...
 async def _model_calls(
-    texts: List[str],
-    api_url: str,
-    headers,
-    timeout_seconds: int,
-    additional_params: dict = None,
+    texts: List[str], api_url: str, headers, additional_params: dict = None
 ) -> List[list]:
-    async with aiohttp.ClientSession(timeout=ClientTimeout(total=timeout_seconds)) as session:
+    async with aiohttp.ClientSession(timeout=ClientTimeout(total=10)) as session:
```
dkolas, line 57: It looks like you accidentally undid the timeout_seconds change in a merge conflict.

src/steamship/utils/huggingface_helper.py
```diff
@@ -69,28 +65,17 @@ async def _model_calls(
...
     start_time = time.perf_counter()
     results = asyncio.run(
-        _model_calls(
-            [block.text for block in blocks],
-            api_url,
-            headers,
-            timeout_seconds=timeout_seconds,
-            additional_params=additional_params,
-        )
+        _model_calls([block.text for block in blocks], api_url, headers, additional_params)
```
dkolas, line 78: Same here for timeout seconds

dkolas (changes requested):
Thanks for fixing the bug - it looks like you accidentally removed the ability to pass timout_seconds.

PR #100, opened 2022-06-20
WIP: New inputs/ outputs for async training

dkolas:


eob (approved):
LGTM! At some point we might want to make the getTrainingParams http path snake case too

PR #101, opened 2022-06-23
Clean up requests and fix client and configuration

EniasCailliau:
Shallow refactor of the client to improve readability. I refactored as part of my mission to understand what is happening in the engine.  

- [x] Simplify client code that performs requests to the server
- [x] Fix tests


src/steamship/base/configuration.py
```diff
@@ -45,34 +50,23 @@ def __init__(
...
         # Then load configuration from a file if provided
         config_dict = self._load_from_file(
             config_file or DEFAULT_CONFIG_FILE,
-            self.profile,
+            profile,
             raise_on_exception=config_file is not None,
         )
         config_dict.update(self._get_config_dict_from_environment())
-        self.update(config_dict)
+        kwargs.update({k: v for k, v in config_dict.items() if kwargs.get(k) is None})
```
eob, line 63: And here again.

eob (dismissed):
This is a fantastic set of cleanups -- thank you!

I think (though may be wrong) that you might need to merge in `main` as it looked like some of the changes from @dkolas 's recent training PRs were being flagged as new code here. Either way, no need to re-submit for review before merging; I just wanted to flag for you!

src/steamship/data/file.py
```diff
@@ -18,7 +18,6 @@
 
 class FileUploadType(str, Enum):
     FILE = "file"  # The CreateRequest contains a file upload that should be used
-    URL = "url"  # The CreateRequest contains a url that should be scraped
```
dkolas, line 21: Did we remove this functionality?

src/steamship/data/plugin.py
```diff
@@ -170,7 +171,7 @@ def create(
         handle: str = None,
         training_platform: Optional[HostingType] = None,
         metadata: Union[str, Dict, List] = None,
-        upsert: bool = None,
+        upsert: bool = True,
```
dkolas, line 174: I think I'd prefer default to be False rather than True - not sure we would want to assume clobbering by default on a method called 'create'

src/steamship/data/plugin_instance.py
```diff
@@ -74,9 +74,14 @@ def create(
         plugin_version_id: str = None,
         plugin_version_handle: str = None,
         handle: str = None,
-        upsert: bool = None,
+        upsert: bool = True,
```
dkolas, line 77: Same comment as above with default = False

src/steamship/data/plugin_version.py
```diff
@@ -81,7 +81,7 @@ def create(
         plugin_id: str = None,
         filename: str = None,
         filebytes: bytes = None,
-        upsert: bool = None,
+        upsert: bool = False,
```
dkolas, line 84: I think False is the correct default here (and they should all be the same)

tests/assets/plugins/taggers/plugin_trainable_tagger.py
```diff
@@ -53,6 +53,9 @@ class TestTrainableTaggerModel(TrainableModel[EmptyConfig]):
       In this example model, the `from_disk(path: Path)` method provides this functionality.
     """
 
+    def train_status(self, input: TrainStatusPluginInput) -> TrainPluginOutput:
+        return TrainPluginOutput(training_complete=True)
+
```
dkolas, line 58: Thanks for this - sorry I merged the async training stuff without fixing the tests.

tests/tests/plugin/unit/test_trainable_tagger.py
```diff
@@ -68,7 +68,7 @@ def test_trainable_tagger():
         ),
         model,
     )
-    assert tagger1.data == TRAIN_RESPONSE.dict()
+    assert tagger1.data == TRAIN_RESPONSE.dict(by_alias=True)
```
dkolas, line 71: Sorry again

src/steamship/base/client.py
```diff
@@ -84,28 +79,13 @@ def _url(
...
-            if self.config and self.config.app_base:
-                base = self.config.app_base
-            if config and config.app_base:
-                base = config.app_base
-            if base is None:
-                return SteamshipError(
-                    code="EndpointMissing",
-                    message="Can not invoke an app endpoint without the App Base variable set.",
-                    suggestion="This should automatically have a good default setting. "
-                    "Reach out to our Steamship support.",
```
dkolas, line 97: Is this message not still relevant?

dkolas (changes requested):
Looks good to me, with a minor request about defaults.

Can we hold on merging this until at least tomorrow so that I can make sure I get the pilot app out the door safely?

PR #102, opened 2022-06-30
Support url-conveyed data in RawDataPluginInput

eob:
Some files are too large to comfortably send as Base64
      within JSON. The Steamship Engine sometimes chooses to send an empty `data` field paired with a non-empty
      `url` field. When this happens, the constructor proactively, synchronously fetches the contents of that `url`
      and assigns it to the `data` field, throwing a SteamshipError if the fetch fails. Again, this is done
      automatically so the Plugin Author can mostly ignore this fact.

### DIFF
--- a/tests/tests/plugin/unit/test_trainable_tagger.py
+++ b/tests/tests/plugin/unit/test_trainable_tagger.py
@@ -5,7 +5,6 @@ from assets.plugins.taggers.plugin_trainable_tagger import (
 )
 
 from steamship import File
-from steamship.base.utils import to_camel
 from steamship.data.block import Block
 from steamship.plugin.inputs.block_and_tag_plugin_input import BlockAndTagPluginInput
 from steamship.plugin.inputs.train_plugin_input import TrainPluginInput
@@ -69,7 +68,7 @@ def test_trainable_tagger():
         ),
         model,
     )
-    assert tagger1.data == {to_camel(k): v for (k, v) in TRAIN_RESPONSE.dict().items()}
+    assert tagger1.data == TRAIN_RESPONSE.dict(by_alias=True)
 
     tagger2 = plugin.train_endpoint(
         **PluginRequest(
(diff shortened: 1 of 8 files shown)


tests/tests/plugin/unit/test_trainable_tagger.py
```diff
@@ -68,7 +69,7 @@ def test_trainable_tagger():
         ),
         model,
     )
-    assert tagger1.data == TRAIN_RESPONSE.dict()
+    assert tagger1.data == {to_camel(k): v for (k, v) in TRAIN_RESPONSE.dict().items()}
```
eob, line 72: @EniasCailliau I don't think my change here should be necessary given my understanding of `CamelModel`... so I wanted to flag you to get your thoughts -- am I doing something wrong in the construction of TRAIN_RESPONSE?

tests/tests/plugin/unit/test_trainable_tagger.py
```diff
@@ -79,7 +80,7 @@ def test_trainable_tagger():
             plugin_instance_id="000",
         ).dict()
     )
-    assert tagger2.data == TRAIN_RESPONSE.dict()
+    assert tagger2.data == {to_camel(k): v for (k, v) in TRAIN_RESPONSE.dict().items()}
```
eob, line 83: @EniasCailliau  Ditto; see above comment

eob:
@EniasCailliau with the change to the Engine I plan to deploy before you wake up, this change ought to pass with all green checks against prod/staging.

Please feel free to merge in if it's useful to you to use larger input files!

tests/tests/plugin/unit/test_trainable_tagger.py
EniasCailliau: We need to enable `use_alias` while calling `dict()`: 

```
TRAIN_RESPONSE.dict(use_alias=True)
```

tests/tests/plugin/unit/test_trainable_tagger.py
EniasCailliau: We need to enable use_alias while calling dict():

TRAIN_RESPONSE.dict(use_alias=True)

EniasCailliau (dismissed):
Clean and simple! Thanks for this. Hopefully Joe rogan will be happy 👍

tests/tests/plugin/unit/test_trainable_tagger.py
eob: Thanks!

tests/tests/plugin/unit/test_trainable_tagger.py
eob: Note: it turns out that it's `by_alias` -- but this comment helped me quickly find that. Thanks!

PR #103, opened 2022-07-01
Unit test expected error responses of apps/plugins

eob:
This is the Python-side work of building end-to-end Exception visibility into our test suite.

It tests that

* Unexpected python Exceptions
* thrown SteamshipErrors
* missing method calls

Are all reported to the user as an appropriately formatted SteamshipResponse object.

## Testing method

The built-in testing app now contains a few functions  which are intended to generate errors with known messages. The test invokes those functions and assertions are placed on the HTTP response object.

## Additional Improvements

This PR includes a PyTest fixture which makes it very easy to test a Steamship App from the perspective of the HTTP caller without actually doing a full deploy & round trip to the engine.

It is useful to have some tests which actually do this full round trip, but aside from testing that the round trip works, we probably want to refactor most tests to use this fixture so that they can run FAR faster than present.

### DIFF
--- a/src/steamship/app/lambda_handler.py
+++ b/src/steamship/app/lambda_handler.py
@@ -56,7 +56,7 @@ def create_handler(app_cls: Type[App]):
             return Response.error(
                 code=HTTPStatus.INTERNAL_SERVER_ERROR,
                 prefix=error_prefix,
-                message=f"Unable to initialize plugin/app. {ex}",
+                message=f"Unable to initialize plugin/app.",
                 exception=ex,
             )
 
--- a/src/steamship/app/lambda_handler.py
+++ b/src/steamship/app/lambda_handler.py
@@ -46,6 +46,11 @@ def create_handler(app_cls: Type[App]):
                 exception=ex,
             )
 
+        if request and request.invocation:
+            error_prefix = f"[ERROR - {request.invocation.httpVerb} {request.invocation.appPath}] "
+        else:
+            error_prefix = f"[ERROR - ?VERB ?PATH] "
+
         try:
             app = app_cls(client=client, config=request.invocation.config)
         except SteamshipError as se:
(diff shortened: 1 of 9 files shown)


src/steamship/app/request.py
```diff
@@ -39,13 +39,17 @@ class Invocation(BaseModel):
...
+    """A request as the Steamship Hosting Framework receives it from the Engine.
 
-    This is the payload sent from the public-facing App Proxy to the
-    private-facing app microservice.
+    This class is different from the other `Request` class:
+     * `steamship.base.request` represents a request from the Steamship Client
+     * this class represents a request from the Steamship Engine to a Steamship-hosted App/Plugin
+
+    It contains both an app/plugin invocation and also the client configuration in which that invocation
+    is intended to execute.
```
dkolas, line 49: This comment is definitely very helpful, but might we just want to also rename the class?

src/steamship/app/request.py
```diff
@@ -39,13 +39,17 @@ class Invocation(BaseModel):
...
     """
 
     # TODO (enias): Is this a replacement for the Request in base?
-
+    # FROM (ted): No; this is the request object a
```
dkolas, line 53: You may have been interrupted when writing this sentence :)

src/steamship/app/lambda_handler.py
```diff
@@ -50,33 +55,30 @@ def _handler(event: Dict, _: Dict = None) -> Response:
             logging.exception(ex)
             return Response.error(
                 code=HTTPStatus.INTERNAL_SERVER_ERROR,
-                message="Handler was unable to initialize plugin/app.",
+                prefix=error_prefix,
+                message=f"Unable to initialize plugin/app. {ex}",
```
dkolas, line 59: Do you want to put the exception in the message and return it in the exception? Given what you did below in response.py, it seems like that will duplicate the exception string.

tests/utils/fixtures.py
```diff
@@ -1,16 +1,68 @@
...
+
+    @pytest.mark.parametrize('app_handler', [TestApp], indirect=True)
+    def test_something(app_handler):
+      response_dict = app_handler('POST', '/hello', dict())
+    ```
+
+    The app will be run its own space that gets cleaned up afterwards, and
+    the test can be written from the perspective of an external caller of the
+    app.
+    """
```
dkolas, line 54: Fancy!

dkolas (dismissed):
Approved with minor comments

tests/utils/fixtures.py
eob: I know!!!

src/steamship/app/request.py
eob: Yeah -- either that or split off the app sdk into its own thing. Will punt for now figuring we'll circle back to this

PR #104, opened 2022-07-12
basic instrumentation of lambdas for remote logging

dkolas:
Accept remote logging config from engine.  Use it to create an async fluentd logging handler.  Pass this logger to app and plugin configs so that it can be used as self.logger.

Should only be merged together with engine pr.


src/steamship/app/lambda_handler.py
```diff
@@ -83,7 +91,57 @@ def _handler(event: Dict, _: Dict = None) -> Response:
...
+        logging.basicConfig(level=logging.INFO)
+        loggingHandler = None
+        # This log statement intentionally goes to the DEFAULT logging handler, to debug logging configuration issues
+        logging.info(f"Logging host: {loggingHost} Logging port: {loggingPort}")
+
+        if (
+            loggingHost != "none"
+        ):  # Key off the string none, not 'is None', to avoid config errors where remote host isn't passed
+            # Configure remote logging
+            if loggingHost is None:
```
eob, line 116: I'd move these up higher with the other check for code readability

src/steamship/app/lambda_handler.py
```diff
@@ -83,7 +91,57 @@ def _handler(event: Dict, _: Dict = None) -> Response:
...
+                ).dict(by_alias=True)
+
+            custom_format = {
+                "level": "%(levelname)s",
+                "host": "%(hostname)s",
+                "where": "%(module)s.%(funcName)s",
+                "type": "%(levelname)s",
+                "stack_trace": "%(exc_text)s",
+            }
+            loggingHandler = fluenthandler.FluentHandler(
```
eob, line 135: Just thinking out loud -- is there an opportunity for us to auto-add things like:

- invoking user id
- lambda arn
- plugin/app instance id
- plugin/app version id
- plugin/app id
- space id

Etc? I realize we don't plumb some of that through to the invocation, but I kind of think it might be worth it to get really searchable logging data? 

(maybe we leave that to a future PR?)

src/steamship/plugin/service.py
```diff
@@ -103,15 +100,15 @@ def model_cls(self) -> Type[TrainableModel]:
...
+        self.logger.info("TrainablePluginService:run() - Loading model")
         model = self.model_cls().load_remote(
             client=self.client,  # This field comes from being a subclass of App
             plugin_instance_id=request.plugin_instance_id,
             checkpoint_handle=None,  # Will use default
             use_cache=True,
             plugin_instance_config=self.config,
         )
-        logging.info("TrainablePluginService:run() - Loaded model; invoking run_with_model")
+        self.logger.info("TrainablePluginService:run() - Loaded model; invoking run_with_model")
```
eob, line 111: Ah, I see -- so we'll encourage folks to use `self.logger`. Fine w me!

It would be cool if there were a way to hijack python's default logger, just to catch other logging messages that might be emitted by third party libraries, etc. Could investigate in a future pr...

eob (dismissed):
Minor change request, but this looks great!!!

src/steamship/app/lambda_handler.py
dkolas: Absolutely, that is the intent - I wanted to get the basic end-to-end plumbing working first, and then go back and start adding bells and whistles.  I think these are some of the first things that should be added.  I have another ticket to do this work later.

src/steamship/app/app.py
```diff
@@ -84,9 +84,13 @@ class App:
 
     _method_mappings = defaultdict(dict)
 
-    def __init__(self, client: Client = None, config: Dict[str, Any] = None):
+    def __init__(self, client: Client = None, config: Dict[str, Any] = None, logger: Logger = None):
         self.client = client
         self.config = config
+        if logger is None:
```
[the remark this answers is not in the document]
dkolas: Agreed.  This way lets you run local or run deployed and something sensible should happen.

src/steamship/app/lambda_handler.py
dkolas: You need the stuff in between to have happened though?

PR #107, opened 2022-07-14
Receive invocation context from Proxy, pipe this into remote logging

dkolas:
This gives the remote logs the plugin/app handle, instance handle, version handle, as well as spaceId, tenantId, userId.

It also undoes the creation of the self.loggers for all apps and plugins, so we return to just using logging.x directly.


src/steamship/app/lambda_handler.py
```diff
@@ -132,19 +138,27 @@ def handler(event: Dict, context: Dict = None) -> dict:
                 "type": "%(levelname)s",
                 "stack_trace": "%(exc_text)s",
                 "component": "app-plugin-lambda",
+                "userId": invocation_context.userId,
```
eob, line 141: <does an excited dance.gif>

eob (dismissed):
This looks great! 

I think the reversion back to just regular logging looks so much more natural to use

src/steamship/app/lambda_handler.py
```diff
@@ -132,19 +138,27 @@ def handler(event: Dict, context: Dict = None) -> dict:
...
+                "spaceId": invocation_context.spaceId,
+                "tenantId": invocation_context.tenantId,
+                "invocableHandle": invocation_context.invocableHandle,
+                "invocableVersionHandle": invocation_context.invocableVersionHandle,
+                "invocableType": invocation_context.invocableType,
+                "path": event.get("invocation").get("appPath")
```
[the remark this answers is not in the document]
EniasCailliau: With the default `{}` we can get rid of the if else below.

src/steamship/app/lambda_handler.py
```diff
@@ -132,19 +138,27 @@ def handler(event: Dict, context: Dict = None) -> dict:
...
+                if event.get("invocation") is not None
+                else None,
```
EniasCailliau, line 149:
```suggestion
```

src/steamship/app/request.py
```diff
@@ -43,6 +43,17 @@ class LoggingConfig(BaseModel):
     loggingPort: str = None
 
 
+class InvocationContext(BaseModel):
+    tenantId: str = None
```
EniasCailliau, line 47: Can you turn this into a CamelModel please with camel_case properties please 😍

src/steamship/app/request.py
```diff
@@ -43,6 +43,17 @@ class LoggingConfig(BaseModel):
...
+    userId: str = None
+    spaceId: str = None
+
+    invocableHandle: str = None
+    invocableVersionHandle: str = None
+    invocableInstanceHandle: str = None
+    invocableType: str = None
+
+
 class Request(BaseModel):
```
EniasCailliau, line 57: Can you turn this into a CamelModel please 🤩

src/steamship/plugin/service.py
```diff
@@ -100,15 +100,15 @@ def model_cls(self) -> Type[TrainableModel]:
 
     def run(self, request: PluginRequest[T]) -> Union[U, Response[U]]:
         """Loads the trainable model before passing the request to the `run_with_model` handler on the subclass."""
-        self.logger.info("TrainablePluginService:run() - Loading model")
+        logging.info("TrainablePluginService:run() - Loading model")
```
EniasCailliau, line 103: I love how detailed your logging statement is, this makes debugging a breeze! I am afraid that we won't always be so detailed. 

Would it be an option to upgrade the logging format to include part of this automatically? 

An example:
```
FORMAT = "[%(filename)s:%(lineno)s - %(funcName)s] %(message)s"
```

dkolas: At least for remote logging, some of these fields are already in the structured part of the message (which makes it to structured fields in Elastic).  I'll add filename and lineno!

src/steamship/app/lambda_handler.py
```diff
@@ -132,19 +138,27 @@ def handler(event: Dict, context: Dict = None) -> dict:
                 "type": "%(levelname)s",
                 "stack_trace": "%(exc_text)s",
                 "component": "app-plugin-lambda",
+                "userId": invocation_context.userId,
+                "spaceId": invocation_context.spaceId,
+                "tenantId": invocation_context.tenantId,
+                "invocableHandle": invocation_context.invocableHandle,
+                "invocableVersionHandle": invocation_context.invocableVersionHandle,
+                "invocableType": invocation_context.invocableType,
```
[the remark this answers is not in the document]
dkolas: That's covered elsewhere - the fluentd container knows what env it's in, and injects it into all the log streams.

src/steamship/app/lambda_handler.py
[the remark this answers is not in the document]
dkolas: Thank you!  I'm not sure why I always forget that shortcut.

PR #111, opened 2022-07-15
Swap out BaseModel in favour of CamelModel

EniasCailliau:
Title says it all. 

Reasons to do this: 

- All comms with server happen in camelCase. 
- Picking 1 option over 2 leaves less room for surprises 
- CamelModel works just like BaseModel without the `by_alias=True` option

Going to develop a pre-commit hook that checks if we are using BaseModel in the future.


tests/tests/app/integration/test_app_error_visibility.py
```diff
@@ -10,20 +10,6 @@
 ERROR_PYTHON_ERROR = "[ERROR - POST raise_python_error] raise_python_error"
 
 
-@pytest.mark.parametrize("app_handler", [TestApp], indirect=True)
```
eob, line 13: What's the thinking behind removing these? 

These help us achieve a guarantee that the python-side of the Client -> Engine -> Proxy -> App/Plugin stack is reporting errors in a standardized way.

PR #114, opened 2022-07-15
Install both deps and development deps with upgrade options

eob:
1) Installs both requirements.txt and requirements.dev.txt in GH Actions testing. 
2) Uses --upgrade option to suppress warnings in GHA logging output


EniasCailliau (approved):
👍

PR #124, opened 2022-08-05
Python sdk changes for large FileImport

dkolas:
This PR allows large file imports beyond the size limit of lambda return values.  The FileImporter plugin must create a steamship signed URL, and pass that as a return value in RawDataPluginOutput.

Along the way, I had to make a change so that Plugins/Apps would get a handle to a Steamship object instead of a Client object.

Depends on engine PR #304.


EniasCailliau (approved):
Looking good! Thanks for adding this in!

PR #125, opened 2022-08-10
Adds Async Plugin Support

eob:
## 1. Support for Async Plugins

Currently only async Blockifiers are supported in the Engine, but it is very easy to roll forward the others. (About ~1 day of eng).

The `PluginRequest` object gets reshaped a bit, to include:
* `status` - a Task object
* `data` - the request payload
* `context` - an object with information about, e.g. pluginInstanceId
* `is_status_check` - whether the request is checking the status of an ongoing operation

If `is_status_check` is true, the plugin should check the status, using `request.status.remote_status_input` as a `dict` object that contains information (such as a job id) to fetch the status.

If `is_status_check` is false, the plugin should initiate work.

To return finished work in either case, just return `Response(data=...)`
To return an ongoing async status in either case, just return:
```
Response(status=Task(
  state=TaskState.running,
  remote_status_input={ ...information to revive... }
))
```

Future PRs could include:
- Putting hard-coded methods on the plugin classes for `run` and `checkStatus` (rather than requiring the plugin implementor to know how to discern the difference between the two modes)
- Putting a preference field on PluginRequest which enables the engine to request either sync or async mode (right now it is entirely up to the plugin author to decide)

## Refactor train_status to use this new method.

This introduces a BREAKING CHANGE! The /train and /train_status endpoints that a trainable plugin previously exposed are now just the /train endpoint. The engine will direct all status requests to the /train endpoint as well. This means we'll need to roll forward all of our trainable plugins and the apps that use them.

## Refactor a few signatures to include PluginRequest[] and Response[] classes

Because it is now important for plugin and model implementations to have access to data outside the `request.data` field, the signatures have been modified a bit. This will also require a change to our trainable plugins. This is primarily a concern for `Model` implementors.

### DIFF
--- a/src/steamship/app/response.py
+++ b/src/steamship/app/response.py
@@ -183,7 +183,7 @@ class Response(GenericModel, Generic[T]):
                 logging.error(f"Failed calling to_dict on response object. {obj}\n {e}")
 
         if isinstance(obj, BaseModel):
-            return Response(json=obj.dict())
+            return Response(json=obj.dict(by_alias=True))
 
         return Response.error(500, message="Handler provided unknown response type.")
 
--- a/src/steamship/app/response.py
+++ b/src/steamship/app/response.py
@@ -183,7 +183,7 @@ class Response(GenericModel, Generic[T]):
                 logging.error(f"Failed calling to_dict on response object. {obj}\n {e}")
 
         if isinstance(obj, BaseModel):
-            return Response(json=obj.dict(by_alias=True))
+            return Response(json=obj.dict())
 
         return Response.error(500, message="Handler provided unknown response type.")
 
--- a/src/steamship/plugin/request.py
+++ b/src/steamship/plugin/request.py
@@ -1,0 +1,47 @@
+from __future__ import annotations
+
+from typing import Generic, Optional, TypeVar
+
+from pydantic.generics import GenericModel
+
+# Note!
+# =====
+#
+# This the files in this package are for Plugin Implementors.
+# If you are using the Steamship Client, you probably are looking for either steamship.client or steamship.data
+#
+from steamship.base import Task
+from steamship.base.configuration import CamelModel
+from steamship.base.utils import to_camel
+
+T = TypeVar("T")
+U = TypeVar("U")
+
+
+class PluginRequestContext(CamelModel):
+    """Contains the context in which"""
+
+    plugin_id: str = None
+    plugin_handle: str = None
+    plugin_version_id: str = None
+    plugin_version_handle: str = None
+    plugin_instance_id: str = None
+    plugin_instance_handle: str = None
+
+
+class PluginRequest(GenericModel, Generic[T]):
+    # The primary payload of the request. E.g. RawDataPluginInput, BlockAndTagPluginInput
+    data: Optional[T] = None
+
+    # The context in which this request is occurring
+    context: Optional[PluginRequestContext] = None
+
+    # The status of the request as perceived by the requester.
+    status: Optional[Task] = None
+
+    # Whether this plugin request is a status check against ongoing work. If True, status must be not None
+    is_status_check: bool = False
+
+    class Config:
+        alias_generator = to_camel
+        allow_population_by_field_name = True
--- a/src/steamship/plugin/service.py
+++ b/src/steamship/plugin/service.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import logging
 from abc import ABC, abstractmethod
-from typing import Any, Dict, Generic, Type, TypeVar, Union
+from typing import Any, Dict, Generic, Optional, Type, TypeVar, Union
 
 from pydantic.generics import GenericModel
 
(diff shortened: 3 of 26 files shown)


src/steamship/app/response.py
```diff
@@ -183,7 +183,7 @@ def from_obj(obj: Any) -> Response:  # noqa: C901
                 logging.error(f"Failed calling to_dict on response object. {obj}\n {e}")
 
         if isinstance(obj, BaseModel):
-            return Response(json=obj.dict())
+            return Response(json=obj.dict(by_alias=True))
```
eob, line 186: @EniasCailliau just want to call out this change -- you OK with this?

src/steamship/plugin/request.py
```diff
@@ -0,0 +1,47 @@
+from __future__ import annotations
```
eob, line 1: Split this file out because I was getting a circular import warning

src/steamship/plugin/service.py
```diff
@@ -143,7 +125,7 @@ def train(
 
     @abstractmethod
     def train_status(
-        self, request: PluginRequest[TrainStatusPluginInput], model: TrainableModel
+        self, request: PluginRequest[TrainPluginInput], model: TrainableModel
```
eob, line 128: There is no "distinct" TrainStatusPluginInput anymore -- instead `status.remote_status_input` contains the data to rehydrate the training process.

tests/assets/plugins/blockifiers/async_blockifier.py
```diff
@@ -0,0 +1,61 @@
...
+from steamship.plugin.blockifier import Blockifier
+from steamship.plugin.inputs.raw_data_plugin_input import RawDataPluginInput
+from steamship.plugin.outputs.block_and_tag_plugin_output import BlockAndTagPluginOutput
+from steamship.plugin.service import PluginRequest
+
+STATUS_CHECK_KEY = "bbq-time"
+STATUS_MESSAGE = "Still working!"
+
+
+class AsyncCsvBlockifier(CsvBlockifier, Blockifier):
```
eob, line 19: This demonstrates how to make a blockifier async! 🚢 🚢 🚢 🚢

tests/assets/plugins/taggers/plugin_third_party_trainable_tagger.py
```diff
@@ -21,7 +21,7 @@
 
 from steamship import Block, File, SteamshipError, Tag
 from steamship.app import Response, create_handler
-from steamship.base import Task
+from steamship.base import Task, TaskState
```
eob, line 24: I went in and changed all the TrainableTagger classes & tests to use the new async reporting style.

src/steamship/app/response.py
EniasCailliau: If all goes well, we don't need to call `by_alias` here. The Response's init should pass by Line 73 of `binary_utils.py` and convert it for you in the camelCase format for you: 

```
elif isinstance(json, CamelModel):
    ret_dict = json.dict(by_alias=True)
```

tests/assets/plugins/blockifiers/async_blockifier.py
EniasCailliau: 🚢 Beautiful!

EniasCailliau (approved):
Looking clean 😎 Added a few comments for you to review before merging.

src/steamship/app/response.py
eob: sg. reverting!

src/steamship/plugin/tagger.py
```diff
@@ -85,13 +84,8 @@ def train_endpoint(self, **kwargs) -> Response[TrainPluginOutput]:
...
 
-    @post("train_status")
-    def train_status_endpoint(self, **kwargs) -> Response[TrainPluginOutput]:
-        """Exposes the Service's `train_status` operation to the Steamship Engine via the expected HTTP path POST /train_status"""
-        logging.info(f"Tagger:train_status_endpoint called. Calling train_status {kwargs}")
-        arg = PluginRequest[TrainStatusPluginInput].parse_obj(kwargs)
-        model = self.model_cls()()
-        model.receive_config(config=self.config)
-        return self.train_status(arg, model)
+        if arg.is_status_check:
```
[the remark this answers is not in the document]
eob: Dave mentioned the thing yesterday. What we can actually do (in fact what's done here) is use the `is_status_check` as the wire format, but then have the base class look at it and redirect to an abstract method on the base class -> `train` or `train_status`.

So the plugin implementor is none the wiser: they just implement `train` and `train_status` but the base class uses this bit to decide.

tests/assets/plugins/blockifiers/async_blockifier.py
```diff
@@ -0,0 +1,61 @@
...
+    """This plugin is identical to the CSV blockifier, but it:
+         1) delays its response until the first re-checking of the task status
+         2) requires that the status check include the previously returned "password"
```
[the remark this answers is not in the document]
eob: Yes -- will swap

src/steamship/app/response.py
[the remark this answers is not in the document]
eob: sg. Will revert.

PR #127, opened 2022-08-12
Support better workspace selection/creation upon client init. 

eob:
This makes it easier to create a new space along with the initialization of a client.

This is an important piece of the client library to streamline because users with multiple instances of an app will want to place each instance in its own workspace to keep the data separate.

Right now if you attempt to create but a space with that name exists, it throws an error. We might want to have a flag that permits a "allow_existing" mode since I think it'll be common for people to just want to "upsert-style" create spaces in automated product code. 

```
# Default Space
Steamship()

# New Space; Autogenerated Handle
Steamship(create_space=True) # Auto-generate handle

# New Space; Known Handle
Steamship(create_space=True, space_handle="foo")

# Existing Space; Known Handle
Steamship(space_handle="foo")

# Existing Space; Known ID
Steamship(space_id="foo")
```


tests/steamship_tests/client/test_client.py
```diff
@@ -16,3 +19,56 @@ def test_user():
...
+def test_client_creation_space_failure_modes():
+    # Can't create a space by specifying an ID
+    with pytest.raises(SteamshipError):
+        get_steamship_client(space_id="foo", create_space=True)
+
+    client1 = get_steamship_client(create_space=True)
+
+    # Can't create a space with a handle that already exists
+    with pytest.raises(SteamshipError):
+        get_steamship_client(space_handle=client1.config.space_handle, create_space=True)
```
EniasCailliau, line 49: In the future, I would like the client to be in charge of deciding which space to use. I worked with 2 different spaces for my app, and passing the space handle to all the calls was dreadful.

tests/steamship_tests/client/test_client.py
```diff
@@ -16,3 +19,56 @@ def test_user():
...
     assert user.id is not None
     assert user.handle is not None
+
+
+def test_client_creation_with_new_autogen_space():
+    client1 = get_steamship_client(create_space=True)
+    client2 = get_steamship_client()
+    client3 = get_steamship_client()
+
+    assert client2.config.space_id is None
```
EniasCailliau, line 29: Wouldn't it be nicer if space_id and space_handle were always populated, even if the default space is used?

EniasCailliau (dismissed):
Less code = Efficient code :) 

We can go ahead and merge this in, just have one request with regards to the space_id and handle being None when the default space is being used.

tests/steamship_tests/client/test_client.py
eob: What do you think about removing the `space_id` and `space_handle` arguments from all our top-level APIs? It seems to me they aren't really necessary anyway, so just add API noise.

The `Configuration` object within the client already provides a way to grab a client-handle that's hard-pinned to a particular space, just as you mention preferring above (I prefer that too)

tests/steamship_tests/client/test_client.py
eob: It's an extra round trip to the server, but I'm not against it!

PR #128, opened 2022-08-16
[SHIP-445] Adds a `GET /__dir__` method to all packages.

eob:
This PR adds a `GET __dir__` and `POST __dir__` endpoint that reports back a manifest file of each package's endpoints:

- The HTTP verbs and paths
- The doc strings
- The arguments and return type

## Why include this now?

Our SDK gets bundled and zipped up into each package that is deployed. So including this BEFORE the SDK launch guarantees that we'll be able to follow up after-the-fact with OpenAPI-style features without having to re-deploy every single package.

## Why not have packages output OpenAPI directly?

OpenAPI is a much broader format than we actually need for Packages. 

A simplistic JSON format provides a dependency-free export style with **low-risk of us having to ever change it**. This is important because any change in the client will require package/plugin redeployments to adopt.

We can then process this simple format into whatever more complex scheme (like OpenAPI) we desire at the Proxy, which then acts as a single point of upgrading, bug fixing, new features, etc that will back-apply to apply to all packages/plugins without redeployment.


src/steamship/base/package_spec.py
```diff
@@ -0,0 +1,123 @@
...
+        """Returns a pretty printable representation of this package."""
+        underline = "=" * len(self.name)
+        ret = f"{self.name}\n{underline}\n"
+        if self.doc:
+            ret += f"{self.doc}\n\n"
+        else:
+            ret += "\n"
+
+        if self.methods:
+            name_width = max([len(method.path) or 0 for method in self.methods])
```
dkolas, line 119: This probably doesn't matter, but is the "or 0" in the right place here?  If method.path is None, won't the len(method.path) still throw an error?

src/steamship/invocable/invocable.py
```diff
@@ -111,43 +114,49 @@ def __init__(self, client: Steamship = None, config: Dict[str, Any] = None):
...
+                    )
+                    cls._package_spec.methods.append(method_spec)
+
+        # Add the HTTP GET /__dir__ method which returns a serialization of the PackageSpec.
+        # Wired up to both GET and POST for convenience (since POST is the default from the Python client, but
+        # GET is the default if using from a browser).
+        cls._register_mapping(name="__steamship_dir__", verb=Verb.GET, path="/__dir__")
+        cls._register_mapping(name="__steamship_dir__", verb=Verb.POST, path="/__dir__")
+        end_time = time.time()
+        logging.info(f"Registered package functions in {end_time - start_time} seconds.")
```
dkolas, line 143: How long does this generally take?

tests/steamship_tests/app/unit/test_demo_app_spec.py
```diff
@@ -0,0 +1,14 @@
...
+from assets.packages.demo_package import TestPackage
+
+
+@pytest.mark.parametrize("invocable_handler", [TestPackage], indirect=True)
+def test_app_spec(invocable_handler: Callable[[str, str, Optional[dict]], dict]):
+    """Test that the handler returns the proper directory information"""
+    response_dict = invocable_handler("GET", "/__dir__", {})
+    pp = pprint.PrettyPrinter(indent=4)
+    print()
+    pp.pprint(response_dict)
```
dkolas, line 14: Is there a missing comparison here?  It looks like it just dumps the content instead of comparing to what's in the test assets.

tests/steamship_tests/app/integration/test_package_instance.py
```diff
@@ -135,6 +135,19 @@ def get_raw(path: str):
         resp_obj = instance.invoke("json_with_status", verb=Verb.POST)
         assert resp_obj == {"status": "a string"}
 
+        # Test that the __steamship_dir__ method works
+        #
+        # Note: The output of the InvocableResponse includes a dynamically generated value:
+        #
+        #       returns': "<class 'l_dd6730cb.InvocableResponse[bytes]'>"
```
dkolas, line 142: Could regex replace the dynamically generated value in both the test asset and the test for comparison.

dkolas (dismissed):
Looks good - minor feedback that you can take or leave :)

src/steamship/base/package_spec.py
eob: Nice catch - thanks

src/steamship/invocable/invocable.py
eob: I don't know; that's why I added it.

This adds a bit of Python introspection to the Package "bootup" process, which happens on every single invocation. I'm not sure how must that costs so figured we could keep track.

An alternative implementation would be to only generate this introspective manifest when __dir__ is actually called. TBH that's probably the better approach.. just didn't occur to me until I'd already written all the code.

tests/steamship_tests/app/integration/test_package_instance.py
eob: Ooh that's a good idea. I'll do that now. Just didn't have the gas left in my tank before to finish the job.

PR #130, opened 2022-08-16
Always load a space_id and space_handle in client, even if using defaults

eob:
This is for you @EniasCailliau :)


EniasCailliau (approved):
Awesome! Thanks!

PR #136, opened 2022-08-28
Always create config object for plugins

eob:
We already require that Blockifiers and Taggers define a config class, but currently we sometimes to not instantiate it.

This change ensures that we always attempt to instantiate the config class. If it can not be instantiated due to a validation error, an error will be thrown. But if an error is not thrown, the implementor will not have to check for None.


dkolas (approved):
These four lines look GREAT :)

src/steamship/plugin/blockifier.py
```diff
@@ -24,6 +24,8 @@ def __init__(self, client: Client = None, config: Dict[str, Any] = None):
         super().__init__(client, config)
         if config:
             self.config = self.config_cls()(**config)
```
EniasCailliau, line 26:
```suggestion
            self.config = self.config_cls()(**config) if config else self.config_cls()()
```

Just showing off 🤓

EniasCailliau (approved):
👍

PR #137, opened 2022-08-29
SteamshipResponse.wait() throws on timeout

dkolas:
Making wait() throw an exception if it times out without the task completing.  Previously it would just happily return, which led to odd behavior when trying to use .data.

### DIFF
--- a/src/steamship/base/response.py
+++ b/src/steamship/base/response.py
@@ -58,6 +58,12 @@ class Response(GenericModel, Generic[T]):
             self.refresh()
             time.sleep(retry_delay_s)
 
+        # If the task did not complete within the timout, throw an error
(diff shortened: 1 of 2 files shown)


src/steamship/base/response.py
```diff
@@ -58,6 +58,12 @@ def wait(self, max_timeout_s: float = 60, retry_delay_s: float = 1):
             self.refresh()
             time.sleep(retry_delay_s)
 
+        # If the task did not complete within the timout, throw an error
+        if self.task.state not in (TaskState.succeeded, TaskState.failed):
```
eob, line 62: TIL a tuple is possible here

eob (approved):
LGTM!

PR #138, opened 2022-08-29
Add serializer for exceptions for remote logging

dkolas:
Currently in elastic logs for python client we sometimes see CRITICAL - can't output to log.  This is because we sometimes call logging with an exception directly, ex:

```
 error = SteamshipError("Something is amiss!")
 logging.error(error)
```

The 'cant output to log' is because the SteamshipError is not msgpack serializable.  This PR makes SteamshipError msgpack serializable with its to_dict method, and makes any other exception serializable with its args.

### DIFF
--- a/src/steamship/app/lambda_handler.py
+++ b/src/steamship/app/lambda_handler.py
@@ -19,10 +19,12 @@ from steamship.utils.signed_urls import upload_to_signed_url
 
 
 def encode_exception(obj):
+    """When logging an exception ex: logging.exception(some_error), the exception must be turned into a string
+    so that it is accepted by elasticsearch"""
     if isinstance(obj, SteamshipError):
-        return obj.to_dict()
+        return json.dumps(obj.to_dict())
     if isinstance(obj, Exception):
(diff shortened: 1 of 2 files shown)


src/steamship/app/lambda_handler.py
```diff
@@ -145,6 +153,7 @@ def handler(event: Dict, context: Dict = None) -> dict:
                 host=logging_host,
                 port=logging_port,
                 nanosecond_precision=True,
+                msgpack_kwargs={"default": encode_exception},
```
eob, line 156: LGTM but this goes over my head so my review is of "sanity check only" quality.

eob (approved):
LGTM

src/steamship/app/lambda_handler.py
dkolas: It's mainly stack overflow copypasta:

The gist is that the fluent sender uses the msgpack library for efficient data sending (something we may consider down the line!), but msgpack does not by default know how to serialize arbitrary objects, including exceptions.  This line tells fluent to tell msgpack to use the new encode_exception function to attempt to serialize them.

PR #139, opened 2022-08-31
Bugfix: the space-loading handlers on Steamship init didn't permit create-or-fetch behavior

eob:


src/steamship/base/client.py
```diff
@@ -63,10 +64,19 @@ def __init__(
...
+            space_id=space_id,
+            space_handle=space_handle,
+            create_space=create_space,
+            fetch_or_create_space=fetch_or_create_space,
+        )
 
     def switch_space(
-        self, space_id: str = None, space_handle: str = None, create_space: bool = False
+        self,
+        space_id: str = None,
```
EniasCailliau, line 76: Do we need both space_id and space_handle? Aren't both unique to a given user? 

You can argue that space_ids are unique across users too, but from the scope of a user, that doesn't increase the utility of using it. 

Possible upside: The user cannot get a conflict by providing a `space_id` that doesn't exist and cannot be created (as the engine doesn't allow you specify the id). This is an exception flow that doesn't exist when the `space_handle` is used.

src/steamship/base/client.py
```diff
@@ -42,6 +42,7 @@ def __init__(
         space_id: str = None,
         space_handle: str = None,
         create_space: bool = False,
+        fetch_or_create_space: bool = False,
```
EniasCailliau, line 45: Trying to simplify it from the perspective of a user. 

Thoughts:
1) If I name a space using a  space_handle, I expect that to be unique. In other words, I cannot create two workspaces with the same name.
2) If a space does not exist, the path of least resistance for me would be that it is created, and I can start using it. In other words, I want you to create a space by default and notify me about it's creation. 
3) I I am running a debug script, I want to make sure I don't accidentally re-use an existing space. So I sometimes want the option to fail_when_exists.

That said I would: 
1. Remove `space_id` option
2. This function essentially does a `get_or_create`. You can override this behavior with a `create_space` and `fetch_or_create_space` flag. 

In the future, I think we should go for a design that revolves around workspaces and uses explicit function handles: 

```
client = Steamship(profile="staging") 
workspace = client.get(space_handle="prod") OR client.get_or_create(space_handle="prod") OR client.create(space_handle="prod") # This is inspired by the Django API, each function has 1 behavior defined by its function handle. 

File.create(workspace, ...) 


OR: 

with workspace: 
    File.create(...)
```

src/steamship/client/client.py
```diff
@@ -28,6 +28,7 @@ def __init__(
         space_id: str = None,
         space_handle: str = None,
         create_space: bool = False,
+        fetch_or_create_space: bool = False,
```
EniasCailliau, line 31: Isn't this what you'd want as default?  If I provide a space_handle, I expect to re-use or create a space. I'm guessing most of our audience would turn this param on as otherwise, the code cannot be run more than once.

EniasCailliau (approved):
Would set `fetch_or_create` default to True. 

I also added feedback about simplifying the interface, but we can focus on that when we think it's time to re-design the python client.

src/steamship/base/client.py
dkolas: I'm confused about the param naming on these. Presumably `create_space` means that you can auto-create it if it doesn't already exist.  But then does `fetch_or_create_space` being true imply that create_space must be true, or are they mutually exclusive?

src/steamship/base/client.py
dkolas: Right now we seem to have a mix of allowing handles / ids to be used to fetch things.  I think some more consistency on this across the board would be nice, but for the switch_space use case specifically, I would be happy to drop ID from the interface.

Question though - when these are being initialized from the proxy in a plugin / app call, is the space handle passed or just the id?

dkolas (approved):
Minor comments about naming and such.

src/steamship/base/client.py
eob: I think there are two separate threads above:

## 1. "Should space, not client, but the root object for many things?"

@EniasCailliau I can get onboard with the idea of `workspace` being the primary handle into many operations, like you sketch above:

```
client = Steamship(..)
workspace = client.use_workspace(..)
package = workspace.use_package(..)
```

@dkolas do you like this option?

I think the best way to achieve this might be to refactor Space/Workspace a subclass of Client. The notion of "which space are you in" is so fundamental to the API that "Workspace" is really more of a client scope than a data object in the sense that other things like File, Tag, etc are data objects. Maybe we kick this into a Notion document where it's easier to bat around a few ideas..

## How can we easily create/reload Spaces?

It sounds like the common feeling is that we should pursue "create if not exist" as the default, but maybe have to figure out how aggressively to do it:

* Steamship(space_handle="FOO") <- @EniasCailliau, above
* Steamship(space_handle="FOO", create_space = True) <- @dkolas, above
* Steamship(space_handle="FOO", create_or_fetch_space = True) <- This PR, currently

I'll refactor for @EniasCailliau's suggestion of 

* Steamship(space_handle="FOO")
* Steamship(space_handle="FOO", fail_if_exists=True)

With the note that this means spaces have slightly different creation semantics than other model objects. (I'm ok w/ that, and/or that we might bring others inline with these semantics as well)

src/steamship/base/client.py
eob: Agree that `*_id` is superfluous given that handles are the preferred way of referencing things.

IIRC both the engine and proxy support `space_handle` in addition to `space_id`, though the current client code may be passing id behind the scenes to avoid an extra db round-trip (since `id` is the fk in the db)

EniasCailliau (approved):
👍 Beauty!

PR #140, opened 2022-08-31
Special case handling: when a package endpoint returns something that looks like SteamshipResponse but isn't

eob:
I found a fun bug testing the ticket-tagger app:

The method:

```
response = instance.post('specialize_status')
```

returns the response:

```
{ "status": "UNSPECIALIZED"}
```

which our base client attempts to helpfully parse as a SteamshipResponse<T> object because of the "status" field. But it isn't one, and as a result it will throw.

This PR only throws when the "status" field fails to parse if it's an API call, but not a package call.


src/steamship/base/client.py
```diff
@@ -371,14 +371,25 @@ def call(  # noqa: C901
...
-                # if task_resp is not None and task_resp.taskId is not None:
-                #     task = Task(client=self)
-                #     task.update(task_resp)
-                if "state" in response_data["status"]:
-                    if response_data["status"]["state"] == "failed":
-                        error = SteamshipError.from_dict(response_data["status"])
-                        logging.error(f"Client received error from server: {error}")
+                try:
+                    task = Task.parse_obj({**response_data["status"], "client": self})
+                    # if task_resp is not None and task_resp.taskId is not None:
```
EniasCailliau, line 376: Can we remove these comments?

src/steamship/base/client.py
```diff
@@ -371,14 +371,25 @@ def call(  # noqa: C901
...
+                    #     task = Task(client=self)
+                    #     task.update(task_resp)
+                    if "state" in response_data["status"]:
+                        if response_data["status"]["state"] == "failed":
+                            error = SteamshipError.from_dict(response_data["status"])
+                            logging.error(f"Client received error from server: {error}")
+                except BaseException as e:
```
EniasCailliau, line 383: I would use a specific exception here as we don't want this except clause to be a catch-all sink.

src/steamship/base/client.py
```diff
@@ -371,14 +371,25 @@ def call(  # noqa: C901
...
-                if "state" in response_data["status"]:
-                    if response_data["status"]["state"] == "failed":
-                        error = SteamshipError.from_dict(response_data["status"])
-                        logging.error(f"Client received error from server: {error}")
+                try:
+                    task = Task.parse_obj({**response_data["status"], "client": self})
+                    # if task_resp is not None and task_resp.taskId is not None:
+                    #     task = Task(client=self)
+                    #     task.update(task_resp)
+                    if "state" in response_data["status"]:
```
EniasCailliau, line 379:
```suggestion
                    if "state" in response_data["status"] and response_data["status"]["state"] == "failed"
```

The right-hand statement is only executed when the left-hand statement is True.

src/steamship/base/client.py
```diff
@@ -371,14 +371,25 @@ def call(  # noqa: C901
...
+                        if response_data["status"]["state"] == "failed":
```
EniasCailliau, line 380:
```suggestion
```
See comment above

EniasCailliau (approved):
Interesting bug! Added 3 optional comments.

dkolas (dismissed):

src/steamship/base/client.py
eob: Done!

src/steamship/base/client.py
eob: Nice catch -- will update.

src/steamship/base/client.py
eob: Fixed this with a slight refactor instead so that we're relying on model objects instead of peeking into dicts

PR #142, opened 2022-09-14
Add get method to Task object

EniasCailliau:


src/steamship/base/tasks.py
```diff
@@ -192,6 +198,17 @@ def dict(self, **kwargs) -> Dict[str, Any]:
             }
         return super().dict(**kwargs)
 
+    @classmethod
+    def get(cls, client, _id) -> IResponse[Task]:
```
eob, line 202: nit: can we type the args of `get`?

also below I think OK to just leave out the `space*` args?

eob (approved):
1 small nit, but LGTM!

PR #144, opened 2022-09-26
Support steamship secrets

EniasCailliau:


eob (approved):
LGTM!

src/steamship/plugin/blockifier.py
```diff
@@ -22,8 +22,8 @@ class Blockifier(PluginService[RawDataPluginInput, BlockAndTagPluginOutput], ABC
     # noinspection PyUnusedLocal
     def __init__(self, client: Client = None, config: Dict[str, Any] = None):
         super().__init__(client, config)
-        if config:
-            self.config = self.config_cls()(**config)
+        if self.config:
```
eob, line 25: I'm pretty confident the prior version is actually correct? I just followed up the `super()` chain and nothing above this class actually does anything with the `config` passed in. It's here that we have to call `self.config_cls()(**config)` if it's non None

eob (changes requested):
I'm pretty sure the changes to the `config` handling introduce an error; lmk if I'm reading it wrong.

Secrets handling looks great!

PR #146, opened 2022-09-30
Remove to dict except for SteamshipError

EniasCailliau:
Sadly we can't upgrade SteamshipError to an Exception as you can't inherit from both `pydantic.BaseModel` and `Exception`.


eob (commented):
LGTM!

PR #147, opened 2022-09-30
Helper KV Store class & Tests (in test folder; not in main client)

eob:
A simple key value store implemented in Steamship.

    Each entry is stored as an empty file whose "Key" has:
      * Kind = KeyValueStore
      * Name = the key of the (kv) pair
      * Value = a dict set to the value

    This class is stored inside the `tests` folder because:
      1) It has proven useful enough in package development that we wish to make sure it's tested, and
      2) We wish to add KV-functionality to Spaces in general, but
      3) We're not eager to directly add this as a utility class in the codebase.

    For the time being; consider this a copy-paste reference implementation along with tests.


tests/steamship_tests/utils/test_key_value_store.py
```diff
@@ -0,0 +1,237 @@
+from enum import Enum
+from typing import Any, Dict, Optional
+
+from steamship_tests.utils.client import get_steamship_client
+from steamship_tests.utils.random import random_name
+
+from steamship import Block, File, Space, Steamship, Tag
+
+
+class KeyValueStore:
```
EniasCailliau, line 10: Should we add this to our utils?

tests/steamship_tests/utils/test_key_value_store.py
```diff
@@ -0,0 +1,237 @@
...
+      2) We wish to add KV-functionality to Spaces in general, but
+      3) We're not eager to directly add this as a utility class in the codebase.
+
+    For the time being; consider this a copy-paste reference implementation along with tests.
+    """
+
+    client: Steamship
+    namespace: str
+
+    def __init__(self, client: Steamship, namespace: str = "KeyValueStore"):
```
EniasCailliau, line 29: Would add docs to explain namespace != workspace, instead it uniquely identifies the key-value store you will use. You can have more than one KV store in a single workspace.

tests/steamship_tests/utils/test_key_value_store.py
```diff
@@ -0,0 +1,237 @@
...
+        """Deletes the entry represented by `key`"""
+        file = self._get_file()
+
+        if file is None:
+            return False
+
+        deleted = False
+        for tag in file.tags:
+            if tag.kind == self.namespace and tag.name == key:
+                tag.delete()
```
EniasCailliau, line 67:
```suggestion
                tag.delete()
                deleted = True
```

tests/steamship_tests/utils/test_key_value_store.py
```diff
@@ -0,0 +1,237 @@
...
+
+class SpecializationStatus(str, Enum):
+    """These are for use with the"""
+
+    UNSPECIALIZED = "unspecialized"
+    SPECIALIZED = "specialized"
+    SPECIALIZATION_IN_PROGRESS = "specialization_in_progress"
+
+    @staticmethod
+    def from_str(string: str) -> "SpecializationStatus":
```
EniasCailliau, line 98: If I'm not mistaken, we don't need this method. Instead, you can use the default init method of Enum: `SpecializationStatus("unspecialized")`.

EniasCailliau (approved):
3 small comments. LGTM! So excited to have this part of our SDK!

tests/steamship_tests/utils/test_key_value_store.py
eob: Nice catch :)

tests/steamship_tests/utils/test_key_value_store.py
eob: Good call -- renamed the variable and added a note about having multiple.

PR #153, opened 2022-10-04
Reduce surface of the python client in preparation of removing circular dependencies

EniasCailliau:
I also added comments referencing some of the Linear tickets

### DIFF
--- a/tests/steamship_tests/plugin/integration/test_e2e_embedder.py
+++ b/tests/steamship_tests/plugin/integration/test_e2e_embedder.py
@@ -42,4 +42,5 @@ def test_e2e_embedder():
         # Now lets run all the other embedding steamship_tests
         basic_embeddings(instance)
         basic_embedding_search(client, plugin_instance=instance.handle)
-        create_index(client, plugin_instance=instance.handle)
+        index = create_index(client, plugin_instance=instance.handle).data
+        index.delete()


src/steamship/app/app.py
```diff
@@ -72,7 +72,7 @@ def post(path: str, **kwargs):
     return endpoint(verb=Verb.POST, path=path, **kwargs)
 
 
-class App:
+class App:  # TODO (enias): Should be renamed to Invocable -> Solves redefinition of app
```
eob, line 75: Perhaps in a base class thought? 

From the user perspective, might make more sense to have `Package` and `Plugin` derive from `Invocable`

src/steamship/app/request.py
```diff
@@ -36,7 +36,7 @@ class InvocationContext(CamelModel):
     invocable_type: str = None
 
 
-class Request(CamelModel):
+class Request(CamelModel):  # TODO (enias): Rename to ExecutableRequest?
```
eob, line 39: InvocableRequest? (given that that's the name you mentioned above in the PR)

eob (commented):
LGTM so far; unsure if ready for merge or just placeholder for future PR!

src/steamship/app/request.py
EniasCailliau: Yes agree! I'll feel it out today in another PR.

PR #157, opened 2022-10-10
Use requirements.txt as a single source of truth for dependencies

EniasCailliau:


EniasCailliau:
☑️ Integration tests for packages using the client are successful, too, confirming pip installing works as expected.

PR #158, opened 2022-10-10
[SHIP-386] Split File.create and File.create_plugin to support future sync and async return 

EniasCailliau:


src/steamship/data/file.py
```diff
@@ -117,59 +117,60 @@ def get(
     @staticmethod
     def create(
         client: Client,
-        filename: str = None,
```
EniasCailliau, line 120: I decided to remove the `filename` arg as I perceive it as having little utility while adding some magic (deriving the filetype based on the filename).

dkolas: I'm good with this.  I think if we're going to have any automatic mappings from file types to plugins, it would be better to do them with mime types than with file extensions.

dkolas (approved):
I think when we talked about this, I was assuming we'd change the engine side too, but since it's done so easily this way, I'm not sure that would be a good use of time.  I vote to carry on with client-side-only and we can come back and change the API if we think it's needed later.

EniasCailliau:
> I think when we talked about this, I was assuming we'd change the engine side too, but since it's done so easily this way, I'm not sure that would be a good use of time. I vote to carry on with client-side-only and we can come back and change the API if we think it's needed later.

100% agree. That's why I love how the python client serves as a facade to cover spots we don't want to reveal yet. Not changing it in the engine leaves us the option to quickly revert our decisions too.

PR #160, opened 2022-10-10
[Ship 388] change all synchronous methods to return objects and async methods to return tasks

EniasCailliau:


tests/steamship_tests/plugin/integration/test_e2e_async_blockifier.py
```diff
@@ -53,4 +46,3 @@ def test_e2e_async_blockifier_plugin():
                 for tag in block.tags:
                     assert tag.name is not None
                     assert tag.kind is not None
-            file.delete()
```
dkolas, line 56: Did you mean to delete this delete?

dkolas: Nevermind you did, because it's within a resource cleanup block

dkolas (dismissed):
This is different than the way I might have done it - just calling `response.data` from within every call to `.create()` for example.  I was worried for a second that it would change the plugin response interface, but it turns out that was already a different class!

So all of this looks great to me.  I think I spotted one minor typo in a test.

tests/steamship_tests/client/operations/test_embedding_index.py
```diff
@@ -238,42 +251,50 @@ def test_multiple_queries():
...
-        assert search_results.data.items[0].value.query == qs4[0]
-        assert search_results.data.items[1].value.value == a3
-        assert search_results.data.items[1].value.query == qs4[1]
+        search_results.wait()
+        search_results = search_results.output
+        assert len(search_results.items) == 2
+        assert search_results.items[0].value.value == a4
+        assert search_results.items[0].value.query == qs4[0]
+        assert search_results.items[1].value.value == a3
+        assert search_results.items[1].value.query == qs4[1]
```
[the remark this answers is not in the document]
EniasCailliau: My mind: ⁍💥☢️

tests/steamship_tests/plugin/integration/test_e2e_async_blockifier.py
EniasCailliau: Yes, we now can't execute `delete()` twice without raising an Exception which I think is what we want.

EniasCailliau:
> This is different than the way I might have done it - just calling `response.data` from within every call to `.create()` for example. I was worried for a second that it would change the plugin response interface, but it turns out that was already a different class!
> 
> So all of this looks great to me. I think I spotted one minor typo in a test.

Interesting. I started doing that but then realized the base.Response class had no purpose anymore.

src/steamship/base/request.py
```diff
@@ -6,6 +6,11 @@ class Config:
         use_enum_values = True
 
 
+class Response(CamelModel):
+    class Config:
+        use_enum_values = True
```
eob, line 11: Small request for a comment to explain what this line does for future readers :)

src/steamship/data/plugin_version.py
```diff
@@ -65,7 +63,7 @@ def create(
         is_public: bool = None,
         is_default: bool = None,
         config_template: Dict[str, Any] = None,
-    ) -> Response[PluginVersion]:
+    ) -> Task[PluginVersion]:
```
eob, line 66: Side comment: I kind of like that we're calling the response `Task` instead of `Response`. 

It feels like it explains the nature of what we're returning far better.

tests/steamship_tests/base/test_task.py
```diff
@@ -14,66 +14,65 @@ def test_background_task_call():
...
-    result_2.wait()
+    result_2_task.wait()
 
     # And now it has completed
-    assert result_2.task.state == TaskState.succeeded
-    assert result_2.data is not None
-    assert type(result_1.data) == NoOpResult
+    assert result_2_task.state == TaskState.succeeded
+    assert result_2_task.output is not None
+    assert type(result_task_1) == NoOpResult
```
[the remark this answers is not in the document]
eob: Eagle eyes!

tests/steamship_tests/client/operations/test_embed_file.py
```diff
@@ -29,36 +29,35 @@ def test_file_parse():
...
         index.insert_file(file.id, reindex=False)
         embed_resp = index.embed()
-        assert embed_resp.error is None
         embed_resp.wait()
 
-        res = index.search("What color are roses?").data
-        assert len(res.items) == 1
+        res = index.search("What color are roses?")
+        res.wait()
+        items = res.output.items
```
eob, line 57: Going to go back and review the Task object, but just noting here aloud that async data now live at `task.output`. 👍

eob (dismissed):
LGTM! 

Leave me wondering if we might eventually want some syntactic sugar like `task.wait_returning()` that returns the wrapped `.output` instead of just delaying the task.

That would enable really clean async code like:

data = client.async_method().wait_returning()

it would throw upon error

src/steamship/data/plugin_version.py
EniasCailliau: I also like the as a user I know what I should do based on the response type. I could write this re-usable code snippet:

```
if isinstance(response, Task): 
  # wait and fetch
  response.wait()
  o = response.output
else: 
   o = response 
```

tests/steamship_tests/base/test_task.py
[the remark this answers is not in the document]
EniasCailliau: Good catch! 🦅👀

PR #165, opened 2022-10-11
Rename core elements

EniasCailliau:


requirements.dev.txt
```diff
@@ -31,7 +31,6 @@ pluggy==1.0.0
 pre-commit==2.19.0
 py==1.11.0
 pycln==1.3.2
-pydantic==1.9.0
```
EniasCailliau, line 34: Had to upgrade because of a pickling bug with HttpUrl.

src/steamship/app/invocable.py
```diff
@@ -76,7 +77,7 @@ def post(path: str, **kwargs):
     return endpoint(verb=Verb.POST, path=path, **kwargs)
 
 
-class App:  # TODO (enias): Should be renamed to Invocable -> Solves redefinition of app
```
EniasCailliau, line 79: Renamed to Invocable (similar to the engine) and made it a base class.

src/steamship/app/invocable_request.py
```diff
@@ -5,15 +5,6 @@
 from steamship.base.configuration import CamelModel, Configuration
 
 
-def event_to_config(event: dict) -> Configuration:
```
EniasCailliau, line 8: This piece of code is not used. Happy to put it back if we need it elsewhere.

src/steamship/app/invocable_request.py
```diff
@@ -36,7 +27,7 @@ class InvocationContext(CamelModel):
     invocable_type: str = None
 
 
-class Request(CamelModel):  # TODO (enias): Rename to ExecutableRequest?
```
EniasCailliau, line 39: Renamed so to resolve conflict with base.request.Request.

src/steamship/app/invocable_response.py
```diff
@@ -31,7 +31,7 @@ class Http(CamelModel):
 T = TypeVar("T")
 
 
-class Response(GenericModel, Generic[T]):  # TODO (enias): Rename to ExecutableResponse?
```
EniasCailliau, line 34: Renamed to resolve conflict with `base.response.Response`

src/steamship/app/invocable_response.py
```diff
@@ -121,7 +121,7 @@ def error(
         error: Optional[SteamshipError] = None,
         exception: Optional[Exception] = None,
         prefix: Optional[str] = None,
-    ) -> Response[T]:
```
EniasCailliau, line 124: I personally like this naming a lot. It's a Response that Invocables will return.

src/steamship/base/base.py
```diff
@@ -1,9 +1,27 @@
...
 from typing import TypeVar
 
+import inflection
+from pydantic import BaseModel
 from pydantic.generics import GenericModel
 
 T = TypeVar("T")  # Declare type variable
 
 
-class IResponse(GenericModel):
```
EniasCailliau, line 8: We don't need the interface anymore 🎉 Let's nuke these circular dependencies ☢️😈

src/steamship/base/base.py
```diff
@@ -1,9 +1,27 @@
...
+def to_camel(s: str) -> str:
```
EniasCailliau, line 11: Moved to base to resolve circular dependency with `base.configuration`

src/steamship/base/client.py
```diff
@@ -141,7 +141,7 @@ def dict(self, **kwargs) -> dict:
         # Because of the trick we do to hack these in as both static and member methods (with different
         # implementations), Pydantic will try to include them by default. So we have to suppress that otherwise
         # downstream serialization into JSON will fail.
-        if "exclude" not in kwargs:
+        if "exclude" not in kwargs:  # TODO (enias): Review
```
EniasCailliau, line 144: I think I've found an alternative solution. The new solution is part of this PR for testing, will circle back with a new PR when I confirmed it works as expected.

src/steamship/base/tasks.py
```diff
@@ -31,7 +30,7 @@ class ListTaskCommentRequest(Request):
 
 
 class TaskComment(CamelModel):
-    client: Any = None
+    client: Any = Field(None, exclude=True)
```
EniasCailliau, line 33: Potentially a cleaner solution to remove client from serialisation (e.g. .dict()). 

This will become way cleaner once all our objects use a base class. But for now I do think it's cleaner since the user can understand why client will be excluded from dict by looking at the data model (vs nested in the lambda handler).

EniasCailliau: WDYT?

eob:
All these comments are really helpful in reading teh PR -- thank you!!!

src/steamship/app/invocable_request.py
eob: Really glad to see this -- if we can I think probably good to avoid any naming duplicates in our code base

src/steamship/app/invocable_response.py
eob: 👍 to all the renames!

src/steamship/base/client.py
eob: 👍

src/steamship/base/client.py
```diff
@@ -287,6 +287,8 @@ def _add_client_to_object(self, expect, response_data):
         if expect and isclass(expect):
             if len(response_data.keys()) == 1 and list(response_data.keys())[0] in (
                 to_camel(expect.__name__),
+                to_camel(expect.__name__).replace("package", "app"),
+                # Hack since engine uses "App" instead of "Package"
```
eob, line 291: Thanks for leaving this comment in -- this would get lost to history otherwise

src/steamship/base/client.py
```diff
@@ -402,8 +404,8 @@ def call(  # noqa: C901
 
             if "data" in response_data:
                 if expect is not None:
-                    if hasattr(expect, "from_dict"):
```
eob, line 405: It's exciting to see the removal of all the to_dict and from_dict spreading :)

src/steamship/base/tasks.py
eob: I don't have a strong opinion personally, though I do agree this is more explicit than being clever and "magical" down at the serialization level. Happy to run with this approach if it works!

src/steamship/data/__init__.py
```diff
@@ -1,19 +1,19 @@
...
 from .plugin_instance import PluginInstance
 from .plugin_version import PluginVersion
 from .space import Space
 from .tags import DocTag, Tag, TagKind, TextTag
 
 __all__ = [
-    "App",
-    "AppInstance",
-    "AppVersion",
+    "Package",
```
eob, line 14: Hooray!!!

eob (dismissed):
LGTM! Excited for all these improvements!

PR #166, opened 2022-10-11
[SHIP-148] Swap kind/name requiredness on Tags

dkolas:
Required test changes to accommodate https://github.com/nludb/nludb/pull/348


eob (dismissed):
LGTM. Some of the semantics feel odd in the tests below as a result; but the semantics aren't important since it's just a unit test.

PR #171, opened 2022-10-14
Remove dynamic excludes in favour of explicit Field(exclude=True) definitions

EniasCailliau:


eob (approved):
LGTM Assuming the tests are still passing (there are tests now which catch the exception that the absence of this was inducing before)

PR #175, opened 2022-10-15
Fixes bug in which snake-cased plugin objects would not get camel cased correctly.

eob:
If a plugin returned a raw response (e.g. BlockAndTagPluginOutput) then the serialization into JSON was not being performed correctly w/r/t snake and camel case.


EniasCailliau (approved):
Thanks for catching this! LGTM 🏎

PR #176, opened 2022-10-16
More constants to encourage Tag.kind, Tag.name, and Tag.value.sub_key convergence

eob:
This PR is a small step in the direction of Tag kind/name/value.key convergence.

* I've combined the various constants files we had into one tag_constants file
* Added an `embedding` kind
* Added a `sentiment` kind with constants appropriate to NLPCloud and OneAI
* Added an `emotion` kind with constants appropriate to NLPCloud and OneAI


src/steamship/__init__.py
```diff
@@ -14,16 +14,19 @@
     Block,
     DocTag,
     EmbeddingIndex,
+    EmotionTag,
```
EniasCailliau, line 17: I would keep these in `steamship.data` as they are not the primary building blocks of Steamship. WDYT?

src/steamship/data/tags/tag_constants.py
```diff
@@ -1,8 +1,21 @@
...
 
 
+class TagKind:
+    """A set of `kind` constants for Tags."""
+
+    pos = "pos"  # Part of speech tags
+    dep = "dep"  # Dependency tags
+    sentiment = "sentiment"  # Sentiment tags
+    emotion = "emotion"  # Emotion tags
+    ent = "ent"  # Entity tags, such as PERSON or ORG
```
EniasCailliau, line 11:
```suggestion
    ent = "entity"  # Entity tags, such as PERSON or ORG
```

I prefer full name vs abbreviations as they require less documentation.

src/steamship/data/tags/tag_constants.py
```diff
@@ -1,8 +1,21 @@
 from typing import Optional
 
 
+class TagKind:
```
EniasCailliau, line 4: Can this be a `(str, Enum)`?

src/steamship/data/tags/tag_constants.py
```diff
@@ -1,8 +1,21 @@
...
+    """A set of `kind` constants for Tags."""
+
+    pos = "pos"  # Part of speech tags
```
EniasCailliau, line 7: If we want to turn this into an Enum it's recommended to use all caps hooks. 

```suggestion
    POS = "pos"  # Part of speech tags
```

EniasCailliau (changes requested):
Really useful collection of tags. I would turn these into Enums to make them work with pydantic. WDYT? I think it's a simple modification that will serve us in the future.

src/steamship/data/tags/tag_constants.py
dkolas: Can we spell out all of the tag names and values?  Every NLP human knows POS = Part of speech, but your avg software dev may not.

src/steamship/data/tags/tag_constants.py
dkolas: Agree. Longer names don't cost us anything.

src/steamship/data/tags/tag_constants.py
```diff
@@ -87,3 +100,69 @@ def from_html_tag(tagname: Optional[str]) -> Optional[str]:  # noqa: C901
             return DocTag.teletype
 
         return None
+
+
+class TokenTag:
+    """A set of `name` constants for Tags with a `kind` of `TagKind.token`; appropriate for parsing-level ideas."""
+
+    TextWithWs = "textWithWs"
```
dkolas, line 108: TextWithWs, like.... "Where's Waldo? He's out Wandering" = true; "The magician made the letter that came before x and after v disappear" = false.

Just being silly - I assume this is whitespace - it should be spelled out.

src/steamship/data/tags/tag_constants.py
```diff
@@ -87,3 +100,69 @@ def from_html_tag(tagname: Optional[str]) -> Optional[str]:  # noqa: C901
...
+    IsLeftPunct = "isLeftPunct"
+    IsRightPunct = "isRightPunct"
+    IsSpace = "isSpace"
+    IsBracket = "isBracket"
+    IsQuote = "isQuote"
+    IsCurrency = "isCurrency"
+    LikeUrl = "likeUrl"
+    LikeNum = "likeNum"
+    LikeEmail = "likeEmail"
+    IsOov = "isOov"
```
dkolas, line 134: Spell me out!

src/steamship/data/tags/tag_constants.py
```diff
@@ -87,3 +100,69 @@ def from_html_tag(tagname: Optional[str]) -> Optional[str]:  # noqa: C901
...
+    IsStop = "isStop"
```
dkolas, line 135: isStopword?

dkolas (changes requested):
I feel pretty strongly that these tag enums need to be spelled out much more before we start using them.

src/steamship/data/tags/tag_constants.py
eob: 😆  "W" finding is big business!

(yeah it's whitespace)

PR #178, opened 2022-10-17
WIP: Add tests for likely error-inducing use case of `use` and `use_plugin`

eob:
@dkolas adding just a few client-side tests to mirror your PR in the Engine since I ran into these while hacking around atop the client. Will wait to test & send review until SHIP-366 is merged


dkolas:
All the relevant engine PRs should be merged now.

PR #179, opened 2022-10-17
Touch-ups to Package/App Configuration [SHIP-423] [SHIP-421] [SHIP-420] [SHIP-418]

eob:
Our current handling of Package & Plugin configuration is inconsistent across types:

At present:
- Some (not all) Plugins require Config-derived Pydantic configs, whereas Packages just use JSON Dicts
- The Pydantic Config object rewrites snake_case to camelCase
- The Config object lives within the .plugin package namespace

This PR unifies our configuration handling. 

Major, breaking changes are:

- Major change: EVERY PACKAGE AND PLUGIN must now implement the `config_cls` endpoint, even if just to defer to the base (empty) `Config` object

And additional conveniences are:

- The config parsing has been moved down to the base `Invocation` object
- Package & Plugin config handling is now identical
- Config no longer derives from CamelModel so that the config keys will be read in EXACTLY as a user defined them

## Additional changes

- Change the Pydantic type of the URLs on the root Configuration object to be more permissive so that `http://localhost` is permitted.

Notes: Passes all tests on localhost against the Engine's `main`


src/steamship/app/config.py
```diff
@@ -6,7 +6,7 @@
 
 
 class Config(CamelModel):
-    """Base class for a Steamship Plugin's configuration object."""
+    """Base class Steamship Package and Plugin configuration objects."""
```
EniasCailliau, line 9:
```suggestion
    """Base class for Steamship Package and Plugin configuration objects."""
```

src/steamship/app/invocable.py
```diff
@@ -101,13 +103,25 @@ def __init__(self, client: Steamship = None, config: Dict[str, Any] = None):
             except (TypeError, FileNotFoundError):
                 secret_kwargs = {}
 
+        # The configuration for the Invocable is the union of:
```
EniasCailliau, line 106: Thanks for writing documentation!

tests/assets/apps/configurable_hello_world.py
```diff
@@ -1,11 +1,46 @@
...
+    # Package configuration is always defined by a Pydantic object which derives from
+    # the base `Config` object.
+    class HelloWorldConfig(Config):
+        greeting: str
+
+        # The casing of config variables is preserved exactly as the Steamship package author defines them.
+        # Snake case is permitted
+        snake_case_config: str
+        # Camel case is permitted
+        camelCaseConfig: str  # noqa: N815
```
EniasCailliau, line 17: I was wondering if the engine would accept snake_case without camelCase transformation. It is. Thanks for checking!

tests/assets/apps/configurable_hello_world.py
```diff
@@ -1,11 +1,46 @@
...
+        defaultConfig: str  # noqa: N815
+
+    # This is unnecessary, since the base `Invocable` declares it, but by re-declaring it here
+    # we get better type-hints from our editor.
+    config: HelloWorldConfig
+
     @post("greet")
     def greet(self, name: str = "Person") -> InvocableResponse:
-        greeting = self.config["greeting"]
+        greeting = self.config.greeting
```
EniasCailliau, line 27: Clean 😎

tests/assets/apps/configurable_hello_world.py
```diff
@@ -1,11 +1,46 @@
...
         return InvocableResponse(string=f"{greeting}, {name}")
 
+    @post("snake")
+    def snake(self) -> InvocableResponse:
+        return InvocableResponse(string=f"{self.config.snake_case_config}")
```
EniasCailliau, line 32:
```suggestion
        return InvocableResponse(string=self.config.snake_case_config)
```

tests/assets/apps/configurable_hello_world.py
```diff
@@ -1,11 +1,46 @@
...
+
+    @post("camel")
+    def camel(self) -> InvocableResponse:
+        return InvocableResponse(string=f"{self.config.camelCaseConfig}")
```
EniasCailliau, line 36:
```suggestion
        return InvocableResponse(string=self.config.camelCaseConfig)
```

tests/assets/apps/configurable_hello_world.py
```diff
@@ -1,11 +1,46 @@
...
+
+    @post("defaulted")
+    def defaulted(self) -> InvocableResponse:
+        return InvocableResponse(string=f"{self.config.defaultConfig}")
```
EniasCailliau, line 40:
```suggestion
        return InvocableResponse(string=self.config.defaultConfig)
```

tests/assets/apps/configurable_hello_world.py
```diff
@@ -1,11 +1,46 @@
...
+
+    def config_cls(self) -> Type[Config]:
+        return HelloWorld.HelloWorldConfig
```
EniasCailliau, line 43: Since this is a instance method, I would use self.HelloWorldConfig. 

We can also turn config_cls into a classmethod and use cls.HelloWorldConfig.

EniasCailliau (approved):
Just a few minor comments. LGTM! 🚀

tests/assets/apps/configurable_hello_world.py
eob: Ah, didn't realize you could use `self` like that. Will do!

PR #181, opened 2022-10-18
Add Task and TaskState to root package

EniasCailliau:
To me it makes sense to put these objects at the root of steamship


dkolas (approved):
👍

PR #184, opened 2022-10-19
[SHIP-427] rename space-> workspace in python

dkolas:


src/steamship/data/package_instance.py
```diff
@@ -71,11 +71,15 @@ def delete(self) -> PackageInstance:
         return self.client.post("package/instance/delete", payload=req, expect=PackageInstance)
 
     def load_missing_vals(self):
```
EniasCailliau, line 73:
```suggestion
    def load_missing_workspace_handle(self):
```

src/steamship/data/plugin.py
```diff
@@ -119,7 +119,7 @@ class PluginAdapterType(str, Enum):
 
 class PluginTargetType(str, Enum):
     file = "file"
```
EniasCailliau, line 121: Enum's are typically all caps. 
```suggestion
    FILE = "file"
```

The effects of this change should be limited to the python-client, in case it's not let's leave it.

src/steamship/data/workspace.py
```diff
@@ -13,15 +13,15 @@
 from steamship.base.response import Response as SteamshipResponse
 
 
-class Space(CamelModel):
+class Workspace(CamelModel):
```
EniasCailliau, line 16: Could you add the missing list static method? Listing workspaces sounds like an important feature.

tests/assets/plugins/taggers/plugin_third_party_trainable_tagger.py
```diff
@@ -180,7 +180,7 @@ def model_cls(self) -> Type[ThirdPartyModel]:
     def run_with_model(
         self, request: PluginRequest[BlockAndTagPluginInput], model: TestTrainableTaggerModel
     ) -> InvocableResponse[BlockAndTagPluginOutput]:
-        """Downloads the model file from the provided space"""
+        """Downloads the model file from the provided workspace"""
```
EniasCailliau, line 183: No need to change it in this PR but a heads-up. As soon as we enable docstring checking, this will get flagged for not ending with a period.

EniasCailliau (approved):
Only 2 optional requests. 

The codebase starts to read like a novel. I get a warm feeling inside when I'm going through our code now. Thanks for making this possible!

PR #185, opened 2022-10-20
Add task dependency support to client requests

eob:
Adds the `wait_on_tasks` kwarg to our base client (supported by the Engine) -- helpful to reveal in the client so that Doug can build the Zoom summarizer application.

```
client.call(wait_on_tasks: List[Task])
```

Will set HTTP headers such that the engine will schedule the work as:
1. A background task defaulting to the `.waiting` state, and
2. A task that first waits upon the other tasks provided

I've plumbed the kwarg up through to the `file.tag` and `file.blockify` calls so that end-level SDK users can use it without dipping down to the `client.post` style of invocation -- if that feels like adding a kwarg too aggressively LMK and we can back it out and just let the Zoom app use the base `client.post` call.

cc @douglas-reid


EniasCailliau (approved):
LGTM

dkolas (approved):
Looks clean to me!

PR #188, opened 2022-10-24
Docs feedback

dkolas:


docs/packages/index.rst
```diff
@@ -5,20 +5,13 @@ Packages
...
 Steamship Packages are full-lifecycle language AI libraries that run in the cloud.
 
-Production language AI typically introduces significant operational complexity to a company:
-piles of data, infrastructure, tasking, and integration must be managed alongside the "mere code" which
-interfaces with the actual AI being used.
-Steamship Packages solve this problem by encapsulating 100% of these concerns.
+Things to know about packages:
 
 - A package's code is versioned and run in the cloud.
 - A package's data is stored in a :ref:`workspace<Workspaces>` specific to each instance
```
douglas-reid, line 11: should we be clear about what the type of instance is here to enhance conceptual understanding?  Either by introducing `instance` in the bullet above, or inserting `package` before instance here, or both?

eob: Yeah I think that makes a ton of sense.

PR #189, opened 2022-10-24
Change output of plugininstance.train

dkolas:


src/steamship/data/plugin_instance.py
```diff
@@ -113,7 +113,7 @@ def delete(self) -> PluginInstance:
         req = DeleteRequest(id=self.id)
         return self.client.post("plugin/instance/delete", payload=req, expect=PluginInstance)
 
-    def train(self, training_request: TrainingParameterPluginInput) -> TrainPluginOutput:
+    def train(self, training_request: TrainingParameterPluginInput) -> Task[TrainPluginOutput]:
```
eob, line 116: Oh this is an interesting question. 

At the "client" level, we basically created a facade in which all methods return either `T` or `Task[T]`.

But we didn't go through and update the Plugin SDK for this as well, which we could have.

I'm pretty sure the response type for this ought to be:

```
Union[T, InvocableResponse[T]]
```

So in this case:

```
Union[TrainPluginOutput, InvocableResponse[TrainPluginOutput]]
```

If the training is still in progress, the plugin returns:

```
InvocableResponse(
  status=Task(stats=TaskState.running)
)
```

src/steamship/data/file.py
```diff
@@ -209,7 +209,7 @@ def tag(
         self,
         plugin_instance: str = None,
         wait_on_tasks: List[Task] = None,
-    ) -> Task[Tag]:
+    ) -> Task[Tag]:  # This actually is Task[TagResponse], right?
```
eob, line 212: Nice catch -- yes!

eob (changes requested):
I think the trainable one is actually something else -- lmk if the comment parses for you too!

src/steamship/data/plugin_instance.py
dkolas: But this IS the client side of the facade, not what the plugin SDK can actually return?  This how the client calls train, not how the plugin responds.

eob: Oh snap, I missed that. Then yes -- def `Task[T]`!!

src/steamship/data/file.py
dkolas: @EniasCailliau It doesn't look like I can just swap it to TagResponse because of the circular dependencies.  Any suggestions?

dkolas:
Included in #192

PR #191, opened 2022-10-25
Error to warning in client

dkolas:
This PR changes another location where a server-side error is logged as an error within the Python SDK.  Unfortunately, this means that any call to get on an object that doesn't exist is logged as error, even if expected.  Changing this to warning here cleans up the logs - if it's really error, it can be logged as such engine side.

Depends on #189


dkolas:
Included in #192

PR #192, opened 2022-10-26
Partially solve circular dependencies

EniasCailliau:
After fighting with the code, I decided to go for a pragmatic solution that only moves a few files around. 

Here are 2 ways I used to fix circular dependencies without moving too many files: 

If the circular dependency only concerns `typing`, we can fix it with this hack: 

```
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from steamship.data.operations.tagger import TagResponse
```

This import will only be run when Mypy is type-checking. This occurs after the whole library is loaded, so the circular dependency is not a problem. 

NOTE: This does not work for Pydantic typing. If you want to fix a circular dependency for a Pydantic-defined field, you can use the following trick:

```
from __future__ import annotations

# YOUR CODE HERE 

from .client import Client  # noqa: E402

Task.update_forward_refs()
TaskComment.update_forward_refs()
```

This uses future annotations which lazily load their dependencies. In past Python iterations we used to use future annotations with double quotes "Client", using the __future__ import this is no longer necessary. 

If you use the circular dependency in your code there is sadly no solution. Today, we only have 1 circular dependency we have to solve at runtime between `File` and `TagRequest`, `TagResponse`.


dkolas (approved):
Looks like a solid incremental improvement to me.

PR #194, opened 2022-10-26
File.list, get return types, and fix for test harness

dkolas:


src/steamship/base/client.py
```diff
@@ -45,7 +45,7 @@ def __init__(
         config_file: str = None,
         config: Configuration = None,
         trust_workspace_config: bool = False,  # For use by lambda_handler; don't fetch the workspace
-        **kwargs,
+        **z,
```
eob, line 48: I think this might be a typo?

eob (changes requested):
LGTM! Potential typo in client.py -- thanks for bringing this back!

src/steamship/base/client.py
dkolas: Fixed, thanks!

PR #202, opened 2022-11-02
config: make empty Config behavior default for Invocables

douglas-reid:
This PR refactors the `config_cls` method from `Invocable` so that it is no longer abstract. This will provide a default behavior for plugins and packages, eliminating boilerplate when configuration is not needed.

Packages and plugins can still override the behavior by defining their own `config_cls` implementation.

If this leads to issues with users not properly providing configuration due to API mishaps, we can add a `Configurable` ABC for use with invocables. This might work something like:

```python
class Configurable(ABC):

  @abstractmethod
  def config_cls(self) -> Type[Config]:
    raise NotImplementedError()


class MyPackage(Configurable, PackageService):

  class MyPackageConfig(Config):
    foo: str


  def config_cls(self) -> Type[Config]:
    return MyPackageConfig
```

### DIFF
--- a/docs/packages/cookbook/adding-configuration.rst
+++ b/docs/packages/cookbook/adding-configuration.rst
@@ -1,0 +1,150 @@
+Making a package or plugin user-configurable
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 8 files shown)


eob (approved):
LGTM!

I just ran tests on localhost and they were all green.

Nice catch on the requirements.txt!

PR #210, opened 2022-11-03
add: link to Steamship plugin registry doc page

douglas-reid:
This might help people quickly jump back and forth between the docs and what's immediately available to play with.

If we'd prefer to link to the `Packages` page of the same doc site instead, that'd be OK too.


eob (approved):
Totally! LGTM!

PR #224, opened 2022-11-05
Add a `list` method to the KVStore class.

eob:
Was goofing off in the package layer and found utility in this:

- The more people build real things in packages, the more they'll need the ability to store app data
- This is a small change that adds a way to treat the KV store as a list of KV Pairs that you can set; e.g. dynamic user settings for the package or some external app that's using Steamship as a data store.

### DIFF
--- a/src/steamship/utils/kv_store.py
+++ b/src/steamship/utils/kv_store.py
@@ -1,8 +1,10 @@
 """A simple key-value store implemented atop Files and Tags."""
 
-from typing import Any, Dict, Optional
+from typing import Any, Dict, List, Optional, Tuple
 
 from steamship import Block, File, Steamship, Tag
+
+KV_STORE_MARKER = "__init__"
 
 
 class KeyValueStore:
@@ -47,13 +49,13 @@ class KeyValueStore:
             return File.create(
                 self.client,
                 blocks=[Block.CreateRequest(text="")],
-                tags=[Tag.CreateRequest(kind=self.store_identifier, name="__init__")],
+                tags=[Tag.CreateRequest(kind=self.store_identifier, name=KV_STORE_MARKER)],
             )
         else:
             return status_files[0]
 
     def get(self, key: str) -> Optional[Dict]:
-        """Gets the value represented by `key`."""
+        """Get the value represented by `key`."""
         file = self._get_file()
 
         if file is None:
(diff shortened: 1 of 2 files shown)


src/steamship/utils/kv_store.py
```diff
@@ -90,8 +92,24 @@ def set(self, key: str, value: Dict[str, Any]):
         req = Tag.CreateRequest(file_id=file.id, kind=self.store_identifier, name=key, value=value)
         return self.client.post("tag/create", req, expect=Tag)
 
+    def list(self, filter_keys: Optional[List[str]] = None) -> List[Tuple[str, Dict[str, Any]]]:
```
EniasCailliau, line 95: I would call it .items() similar to how you would call dict.items()

EniasCailliau:
```suggestion
    def list(self, filter_keys: Optional[List[str]] = None) -> List[Tuple[str, Any]]:
```

I guess we also support strings, bools, ints, and floats as values. If yes, then we cover `Any`.

src/steamship/utils/kv_store.py
```diff
@@ -1,9 +1,11 @@
...
 
-from typing import Any, Dict, Optional
+from typing import Any, Dict, List, Optional, Tuple
 
 from steamship import Block, File, Steamship, Tag
 
+KV_STORE_MARKER = "__init__"
+
 
 class KeyValueStore:
```
EniasCailliau, line 10: Just sharing: 

In case you want to support accessing the kv store like a dict (`my_kv["key"] = my_value` and `my_kv["key"]`) we only need to implement two functions: 

```
    def __getitem__(self, key):
    def __setitem__(self, key, value):
```

EniasCailliau: Bonus for supporting `del my_kv["key"]`: 
```
    def __delitem__(self, key):
```

EniasCailliau (dismissed):
Added a few comments. Nothing that blocks us from merging! LGTM! 🚀

src/steamship/utils/kv_store.py
eob: Oh that's nice. I'm a *bit* hesitant to do it here since the get/set/delete involves HTTP calls over the network and for some reason having builtin `[]` operators backed by an HTTP call feels funny -- but appreciate the tip.

src/steamship/utils/kv_store.py
eob: At the moment it's hard-coded to only support `Dict` objects so the `Dict[str, Any]` is actually the correct type here unless we relax that requirement

eob: Updated to `items`

eob:
tests green against staging; merging

PR #229, opened 2022-11-14
Fix SHIP-479

EniasCailliau:
https://linear.app/steamship/issue/SHIP-479/tagcreate-stringifies-value-improperly

### DIFF
--- a/src/steamship/data/tags/tag.py
+++ b/src/steamship/data/tags/tag.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import json
 from typing import Any, Dict, List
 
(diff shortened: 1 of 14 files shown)


src/steamship/data/tags/tag.py
```diff
@@ -57,11 +56,8 @@ def create(
         name: str = None,
         start_idx: int = None,
         end_idx: int = None,
-        value: Any = None,
+        value: Dict[str, Any] = None,
```
eob, line 59: Good catch

eob (approved):
LGTM!

PR #231, opened 2022-11-17
[SHIP-453] Provide safe handler for invocables so that import issues can be caught and logged

dkolas:


src/steamship/invocable/entrypoint.py
```diff
@@ -0,0 +1,19 @@
+"""
+This class is necessary to be able to please the entrypoints of both localstack and AWS.
+
+If we set the entrypoint directly to steamship.invocable.safe_handler (imported in the init
```
EniasCailliau, line 4: Wow! Reads like it has been a struggle to figure things out. Thanks for putting in the effort!

src/steamship/invocable/lambda_handler.py
```diff
@@ -27,194 +28,244 @@ def encode_exception(obj):
...
-            return InvocableResponse.error(
-                code=HTTPStatus.INTERNAL_SERVER_ERROR,
-                prefix=error_prefix,
-                message="Unable to initialize package/plugin.",
-                exception=ex,
-            )
+    try:
+        invocable = invocable_cls_func()(client=client, config=request.invocation.config)
+    except SteamshipError as se:
+        return InvocableResponse.from_obj(se)
```
EniasCailliau, line 60: Do we want to log this exception too? 

```suggestion
        logging.exception(ex)
        return InvocableResponse.from_obj(se)
```

src/steamship/invocable/lambda_handler.py
```diff
@@ -27,194 +28,244 @@ def encode_exception(obj):
...
+    return lambda event, context=None: handler(bound_internal_handler, event, context)
+
 
-            logging.info(f"Got signed url for writing: {signed_url}")
+def safely_find_invocable_class() -> Type[Invocable]:
+    """
+    Safely find the invocable class within invocable code.
+    """
+    try:
+        import api
```
EniasCailliau, line 237: Does this mean we always need to implement our packages and plugins in `src/apy.py`?

src/steamship/invocable/lambda_handler.py
```diff
@@ -27,194 +28,244 @@ def encode_exception(obj):
...
-            upload_to_signed_url(signed_url, data)
+    invocable_classes = []
+    for element in [getattr(api, x) for x in dir(api)]:
+        if inspect.isclass(element):
+            # Using names and not issubclass(element, Invocable) because latter was returning false?
+            superclass_names = [c.__name__ for c in inspect.getmro(element)]
+            if "Invocable" in superclass_names and element.__module__ == "api":
+                invocable_classes.append(element)
+    if len(invocable_classes) == 0:
+        raise SteamshipError(message="Could not find package or plugin class in api.py.")
```
EniasCailliau, line 252: Would this help the user to debug without reading the docs?

```suggestion
        raise SteamshipError(message="Could not find package or plugin class in api.py. Define your package or plugin by subclassing from PluginService or PackageService.")
```

EniasCailliau (approved):
Looking good! Thanks for fixing this! Added a few comment you can just accept or discard :)

src/steamship/invocable/entrypoint.py
dkolas: Yeah, that bit was painful, and wanted the reasoning to live on for posterity, so that we didn't forget about it later and mistakenly see it as an opportunity for cleanup.

src/steamship/invocable/lambda_handler.py
dkolas: Yes.  In a later PR I improved the error message to make that clearer.

src/steamship/invocable/lambda_handler.py
dkolas: Yes, great suggestion.  I just slipped it into my next PR.

src/steamship/invocable/lambda_handler.py
dkolas: Yup.  Added it to next PR.

PR #232, opened 2022-11-21
Add a text field to `Tag`

eob:
I'm going to try splitting up the embedding refactor into several smaller commits.

All of our conversations around design involved the `Tag` field gaining a `text` field so that it could be stored/retreived/worked-with as an atomic object separate from the file from which it came.

This PR adds that field.

### DIFF
--- a/src/steamship/data/tags/tag.py
+++ b/src/steamship/data/tags/tag.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Optional
 
 from pydantic import Field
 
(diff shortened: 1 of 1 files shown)


src/steamship/data/tags/tag.py
```diff
@@ -15,17 +15,40 @@ class TagQueryRequest(Request):
...
-    start_idx: int = None  # w/r/t block.text. None means 0 if blockId is not None
-    end_idx: int = None  # w/r/t block.text. None means -1 if blockId is not None
+
+    # The name of tag. See the ``DocTag``, ``TokenTag``, etc enum classes for suggestions.
+    name: Optional[str] = None  # E.g. person
+
+    # The value payload of the tag. Always a JSON-style object.
+    value: Optional[Dict[str, Any]] = None
+
+    # Start-index of the text covered by the tag, using python slice semantics.
```
dkolas, line 39: I don't know what you meant to convey with 'using python slice semantics', but I think that comment should be removed.  You can't, for example, use negative indices to imply an offset from the back, nor does omitting just start or end mean the beginning/end, so I don't think these use slice semantics at all.

src/steamship/data/tags/tag.py
```diff
@@ -15,17 +15,40 @@ class TagQueryRequest(Request):
...
+    start_idx: Optional[int] = None
+
+    # End-index of the text covered by the tag, using python slice semantics.
+    end_idx: Optional[int] = None
+
+    # The text covered by the tag. Note that text will not always be materialized into the tag object
+    # itself; you may have to fetch it with file.text[tag.start_idx:tag.end_idx]
+    text: Optional[str] = None
```
dkolas, line 47: You may also want to note that setting /overwriting this field will have no effect.

dkolas (approved):
Looks good! A couple comments about your comments :)

src/steamship/data/tags/tag.py
EniasCailliau: I would highlight that we're using a character index (vs for example a token index)

src/steamship/data/tags/tag.py
EniasCailliau: What if we make text a read-only property that cannot be set by the user? 

That would avoid the weird case where the user ads text that does not match the highlighted span. 

For efficiency we could instruct the engine to include or exclude the text field when returning tags.

EniasCailliau (commented):
Maybe we can make text read-only? If people can write anything in the text field they will abuse it sooner than later.

src/steamship/data/tags/tag.py
eob: Thanks for the catch -- I'll update to mention character index and just be specific about start inclusivity and end exclusivity

src/steamship/data/tags/tag.py
eob: I think a read-only property might work really well for that. Going to wait on that since the PRs after this one are going to use the text field heavily & that might provide some design information too

PR #233, opened 2022-11-21
Introduce EmbeddingIndexPluginInstance pseudo PluginInstance

eob:
This PR does tree thigns:

Introduces a EmbeddingIndexPluginInstance class which wraps the following in the interface of a PluginInstance:
An EmbeddingIndex
An Embedder (Tagger)
Introduce a shim in Steamship.use which enables the creation of this new pseudo-PluginInstance via the regular Steamship.use and steamship.use pathways.

Convert nearly all tests to use this new steamship.use_plugin("embedding-index") style of invocation.

This PR does not:

Update how embedding indices are interacted with. That is now done through indirection via the shim. That will be the next PR.

Depends upon #232.

### DIFF
--- a/src/steamship/data/plugin/index_plugin_instance.py
+++ b/src/steamship/data/plugin/index_plugin_instance.py
@@ -1,0 +1,80 @@
+from typing import Any, Dict, Optional
+
+from pydantic import Field
+
+from steamship import SteamshipError
+from steamship.base.client import Client
+from steamship.base.model import CamelModel
+from steamship.data.embeddings import EmbeddingIndex
+from steamship.data.plugin.plugin_instance import PluginInstance
+
+
+class EmbedderInvocation(CamelModel):
+    """The parameters capable of creating/fetching an Embedder (Tagger) Plugin Instance."""
+
+    plugin_handle: str
+    instance_handle: Optional[str] = None
+    config: Optional[Dict[str, Any]] = None
+    version: Optional[str] = None
+    fetch_if_exists: bool = True
+
+
+# Hard coded list of plugin handles that the Steamship client will use to create
+# instance of this object instead of a normal PluginInstance.
+SHIMMED_INDEX_PLUGIN_HANDLES = ["embedding-index"]
+
+
+class EmbeddingIndexPluginInstance(PluginInstance):
+    """A persistent, read-optimized index over embeddings.
+
+    This is currently implemented as an object which behaves like a PluginInstance even though
+    it isn't from an implementation perspective on the back-end.
+    """
+
+    client: Client = Field(None, exclude=True)
+    embedder: PluginInstance = Field(None, exclude=True)
+    index: EmbeddingIndex = Field(None, exclude=True)
+
+    def delete(self):
+        return self.index.delete()
+
+    @staticmethod
+    def create(
+        client: Any,
+        plugin_id: str = None,
+        plugin_handle: str = None,
+        plugin_version_id: str = None,
+        plugin_version_handle: str = None,
+        handle: str = None,
+        fetch_if_exists: bool = True,
+        config: Dict[str, Any] = None,
+    ) -> "EmbeddingIndexPluginInstance":
+        """Create a class that simulates an embedding index re-implemented as a PluginInstance."""
+        if plugin_handle not in SHIMMED_INDEX_PLUGIN_HANDLES:
+            raise SteamshipError(message=f"No Embedding Index of type {plugin_handle} was found.")
+
+        # Perform a manual config validation check since the configuration isn't actually being sent up to the Engine.
+        # In this case, an embedding index has special behavior which is to instantiate/fetch an Embedder that it can use.
+        if "embedder" not in config:
+            raise SteamshipError(
+                message="Config key missing. Please include a field named `embedder` with type `EmbedderInvocation`."
+            )
+
+        # Just for pydantic validation.
+        embedder_invocation = EmbedderInvocation.parse_obj(config.get("embedder"))
+
+        # Create the embedder
+        embedder = client.use_plugin(**embedder_invocation.dict())
+
+        # Create the index
+        index = EmbeddingIndex.create(
+            client=client,
+            handle=handle,
+            plugin_instance=embedder.handle,
+            fetch_if_exists=fetch_if_exists,
+        )
+
+        # Now return the plugin wrapper
+        return EmbeddingIndexPluginInstance(
+            id=index.id, handle=index.handle, index=index, embedder=embedder
+        )
--- a/src/steamship/data/plugin/index_plugin_instance.py
+++ b/src/steamship/data/plugin/index_plugin_instance.py
@@ -31,9 +31,9 @@ class EmbeddingIndexPluginInstance(PluginInstance):
     it isn't from an implementation perspective on the back-end.
     """
 
-    client: Client = Field(None, exclude=True)
-    embedder: PluginInstance = Field(None, exclude=True)
-    index: EmbeddingIndex = Field(None, exclude=True)
+    client: Client = Field(..., exclude=True)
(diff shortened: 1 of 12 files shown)


src/steamship/data/plugin/index_plugin_instance.py
```diff
@@ -0,0 +1,80 @@
...
+        embedder_invocation = EmbedderInvocation.parse_obj(config.get("embedder"))
+
+        # Create the embedder
+        embedder = client.use_plugin(**embedder_invocation.dict())
+
+        # Create the index
+        index = EmbeddingIndex.create(
+            client=client,
+            handle=handle,
+            plugin_instance=embedder.handle,
```
EniasCailliau, line 73:
```suggestion
            embedder_plugin_instance=embedder.handle,
```

Am I correct we only accept embedders here? It's not enforced by type, but we can hint at it.

src/steamship/data/plugin/index_plugin_instance.py
```diff
@@ -0,0 +1,80 @@
...
+            fetch_if_exists=fetch_if_exists,
+        )
+
+        # Now return the plugin wrapper
+        return EmbeddingIndexPluginInstance(
+            id=index.id, handle=index.handle, index=index, embedder=embedder
```
EniasCailliau, line 79:
```suggestion
            id=index.id, index_handle=index.handle, index=index, embedder=embedder
```

Q: Why do we need both `handle` and `index`? If the `EmbeddingIndexPluginInstance` has access to the index object it can fetch the handle itself.

src/steamship/data/tags/tag.py
```diff
@@ -15,17 +15,40 @@ class TagQueryRequest(Request):
 
 
 class Tag(CamelModel):
+    # Steamship client.
```
EniasCailliau, line 18: See #232 for comments

tests/steamship_tests/utils/random.py
```diff
@@ -19,9 +19,11 @@ def random_name() -> str:
 
 @contextlib.contextmanager
 def random_index(steamship: Steamship, plugin_instance: str) -> EmbeddingIndex:
-    index = EmbeddingIndex.create(
-        client=steamship,
-        plugin_instance=plugin_instance,
+    index = steamship.use_plugin(
+        "embedding-index",
+        random_name(),
+        config={"embedder": {"plugin_handle": plugin_instance, "fetch_if_exists": True}},
```
EniasCailliau, line 25: Does it make sense to ask the user to instantiate the embedder themselves and pass the plugin_instance_handle of the embedder? 

That would result in 2 use_plugin calls but would reduce the interface of the config object and potentially simplify the logic of the embedding-index.

src/steamship/data/plugin/index_plugin_instance.py
```diff
@@ -0,0 +1,80 @@
...
+
+        # Perform a manual config validation check since the configuration isn't actually being sent up to the Engine.
+        # In this case, an embedding index has special behavior which is to instantiate/fetch an Embedder that it can use.
+        if "embedder" not in config:
+            raise SteamshipError(
+                message="Config key missing. Please include a field named `embedder` with type `EmbedderInvocation`."
+            )
+
+        # Just for pydantic validation.
+        embedder_invocation = EmbedderInvocation.parse_obj(config.get("embedder"))
```
EniasCailliau, line 64:
```suggestion
        embedder_invocation = EmbedderInvocation.parse_obj(config["embedder"])
```

We know "embedder" is present.

tests/steamship_tests/client/operations/test_embedding_index.py
```diff
@@ -1,65 +1,75 @@
...
     # Duplicate creation should fail with fetch_if_exists=False
     with pytest.raises(SteamshipError):
-        _ = create_index(
-            client=client,
-            handle=index.handle,
-            plugin_instance=plugin_instance.handle,
+        index = client.use_plugin(
+            "embedding-index",
+            index_handle,
+            config={"embedder": {"plugin_handle": _TEST_EMBEDDER, "fetch_if_exists": True}},
```
EniasCailliau, line 30: This config object feels heavy to me. Can we just pass a embedder_instance_handle we retrieve using `use_plugin`?

src/steamship/data/plugin/index_plugin_instance.py
```diff
@@ -0,0 +1,80 @@
...
+    version: Optional[str] = None
+    fetch_if_exists: bool = True
+
+
+# Hard coded list of plugin handles that the Steamship client will use to create
+# instance of this object instead of a normal PluginInstance.
+SHIMMED_INDEX_PLUGIN_HANDLES = ["embedding-index"]
+
+
+class EmbeddingIndexPluginInstance(PluginInstance):
```
EniasCailliau, line 27: Would it be a good idea to re-define the methods of `EmbeddingIndex` in this class? 

For example we would add 

```
def search(
        self,
        query: Union[str, List[str]],
        k: int = 1,
        include_metadata: bool = False,
    ) 
    self.index.search(query, k, include_metadata)
```

Benefit: I won't have to call `index.index` to operate with the index, instead I'm going to use the pluginInstance directly, similar to how we define .blockify on the level of the Blockifier

EniasCailliau: Ignore what I said above. After re-reading the description, I realize you'll address this in another PR

EniasCailliau (approved):
I would rename a few variables to be more self-documenting. LGTM!

src/steamship/data/plugin/index_plugin_instance.py
```diff
@@ -0,0 +1,80 @@
...
+    """A persistent, read-optimized index over embeddings.
+
+    This is currently implemented as an object which behaves like a PluginInstance even though
+    it isn't from an implementation perspective on the back-end.
+    """
+
+    client: Client = Field(None, exclude=True)
```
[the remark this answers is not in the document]
eob: TIL -- I had no idea that was a valid operator

src/steamship/data/plugin/index_plugin_instance.py
eob: Agree! Done!

src/steamship/data/plugin/index_plugin_instance.py
eob: I think this might evolve as these PRs progress.

The answer, at the moment is:
- index is necessary cause that's the thing we're wrapping
- id & handle are useful so this can "pretend" to be a PluginInstance -- an object which other code expects to had an id and handle field
- my attempts to make this a @property field all failed; I suspect because it interacted strangely with Pydantic's base class

src/steamship/data/tags/tag.py
eob: 👍 will address there

tests/steamship_tests/utils/random.py
eob: If you're comfortable with it -- I'm going to push forward on this and return to this comment in a future PR. 

I think both have some advantages, and it'll be easy to experiment with that in a way we can all take a look at a PRs downstream from this.

tests/steamship_tests/client/operations/test_embedding_index.py
eob: (agree! see other comment-- will be easier for us to make a call when we can experiment w/ both side by side; will show in future pr)

src/steamship/data/plugin/index_plugin_instance.py
eob: 👍 Your comment is right-on though.

And yeah, I'm trying to do this across several PRs in an attempt to make each change easy to see from you & Dave's perspective since this is a pretty big shift in surface area. Next pr begins to do exactly what you suggest

PR #234, opened 2022-11-21
Migrate to new `insert` and `search` endpoints for Embedding Indices

eob:
This PR:

* Converts all `index.insert` operations to ones that use the `Tag` object
* Converts all `index.search` operations to ones that return `Tag` objects
* removes testing for features that won't be supported in the new pluginInstance style interface
  - doing multiple searches at a time
  - listing the contents of an embedding index
  
Note:

* It looks like we'll need a very small change on the Engine to support this; the client & engine appear out of sync on data model with this -- DBSimdexRecord will need a tagId field, and the API endpoint will need to report it. I can make but wanted to flag for you @dkolas

### DIFF
--- a/src/steamship/data/block.py
+++ b/src/steamship/data/block.py
@@ -79,10 +79,19 @@ class Block(CamelModel):
         )
         return res
 
-    def index(self, plugin_instance: Any = None):
+    def index(self, embedding_plugin_instance: Any = None):
         """Index this block."""
-        tags = [Tag(text=self.text, file_id=self.id, block_id=self.id, kind="block")]
-        return plugin_instance.insert(tags)
+        tags = [
+            Tag(
+                text=self.text,
+                file_id=self.file_id,
+                block_id=self.id,
+                kind="block",
+                start_idx=0,
+                end_idx=len(self.text),
+            )
+        ]
+        return embedding_plugin_instance.insert(tags)
 
 
 class BlockQueryResponse(Response):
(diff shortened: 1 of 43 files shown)


src/steamship/data/block.py
```diff
@@ -79,6 +79,11 @@ def query(
         )
         return res
 
+    def index(self, plugin_instance: Any = None):
+        """Index this block."""
+        tags = [Tag(text=self.text, file_id=self.id, block_id=self.id, kind="block")]
```
EniasCailliau, line 84: Flagging this to make sure this is intended: 

Do we want to assign the block_id to the file_id?

src/steamship/data/block.py
```diff
@@ -79,6 +79,11 @@ def query(
         )
         return res
 
+    def index(self, plugin_instance: Any = None):
```
EniasCailliau, line 82:
```suggestion
    def index(self, embedding_plugin_instance: EmbeddingIndex = None):
```

This only works for embedding_plugin_instances right?

EniasCailliau (approved):
LGTM

src/steamship/data/block.py
eob: Yikes nice catch @EniasCailliau !

src/steamship/data/block.py
eob: Yes -- will update w/ this suggestion. Thanks!

eob:
Tests green against staging so merging.

PR #235, opened 2022-11-22
Add a `config` object on method registration, to be returned in __dir__

eob:
There's value in exploring the possibility of "public" package endpoints:

* Certain operations, like web hooks, are not configurable with an `Authentication` header. The Telegram bot api is an example of this.
* Hosting a simple web UI from within a package 
* Hosting an endpoint that a web form or other unauth'ed user could post to

One way to support this would be via an annotation, such as:

```
@post("/add-data", public=True)
def add_data(self..)
```

This PR transforms all primitive-valued kwargs after the `path` in the @get and @post comments into a `config` object which is returned in the __dir__ object.

This permits us to experiment with support for things like public endpoints without having to fiddle with custom client versions; we just put it in the kwargs and have the engine react to it once the AppVersion is deployed.

### DIFF
--- a/src/steamship/invocable/invocable.py
+++ b/src/steamship/invocable/invocable.py
@@ -62,14 +62,8 @@ def endpoint(verb: str = None, path: str = None, **kwargs):
         # Build a dictionary of String->Primitive Types to pass back with endpoint
         # This enables the Engine to add support for features like public=True, etc, without the Client changing.
         config: Dict[str, Union[str, bool, int, float]] = {}
-        for key in kwargs:
-            val = kwargs[key]
-            if (
-                isinstance(val, str)
-                or isinstance(val, bool)
-                or isinstance(val, int)
-                or isinstance(val, float)
-            ):
+        for key, val in kwargs.items():
+            if isinstance(val, (str, bool, int, float)):
                 config[key] = val
 
         wrap.__path__ = path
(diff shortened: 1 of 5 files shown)


dkolas (approved):
I think this is fine as an intermediate step, but I'd love to see the extra args made explicit once we know we're going to keep them.

src/steamship/invocable/invocable.py
```diff
@@ -60,8 +59,23 @@ def decorator(function):
         def wrap(self, *args, **kwargs):
             return function(self, *args, **kwargs)
 
+        # Build a dictionary of String->Primitive Types to pass back with endpoint
+        # This enables the Engine to add support for features like public=True, etc, without the Client changing.
+        config: Dict[str, Union[str, bool, int, float]] = {}
+        for key in kwargs:
```
EniasCailliau, line 65:
```suggestion
        for key, value in kwargs.items():
```

src/steamship/invocable/invocable.py
```diff
@@ -60,8 +59,23 @@ def decorator(function):
...
+            val = kwargs[key]
```
EniasCailliau, line 66:
```suggestion
```

src/steamship/invocable/invocable.py
```diff
@@ -60,8 +59,23 @@ def decorator(function):
...
+            if (
+                isinstance(val, str)
```
EniasCailliau, line 68:
```suggestion
                isinstance(val, (str, bool, int, float))
```

EniasCailliau (dismissed):
Have 2 minor simplifications we might want to consider - LGTM!

eob:
Thanks! Also agree these should migrate to named args

eob:
Green against `staging`

PR #236, opened 2022-11-22
Fix two bugs: File.create when both `content` and `tags` are present, and multipart Mime-posting of nested objects

eob:
There are two ways to create a file:
* Provide Block JSON (blocks and tags)
* Provide content

There's a third way which should be supported:
* Provide content along with file-tags

But when `File.create` was called with both `content` and `tags`, the client would incorrectly set the upload type to BLOCK_JSON, which would result in an empty file (but one that contained tags).

## Change

This PR changes the behavior of File.create to make the upload type decision based only on the presence/absence of `blocks` or `content` --- not the `tags` property. 

This should allow for files to be created with `File.create(content=..., tags=...)`

## .. But then there's a new problem

Fixing this bug uncovers a new bug lurking in our multi-part POST handling. It's probably been there all along and we just never noticed.

When running the new test:

```
test_file::test_file_upload_with_content_and_tags
```

We get the following failure:

```
E           steamship.base.error.SteamshipError: [AbortError]
E           Value of type 'MultipartPart' required for key 'tags'.
```

Here's how we encode the file data:

```
            ("file-part", content, "multipart/form-data")
```

And here's how we encode the fields on the object:

```
                    result[key] = (None, json.dumps(val), "application/json")
```

What I suspect is happening is that the way we're encoding the POST causes:
- Swift to correctly understand the file bytes;
- Swift to correctly understand individual primitive types (bool, string, int) on the uploaded file. I added the `handle` parameter to File.CreateRequest and File.create to test & verify that primitive types worked in a multi-part request, and it appears they do.
- Swift to not understand complex types (tags, which is an array of Tag.CreateRequest encoded as JSON

This might be a Swift problem -- perhaps Vapor (swift web framework) doesn't handle unpacking multi-part requests well into a nested object.

# Update

The Mime problem is solved! Vapor just wants a very specific encoding.

cc @douglas-reid

### DIFF
--- a/src/steamship/base/client.py
+++ b/src/steamship/base/client.py
@@ -5,7 +5,7 @@ import logging
 import typing
 from abc import ABC
 from inspect import isclass
-from typing import Any, Dict, List, Type, TypeVar, Union
+from typing import Any, List, Type, TypeVar, Union
 
 import inflection
 from pydantic import BaseModel, PrivateAttr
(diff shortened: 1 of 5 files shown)


EniasCailliau (approved):
Wow, that format is complex. Nice solution, LGTM!

eob:
Tests pass against staging; merging

PR #237, opened 2022-11-22
Updates for engine-side default safe handler; localstack version upgrade

dkolas:
Updates for engine-side default safe handler; localstack version upgrade.  Depends on https://github.com/nludb/nludb/pull/376

- Add bad package engine test asset
- Deprecate old lambda handler 
- Update signed url rewriting behavior
- Remove special casing for using safe handler

(Tests will fail until above engine branch is deployed)

### DIFF
--- a/scripts/create_engine_test_assets.py
+++ b/scripts/create_engine_test_assets.py
@@ -17,6 +17,7 @@ def main():
         (PACKAGES_PATH / "demo_package.py", "demo-package.zip"),
         (PACKAGES_PATH / "configurable_hello_world.py", "configurable_hello_world.zip"),
         (PLUGINS_PATH / "blockifiers" / "blockifier.py", "dummy_blockifier.zip"),
+        (PACKAGES_PATH / "bad_package.py", "bad-package.zip"),
     ]
 
     for path, output in assets_to_build:
--- a/src/steamship/invocable/lambda_handler.py
+++ b/src/steamship/invocable/lambda_handler.py
@@ -1,3 +1,4 @@
+import inspect
 import json
 import logging
 import sys
@@ -13,8 +14,7 @@ from steamship import Configuration
 from steamship.base import SteamshipError
 from steamship.client import Steamship
 from steamship.data.workspace import SignedUrl
-from steamship.invocable import InvocableRequest, InvocableResponse, InvocationContext
-from steamship.invocable.invocable import Invocable
+from steamship.invocable import Invocable, InvocableRequest, InvocableResponse, InvocationContext
 from steamship.utils.signed_urls import upload_to_signed_url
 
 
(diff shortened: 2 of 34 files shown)


scripts/create_engine_test_assets.py
```diff
@@ -17,6 +17,7 @@ def main():
         (PACKAGES_PATH / "demo_package.py", "demo-package.zip"),
         (PACKAGES_PATH / "configurable_hello_world.py", "configurable_hello_world.zip"),
         (PLUGINS_PATH / "blockifiers" / "blockifier.py", "dummy_blockifier.zip"),
+        (PACKAGES_PATH / "bad_package.py", "bad-package.zip"),
```
eob, line 20: oh nice. @douglas-reid could use this too to test deploying something that shouldn't work

src/steamship/invocable/lambda_handler.py
```diff
@@ -218,15 +220,19 @@ def handler(internal_handler, event: Dict, _: Dict = None) -> dict:  # noqa: C90
 
 
 def create_handler(invocable_cls: Type[Invocable]):
-    """Legacy wrapper function for a Steamship invocable within an AWS Lambda function. Called by code within a
+    """Deprecated wrapper function for a Steamship invocable within an AWS Lambda function. Called by code within a
```
eob, line 223: Ah, so we'll want to update our templates

src/steamship/utils/signed_urls.py
```diff
@@ -90,36 +89,36 @@ def upload_to_signed_url(url: str, _bytes: Optional[bytes] = None, filepath: Opt
...
-        # The key, and a selected subset of the former query args, become multi-part mime data
-        key = "/".join(path_parts[1:])
-        data = {
-            "key": key,
-            "AWSAccessKeyId": params["X-Amz-Credential"].split("/")[0],
-            "signature": params["X-Amz-Signature"],
-        }
-        files = {"file": _bytes}
-        http_response = requests.post(newurl, data=data, files=files)
+    # parsed_url = urllib.parse.urlparse(url)
```
eob, line 92: This is very exciting. Thinking is to leave it commented until we're 100% sure this all works?

tests/assets/packages/bad_package.py
```diff
@@ -0,0 +1,12 @@
...
+from steamship import Steamship
+from steamship.invocable import Invocable
+
+
+class BadPackage(Invocable):
+    """Bad Package should not start when deployed."""
+
+    def __init__(self, client: Steamship = None, config: Dict[str, Any] = None):
+        super().__init__(client, config)
+        raise RuntimeError("this should fail immediately")
```
eob, line 12: world's worst package 📦

tests/steamship_tests/utils/fixtures.py
```diff
@@ -62,7 +62,7 @@ def _test_something(invocable_handler):
     new_client = get_steamship_client(workspace=workspace_handle)
 
     def handle(verb: str, invocation_path: str, arguments: Optional[dict] = None) -> dict:
-        _handler = _create_handler(invocable)
+        _handler = _create_handler(known_invocable_for_testing=invocable)
```
eob, line 65: ah that's why it was necessary. cool

eob (approved):
LGTM!!

src/steamship/invocable/lambda_handler.py
dkolas: Yes.  I was able to make it so that if you don't update the code, things should still work, but we should do a pass to remove it from templates (one-line delete)

scripts/create_engine_test_assets.py
dkolas: I think @douglas-reid created it - I just committed the test asset so it could be rebuilt :)

src/steamship/utils/signed_urls.py
dkolas: Actually no, I meant to remove this, thanks for spotting :)

PR #239, opened 2022-11-30
Adds an `invoke_later` method to invocables

eob:
**Tests green against `engine@main` but will fail on prod/staging since they're not up to date**

This adds a new method to the `Invocable` class called `invoke_later` which schedules an invocation against that Invocable's own method call. This enables sequences of long-running operations to be chained together by a package.

In order to implement this, a new variable needed to be passed into the Invocable -- the InvocationContext. That is now a piece of the base-class init method and has to be plumbed down by any plugins/packages that choose to implement their own custom init.

### DIFF
--- a/src/steamship/base/client.py
+++ b/src/steamship/base/client.py
@@ -227,14 +227,14 @@ class Client(CamelModel, ABC):
 
         return f"{base}/{operation}"
 
-    def _headers(
+    def _headers(  # noqa: C901
         self,
         is_package_call: bool = False,
         package_owner: str = None,
         package_id: str = None,
         package_instance_id: str = None,
         as_background_task: bool = False,
-        wait_on_tasks: List[Task] = None,
+        wait_on_tasks: List[Union[str, Task]] = None,
     ):
         headers = {"Authorization": f"Bearer {self.config.api_key}"}
 
@@ -256,7 +256,18 @@ class Client(CamelModel, ABC):
             # execution. Additionally, the task will be scheduled to first wait on the other tasks
             # provided in the list of IDs.
             as_background_task = True
-            headers["X-Task-Dependency"] = ",".join([task.task_id for task in wait_on_tasks])
+            task_ids = []
+            for task_or_id in wait_on_tasks:
+                if isinstance(task_or_id, str):
+                    task_ids.append(task_or_id)
+                elif isinstance(task_or_id, Task):
+                    task_ids.append(task_or_id.task_id)
+                else:
+                    raise SteamshipError(
+                        message=f"`wait_on_tasks` should only contain Task of str objects. Got a {type(task_or_id)}."
+                    )
+
+            headers["X-Task-Dependency"] = ",".join(task_ids)
 
         if as_background_task:
             # Will result in the engine persisting the inbound HTTP request as a Task for deferred
@@ -368,7 +379,7 @@ class Client(CamelModel, ABC):
         package_id: str = None,
         package_instance_id: str = None,
         as_background_task: bool = False,
-        wait_on_tasks: List[Task] = None,
+        wait_on_tasks: List[Union[str, Task]] = None,
     ) -> Union[
         Any, Task
     ]:  # TODO (enias): I would like to list all possible return types using interfaces instead of Any
@@ -499,7 +510,7 @@ class Client(CamelModel, ABC):
         package_id: str = None,
         package_instance_id: str = None,
         as_background_task: bool = False,
-        wait_on_tasks: List[Task] = None,
+        wait_on_tasks: List[Union[str, Task]] = None,
     ) -> Union[
         Any, Task
     ]:  # TODO (enias): I would like to list all possible return types using interfaces instead of Any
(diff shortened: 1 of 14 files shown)


dkolas (dismissed):
This came out really nice and clean, good work!

A couple questions about usage / naming below.

src/steamship/base/client.py
```diff
@@ -227,14 +227,14 @@ def _url(
...
-    def _headers(
+    def _headers(  # noqa: C901
         self,
         is_package_call: bool = False,
         package_owner: str = None,
         package_id: str = None,
         package_instance_id: str = None,
         as_background_task: bool = False,
-        wait_on_tasks: List[Task] = None,
+        wait_on_tasks: List[Union[str, Task]] = None,
```
EniasCailliau, line 237: I think it would be better if we only supported List[Task]. Don't know yet why we need to support task_id. Will come back to this.

src/steamship/base/client.py
```diff
@@ -254,9 +254,20 @@ def _headers(
...
-            headers["X-Task-Dependency"] = ",".join([task.task_id for task in wait_on_tasks])
+            task_ids = []
+            for task_or_id in wait_on_tasks:
+                if isinstance(task_or_id, str):
+                    task_ids.append(task_or_id)
+                elif isinstance(task_or_id, Task):
+                    task_ids.append(task_or_id.task_id)
+                else:
+                    raise SteamshipError(
+                        message=f"`wait_on_tasks` should only contain Task of str objects. Got a {type(task_or_id)}."
```
EniasCailliau, line 267:
```suggestion
                        message=f"`wait_on_tasks` should only contain Task or str objects. Got a {type(task_or_id)}."
```

src/steamship/invocable/invocable.py
```diff
@@ -180,6 +189,55 @@ def config_cls(self):
...
+    def invoke_later(
+        self,
+        method: str,
+        verb: Verb = Verb.POST,
+        wait_on_tasks: List[Union[Task, str]] = None,
+        arguments: Dict[str, Any] = None,
+    ) -> Task[Any]:
+        """Schedule a method for future invocation.
+
+        :param method: The method to invoke, as registered with Steamship in the @get or @post decorator.
```
EniasCailliau, line 201: I guess your Pycharm is not set to use numpy docstring's.

tests/steamship_tests/app/integration/test_package_instance.py
```diff
@@ -157,6 +157,22 @@ def replace_string(s: str):
             steamship_dir_fixed = json.loads(replace_string(json.dumps(steamship_dir)))
             assert steamship_dir_fixed == steamship_dir_golden
 
+        # Test that we can call a task which SCHEDULES another task
+        future_greet_resp = instance.invoke("future_greet", name="Unicorn")
+        future_task = Task(client=client, **future_greet_resp)
+        future_task.wait()
+        future_task_bytes = base64.b64decode(future_task.output)
```
EniasCailliau, line 164: Is there a way for the task output not to be b64encoded?

tests/assets/packages/demo_package.py
```diff
@@ -61,6 +73,17 @@ def greet1(self, name: str = "Person") -> InvocableResponse[str]:
     def greet2(self, name: str = "Person") -> InvocableResponse[str]:
         return InvocableResponse(string=f"Hello, {name}!")
 
+    @post("future_greet")
+    def future_greet(self, name: str = "Person") -> InvocableResponse[Task]:
+        task_1 = self.invoke_later("greet", arguments={"name": name})
```
EniasCailliau, line 78: Isn't it weird to call invoke_later without wait_on_tasks?

src/steamship/invocable/invocable.py
```diff
@@ -180,6 +189,55 @@ def config_cls(self):
         """
         return Config
 
+    def invoke_later(
```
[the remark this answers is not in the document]
EniasCailliau: +1 

1. Would also put it one layer down on the PackageService
3. I would also make the wait_on_tasks required.

src/steamship/base/client.py
EniasCailliau: I didn't see any spots where we need to work with just task_id. So I would only support List[Task]

EniasCailliau (approved):

src/steamship/base/client.py
eob: Happy to collapse it, though a trivial example of where taskIds get passed around is the return value of some package methods

tests/steamship_tests/app/integration/test_package_instance.py
eob: Yeah this is such a tricky spot -- base64 is super annoying, but if we don't do it, then we either need to: (1) forbid base64 output or (2) base64 sometimes but not always.

All of those options feel not great to me -- should we pursue some sort of "test the base64 decode and do it automatically if it's viable" trick? 

That way the user always gets back a reasonable python type

src/steamship/invocable/invocable.py
eob: It is (just checked). Is that supposed to throw an error on a docstring like this?

PR #240, opened 2022-11-30
Support looking up plugins via language skills

EniasCailliau:
This PR adds support for looking up plugins via language skills. 

# Interface design 

I swapped out the interface in our design doc with one that I think is more self-documenting. Let's use this PR to discuss whether that is the case or not. 

## Old design

In our original design doc, we added a `skill` parameter to `use_plugin` to look up a plugin using one (or more) skill keywords. 

```
def __init__() -> None: 
	self.entity_tagger = client.use_plugin(skill="entities")
```

We also decided to allow users to decide which vendor offers a specific skill as follows: 

```
def __init__() -> None: 
	self.entity_tagger = client.use_plugin(skill="entities-oneai")
```

As I was implementing this extension, I felt like this was a bad idea because: 

1. It introduces an alternative usage pattern for the `use_plugin` function, which needs external documentation (outside of the function docstring) to separate them. Without reading the documentation users can try out the following access patterns:

* `ship.use_plugin(plugin_handle="oneai-tagger", skill="entities")`
*  `ship.use_plugin(skill="entities", config={}") 

Both would have to raise exceptions even though these usage paths semantically make sense. 

2. It uses a string to point to 2 separate concepts: the skill and the vendor. Although combining these 2 selectors in a string makes sense from the perspective of aliasing, it feels impractical for discovery. 

##  New design 

I decided to experiment with introducing a new function for skill-based import of plugins `use_skill`. It enables me to write an interface that doesn't require reading external documentation.

```
def use_skill(skill: Skill, vendor: Optional[Vendor] = None, ...) -> PluginInstance:   
```

This interface instructs the user to: 
1. Use the Skill Enum to discover the skills supported by Steamship 
2. Use the optional Vendor to select a vendor for a given skill

### DIFF
--- a/src/steamship/client/steamship.py
+++ b/src/steamship/client/steamship.py
@@ -7,6 +7,7 @@ from pydantic import BaseModel
 
 from steamship import Configuration, PackageInstance, PluginInstance, SteamshipError, Workspace
 from steamship.base.client import Client
+from steamship.client.skill_to_provider import SKILL_TO_PROVIDER
 from steamship.data.embeddings import EmbedAndSearchRequest, QueryResults
 from steamship.data.plugin.index_plugin_instance import (
     SHIMMED_INDEX_PLUGIN_HANDLES,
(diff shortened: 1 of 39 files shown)


src/steamship/client/steamship.py
```diff
@@ -183,19 +186,53 @@ def use_plugin(
...
+            raise SteamshipError(
+                f"The provider {provider} has no support for the skill {skill}."
+                f"Use one of the providers that support your skill: "
+                f"{','.join(SKILL_TO_PROVIDER[skill])}"
+            )
+
+        plugin_setup = (
+            SKILL_TO_PROVIDER[skill][provider]
+            if provider
+            else list(SKILL_TO_PROVIDER[skill].values())[0]
```
dkolas, line 213: Is the order of .values() guaranteed stable?  Do we want to have an explicit 'default' vendor in the map?

dkolas (dismissed):
Looks great to me! I like the changes you made to the interface.  Just one comment below about default vendors.

src/steamship/client/steamship.py
EniasCailliau: .values() should in theory be stable. The keys are sorted by order of insertion since Python 3.7 ✅

PR #241, opened 2022-11-30
feat: add support for directory listing without creating an instance

douglas-reid:
Required to allow use of `/__dir__`-based healthchecks against "fake" instances that are not fully-configured. This will allow additional confidence that a package version can be used (and marked as default, etc.)

### DIFF
--- a/src/steamship/invocable/lambda_handler.py
+++ b/src/steamship/invocable/lambda_handler.py
@@ -57,7 +57,8 @@ def internal_handler(  # noqa: C901
 
     if request.invocation.invocation_path == "/__dir__":
         # Return the DIR result without (1) Constructing invocable_cls or (2) Parsing its config (in the constructor)
-        return InvocableResponse(json=invocable_cls_func().__steamship_dir__(invocable_cls_func()))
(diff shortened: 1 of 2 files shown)


douglas-reid:
note: Using CI to validate testing as my local test setup is borked at the moment.

douglas-reid:
reviews: validated the new test pass in staging.

src/steamship/invocable/lambda_handler.py
```diff
@@ -55,6 +55,10 @@ def internal_handler(  # noqa: C901
     else:
         error_prefix = "[ERROR - ?VERB ?PATH] "
 
+    if request.invocation.invocation_path == "/__dir__":
+        # Return the DIR result without (1) Constructing invocable_cls or (2) Parsing its config (in the constructor)
+        return InvocableResponse(json=invocable_cls_func().__steamship_dir__(invocable_cls_func()))
```
dkolas, line 60: I think by calling `invocable_cls_func()` twice, you'd be doing the reflection on the class twice.

PR #242, opened 2022-12-12
[SHIP-510] Add plugin / package and pluginVersion/packageVersion handles to instances

dkolas:
Depends on https://github.com/nludb/nludb/pull/386


eob (approved):
LGTM!

PR #243, opened 2022-12-14
[SHIP-511] Add easier version of PluginInstance.train

dkolas:
Add more params to train method, allowing one-line invocation, and avoiding the possibility of passing a different plugin instance.  Leave the old way for back-compat.


eob (approved):
LGTM

PR #246, opened 2022-12-15
Add new tag constants.

eob:


douglas-reid:
Did we want to add stuff like:

```python
def create_speaker(start_idx, end_idx: int, speaker: str) -> Tag.CreateRequest:
    """Create a speaker tag."""
    return Tag.CreateRequest(
        kind=PROVENANCE,
        start_idx=start_idx,
        end_idx=end_idx,
        name=SPEAKER,
        value={NAME: speaker},
    )
```

Or is that coming in a follow-on PR ?

PR #249, opened 2022-12-16
fix: run tests against staging as well as prod for main branch

douglas-reid:
This PR attempts to setup tests against Staging and Production in a way that allows Staging tests to provide ultimate validation while Production provide advisory information. This _should_ allow for evolution of the client away from Production, keeping in sync with Staging, during development.

This also sets up JUnit reporting tasks for tests to all better test failure inspection directly from the GH status page.


dkolas (approved):
MOAR TESTING

PR #251, opened 2022-12-22
[WIP] Proposal: Introduce transcriber interface

EniasCailliau:
I felt the need to make a Transcriber interface to simplify the creation of transcribers and reduce the code duplication across the ones we already have. 

Open for feedback! 🔫


src/steamship/plugin/blockifier/transcriber.py
```diff
@@ -0,0 +1,77 @@
...
+
+TRANSCRIPT_ID = "transcript_id"
+
+
+class Transcriber(Blockifier):
+    @abstractmethod
+    def start_transcription(
+        self, audio_file: PluginRequest[RawDataPluginInput], mime_type: MimeTypes
+    ) -> str:
+        """Transcribe an audio file and turn it into a transcription and optional Tags."""
```
dkolas, line 20: Looks from below like the `str` return is the `transcript_id`, may want to make a note of that in the docstring since it kinda implies the method returns a transcription and tags.

dkolas (dismissed):
If I'm understanding the goal of this, it's to make the writing of transcription plugins easier (without changing the external interface) by abstracting the status checks, and where to put the text?  Seems fine to me

src/steamship/plugin/blockifier/transcriber.py
```diff
@@ -0,0 +1,77 @@
+from abc import abstractmethod
+from typing import List, Optional, Set
+
+from steamship import Block, File, SteamshipError, Tag, Task, TaskState
+from steamship.base.mime_types import MimeTypes
+from steamship.invocable import InvocableResponse
+from steamship.invocable.plugin_service import PluginRequest
+from steamship.plugin.blockifier.blockifier import Blockifier
```
douglas-reid, line 8: is there any pythonic way to avoid the stutter of `.blockifier.blockifier` ?   maybe we could rename the parent to something like `.content` or `.importer` or `.converter` or something?

src/steamship/plugin/blockifier/transcriber.py
EniasCailliau: Oops. Too fast! Good catch!

src/steamship/plugin/blockifier/transcriber.py
```diff
@@ -0,0 +1,77 @@
...
+from steamship.plugin.inputs.raw_data_plugin_input import RawDataPluginInput
+from steamship.plugin.outputs.block_and_tag_plugin_output import BlockAndTagPluginOutput
+
+TRANSCRIPT_ID = "transcript_id"
+
+
+class Transcriber(Blockifier):
+    @abstractmethod
+    def start_transcription(
+        self, audio_file: PluginRequest[RawDataPluginInput], mime_type: MimeTypes
```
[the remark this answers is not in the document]
EniasCailliau: Fair point. Just checked, both DeepGram and AssemblyAI consider mime_type optional. Going to try it out.

src/steamship/plugin/blockifier/transcriber.py
EniasCailliau: The stutter only applies to this file though. Since I included both Blockifier and Transcriber in the __init__ you can import Transcriber and Blockifier from the steamship.plugin.blockifier namespace (without needing the additional .blockifier).

EniasCailliau:
> If I'm understanding the goal of this, it's to make the writing of transcription plugins easier (without changing the external interface) by abstracting the status checks, and where to put the text? Seems fine to me

Yes, it's not ideal but I think it's good to try things out and see how they feel :)

PR #252, opened 2022-12-22
Proposal: Introduce kind-specific tags

EniasCailliau:


src/steamship/data/block.py
```diff
@@ -22,12 +22,6 @@ class Block(CamelModel):
     text: str = None
     tags: Optional[List[Tag]] = []
 
-    class CreateRequest(Request):
-        id: str = None
-        file_id: str = None
-        text: str = None
-        tags: Optional[List[Tag.CreateRequest]] = []
-
```
dkolas, line 30: Was the removal of `Block.CreateRequest` and `Tag.CreateRequest` something that was necessary for the kind-specific tag classes, or just felt right along the way?

I don't really have a clear opinion / solution in my head, because I find the need to use the `CreateRequest` s VERY annoying, but at the same time, the `File` object and its `CreateRequest` are somewhat different from each other, and I think there's value in a consistent pattern across `File`, `Block`, and `Tag` (and all model objects, really).

Unless I'm reading the code wrong, it looks like you could only use the new Tag subclasses for creation, right?  Could they just subclass Tag.CreateRequest instead?

src/steamship/data/tags/tag.py
```diff
@@ -122,29 +113,194 @@ def query(
...
-    #     return res
+
+class TimestampTag(Tag):
+    def __init__(
+        self,
+        start_time_s: float,
+        end_time_s: float,
+        start_idx: Optional[int] = None,
+        end_idx: Optional[int] = None,
+        value: Optional[Dict[str, Any]] = None,
```
dkolas, line 124: This is definitely a much more pleasant interface for tag creation!

dkolas (commented):
I like the new tag subclasses, but am undecided whether removing the `CreateRequest`s seems like a good idea at the moment.  See below

src/steamship/data/tags/tag.py
eob: Agree -- this is really nice.

src/steamship/data/tags/tag.py
```diff
@@ -122,29 +113,194 @@ def query(
...
+            value={
+                **value,
+                TagValueKey.START_TIME_S: start_time_s,
+                TagValueKey.END_TIME_S: end_time_s,
+            },
+        )
+
+
+class TokenizationTag(Tag):
+    class Type(str, Enum):
```
eob, line 140: I like this constant refactoring into the subclass that you've done!

src/steamship/data/tags/tag_constants.py
```diff
@@ -161,92 +165,25 @@ class TokenTag(str, Enum):
...
     # Catch-all for confidence, score, ranking
     SCORE = "score"
 
-    # A list of numbers. E.g. for an embedding
-    VECTOR_VALUE = "vector-value"
-    NUMBER_VALUE = "number-value"
-    BOOL_VALUE = "bool-value"
-    STRING_VALUE = "string-value"
+    # A list of numbers
+    VALUE = "value"
```
eob, line 175: @dkolas what do you think about the `X-value` versus `value` bit here?

I'm relatively agnostic, but I recall you having specific thoughts on this.

src/steamship/data/block.py
eob: It'd be easier if the two things were split into different PRs, but I agree it'd be lovely to get rid of the `.CreateRequest` objects and don't really mind rolling it up into this one.

Is there a way to get rid of ALL of the `CreateRequest` request objects in this PR so that it's a complete switch over?

eob (changes requested):

src/steamship/data/block.py
EniasCailliau: Sure! Was just testing how the removal of CreateRequest would impact the codebase :) Going to revert these changes and keep that discussion open in another PR

src/steamship/data/tags/tag.py
```diff
@@ -122,29 +125,194 @@ def query(
...
+        start_idx: Optional[int] = None,
+        end_idx: Optional[int] = None,
+        value: Optional[Dict[str, Any]] = None,
+    ):
+        value = value or {}
+        super().__init__(
+            kind=TagKind.SUMMARY,
+            start_idx=start_idx,
+            end_idx=end_idx,
+            value={**value, TagValueKey.VALUE: summary},
```
douglas-reid, line 183: just curious: why not TagValueKey.SUMMARY ?

PR #253, opened 2022-12-22
Remove static use and use_plugin on the Steamship class and add recreate parameter

EniasCailliau:
This PR: 

- Removes the static `use` and `use_plugin` methods on the Steamship client. 
- Add a `recreate` parameter to force the recreation of a package or plugin. Useful for debugging and testing when you always want the latest version of a plugin.

### DIFF
--- a/src/steamship/client/steamship.py
+++ b/src/steamship/client/steamship.py
@@ -155,6 +155,7 @@ class Steamship(Client):
         config: Optional[Dict[str, Any]] = None,
         version: Optional[str] = None,
         fetch_if_exists: Optional[bool] = True,
+        recreate: Optional[bool] = False,
     ) -> PluginInstance:
         """Creates/loads an instance of plugin `plugin_handle`.
 
@@ -190,6 +191,12 @@ class Steamship(Client):
                 fetch_if_exists=fetch_if_exists,
             )
             return instance
+
+        if recreate:
+            try:
+                PluginInstance.get(client=self, handle=instance_handle).delete()
+            except SteamshipError:
+                pass
 
         instance = PluginInstance.create(
             self,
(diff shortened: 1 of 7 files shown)


src/steamship/client/steamship.py
```diff
@@ -77,65 +73,46 @@ def embed_and_search(
...
-    def _instance_use(
-        self,
-        package_handle: str,
-        instance_handle: Optional[str] = None,
-        config: Optional[Dict[str, Any]] = None,
-        version: Optional[str] = None,
-        fetch_if_exists: bool = True,
-    ) -> PackageInstance:
-        """Creates/loads an instance of package `package_handle`.
+        if recreate:
```
dkolas, line 109: That turned out not bad :)

src/steamship/client/steamship.py
```diff
@@ -77,65 +73,46 @@ def embed_and_search(
...
-        workspace_handle: Optional[str] = None,
-        **kwargs,
+        recreate: Optional[bool] = False,
     ) -> PackageInstance:
         """Creates/loads an instance of package `package_handle`.
 
-        The instance is named `instance_handle` and located in the Workspace named `instance_handle`. If no
-        `instance_handle` is provided, the default is `package_handle`.
+        The instance is named `instance_handle` and located in the workspace this client is anchored to.
+        If no `instance_handle` is provided, the default is `package_handle`.
```
dkolas, line 88: I think we should figure out how we want to generally steer people into a new workspace per package before we remove the static methods.

dkolas (commented):
The recreate bit is good, but after our conversation yesterday I don't think we should rip out the static `use` and `use_plugin` methods until we have a clear path for nudging users towards putting new packages in new workspaces.

eob (changes requested):
I definitely think we need another round of UX iteration on the use & use_plugin method calls, but these both seem dangerous to me:

- recreate -> hides deletion of data
- removing the static variants -> makes it extremely easy for two packages to accidentally intermingle data.

With a ton of appreciation for the intent, might be best for us to take this out of the PR list and do some design sketching in notion first since there are so many edge cases to consider first.

PR #254, opened 2022-12-22
Fix empty api-key test

EniasCailliau:


douglas-reid (approved):

douglas-reid:
This seems to fix `Staging`. Let's merge and hope that Production soon catches up.

PR #257, opened 2023-01-04
Add force_create parameter to use and use_plugin to facilitate reliable testing

EniasCailliau:
I know our last PR flagged a `recreate` parameter as dangerous. I'd like to use this PR to revisit the discussion. 

I think introducing a force_create parameter that defaults to False doesn't do a lot of harm but makes integration testing more reliable for plugins and packages. 

I personally think the alternative where the user has to remove the package themselves (e.g. PackageInstance.get().delete()) is too low-level as every integration test will likely want to do this.


eob:
Not opposed! Just want to understand more & see if this is the way it should go in.

Is the idea that this is a sneak route around immutable config? I think I'm OK with that. 

I think `recreate_if_exist` is probably a better name since:

1) It's parallel to `fetch_if_exist`, and
2) It feels more to me like you're going to re-do something than `force` 

If the use case really is just for testing then maybe we put it in a test utility? LMK!

dkolas:
Can we maybe close this PR without merge now that we have the temporary workspace context manager?  IMO that's the safer way to handle testing use cases rather than introducing this.

PR #258, opened 2023-01-04
Remove CreateRequest from public interface

EniasCailliau:
This PR removes the `CreateRequest` interface used for the creation of Files, Blocks, and Tags. 

The removal aims to simplify the public interface of Steamship used to develop and use Invocables.  

The interface of Blocks and Tags were both similar to their CreateRequest counterparts, making the removal easy. 

That wasn't the case for File and File.CreateRequest. I propose a solution where the Steamship client sends custom request payloads to the `file/create` endpoint. This seems an acceptable solution as only the Steamship client creates files using custom payloads, while Invocables usually create "simple" fails containing blocks and tags.


dkolas (approved):
I like how this shakes out more than I expected to.  It still feels like there's a little possible confusion around whether or not a tag or block has been saved, but maybe that's worthwhile to eliminate the need for a child class.  

Maybe at some point we should add a property to CamelModel to say whether or not the object has been saved (and thus has an id)?

src/steamship/data/file.py
```diff
@@ -23,15 +22,9 @@
 
 
 class FileUploadType(str, Enum):
-    FILE = "file"  # The CreateRequest contains a file upload that should be used
-    VALUE = "value"  # The Create Request contains a `text` field that should be used
```
eob, line 27: Flagging to make sure removing VALUE was intentional. I think it was and that that's correct, but just being safe :)

src/steamship/data/tags/tag.py
```diff
@@ -303,23 +291,6 @@ def __init__(
         )
 
 
-class EmbeddingTag(Tag):
```
eob, line 306: Curious why you're deleting this one -- it's super useful. I'll roll forward the various embedder plugins to use it instead of manually creating embedding tags.

eob (changes requested):
It's not currently passing tests, and I think we should preserve the EmbeddingTag (unless I missed something and was wrong in my comment), but other than that this looks great.

I agree w/ Dave a good pin for later is a dirty bit re: whether something has been persisted, but I don't think that needs to be in here.

This is an awesome simplification!

PR #261, opened 2023-01-10
Changes to expose more type info to __dir__

dkolas:


eob (dismissed):
Looks fantastic -- I can start building in support on the front-end!

PR #263, opened 2023-01-19
Playing around with simplified prompt interface

eob:
## Description

This PR introduces our second PluginInstance subclass -- `PromptGenerationPluginInstance`.

It changes no semantics of how plugins, tagging, or existing LLM plugins work; it only acts as a convenience wrapper around PluginInstances that happen to be associated with an LLM tagger.

* It adds a `generate(prompt, prompt_variables)` method which synchronously returns a string.
* It adds an output sanitization method.

## Example Usage

Using this new subclass, one can invoke an LLM with two lines of code and two lines of data:

```
PROMPT = "Greet {name} as if he were a {relation}."
VARIABLES = { "name": "Ted", "relation": "friend" }

llm = Steamship.use('prompt-generation-default', config={ "temperature": 0.9 })
greeting = llm.generate(PROMPT, VARIABLES)
```

## Implementation

* The implementation of `generate` is copy-pasted directly from our current prompt-api template. The intention is to simplify our code with as small as possible a change in present-day mechanics.
* The knowledge of when to use this subclass is hard-coded into our client, as we do with embedding indices, which seems OK since at most there are only 4-5 LLM providers that matter at present.

### DIFF
--- a/src/steamship/client/steamship.py
+++ b/src/steamship/client/steamship.py
@@ -15,12 +15,20 @@ from steamship.data.plugin.index_plugin_instance import (
     SHIMMED_INDEX_PLUGIN_HANDLES,
     EmbeddingIndexPluginInstance,
 )
+from steamship.data.plugin.prompt_generation_plugin_instance import PromptGenerationPluginInstance
 
 _logger = logging.getLogger(__name__)
 
 
 class Steamship(Client):
     """Steamship Python Client."""
+
+    # Some plugins get special subclasses to support experimentation with helper methods and/or
+    # complex plugins (like embedding indices) that rely on the coordination of several moving parts.
+    _PLUGIN_SHIMS = {
+        "prompt-generation-default": PromptGenerationPluginInstance,
+        "prompt-generation-trainable-default": PromptGenerationPluginInstance,
+    }
 
     def __init__(
         self,
@@ -147,6 +155,10 @@ class Steamship(Client):
         return instance
 
     @staticmethod
+    def register_plugin_shim(plugin_handle: str, plugin_class: PluginInstance):
+        Steamship._PLUGIN_SHIMS[plugin_handle] = plugin_class
+
+    @staticmethod
     def use_plugin(
         plugin_handle: str,
         instance_handle: Optional[str] = None,
@@ -239,7 +251,7 @@ class Steamship(Client):
         # This is a shim to make Embedding Indices feel as if they are Plugins.
         # If it works well, we'll turn them into actual plugins on the engine-side.
         if plugin_handle in SHIMMED_INDEX_PLUGIN_HANDLES:
-            instance = EmbeddingIndexPluginInstance.create(
+            return EmbeddingIndexPluginInstance.create(
                 self,
                 plugin_handle=plugin_handle,
                 plugin_version_handle=version,
@@ -247,7 +259,15 @@ class Steamship(Client):
                 config=config,
                 fetch_if_exists=fetch_if_exists,
             )
-            return instance
+        elif plugin_handle in Steamship._PLUGIN_SHIMS:
+            return Steamship._PLUGIN_SHIMS[plugin_handle].create(
+                self,
+                plugin_handle=plugin_handle,
+                plugin_version_handle=version,
+                handle=instance_handle,
+                config=config,
+                fetch_if_exists=fetch_if_exists,
+            )
 
         instance = PluginInstance.create(
             self,
--- a/src/steamship/client/steamship.py
+++ b/src/steamship/client/steamship.py
@@ -28,6 +28,8 @@ class Steamship(Client):
     _PLUGIN_INSTANCE_SUBCLASS_OVERRIDES = {
         "prompt-generation-default": PromptGenerationPluginInstance,
         "prompt-generation-trainable-default": PromptGenerationPluginInstance,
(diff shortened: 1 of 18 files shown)


src/steamship/client/steamship.py
```diff
@@ -144,7 +153,10 @@ def _instance_use(
             fetch_if_exists=fetch_if_exists,
         )
 
-        return instance
+    @staticmethod
+    def register_plugin_instance_subclass(plugin_handle: str, plugin_class: PluginInstance):
```
EniasCailliau, line 157: If we're only planning to use this method during testing I would directly access the static variable in our test scripts instead of using a static function. 

In the tests this would look like: 

```
Steamship._PLUGIN_INSTANCE_SUBCLASS_OVERRIDES[plugin_handle] = plugin_class
```

src/steamship/data/plugin/prompt_generation_plugin_instance.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+class PromptGenerationPluginInstance(PluginInstance):
+    """An instance of a configured prompt completion service such as GPT-3.
+
+    The `generate` method synchronously invokes the prompt against a set of variables that parameterize it.
+    The return value is a single string.
+
+    Example Usage:
+       llm = Steamship.use('prompt-generation-default', config={ "temperature": 0.9 })
```
EniasCailliau, line 16: Does it make sense to use `gpt3` here instead of `prompt-generation-default`?

EniasCailliau (approved):
Looks good!

src/steamship/client/steamship.py
dkolas: +1 to this- I think more thought would be necessary to figure out an extension pattern for user-submitted plugins.

src/steamship/data/plugin/prompt_generation_plugin_instance.py
```diff
@@ -0,0 +1,105 @@
...
+        #
+        # To learn about the Steamship data model, please consult our the docs:
+        # https://docs.steamship.com/workspaces/data_model/index.html
+        file = File.create(self.client, blocks=[Block.CreateRequest(text=prompt_text)])
+
+        # This requests generation from the parameterized prompt. Tagging with our prompt generator
+        # plugin will result in a new tag that contains the generated output.
+        # We `wait()` because generation of text is done asynchronously and may take a few moments
+        # (somewhat depending on the complexity of your prompt).
+        tag_task = file.tag(plugin_instance=self.handle)
```
dkolas, line 44: If we're abstracting away seeing the file created, would it make more sense for this to use the inline tagging method?

tag_task = `self.tag(doc=prompt_text)`

src/steamship/data/plugin/prompt_generation_plugin_instance.py
```diff
@@ -0,0 +1,105 @@
...
+        # Here, we iterate through the content blocks associated with a file
+        # as well as any tags on that content to find the generated text.
+        #
+        # The Steamship data model provides flexible content organization,
+        # storage, and lookup. Read more about the data model via:
+        # https://docs.steamship.com/workspaces/data_model/index.html
+        for text_block in tag_task.output.file.blocks:
+            for block_tag in text_block.tags:
+                if block_tag.kind == TagKind.GENERATION:
+                    return self._clean_output(block_tag.value[TagValueKey.STRING_VALUE])
```
dkolas, line 58: If this isn't going to be in user code, I think there should be a flag on generate (default true) on whether to do this.  It's not clear to me it's the right behavior for all applications.

tests/steamship_tests/plugin/integration/test_e2e_prompt_plugin.py
```diff
@@ -0,0 +1,38 @@
...
+
+        # Now we just invoke use_plugin.
+        # Note: this just applies Steamship.use_plugin with the same testing environment as the unit tests.
+        with steamship_use_plugin(plugin.handle, delete_workspace=True) as llm:
+            # And we can run a prompt. Covering some bases to encode edge cases and their expected behavior.
+
+            # Prompt vars: 1, Provided vars: 1
+            assert llm.generate("Hello, {name}!", {"name": "world"}) == "Hello, world!"
+            # Prompt vars: 0, Provided vars: 1
+            assert llm.generate("Hello, World!", {"name": "world"}) == "Hello, World!"
```
dkolas, line 33: Can this throw an error?

dkolas (approved):
Looks good - a few comments / suggestions

src/steamship/client/steamship.py
eob: Will do!

src/steamship/data/plugin/prompt_generation_plugin_instance.py
eob: Done!

src/steamship/data/plugin/prompt_generation_plugin_instance.py
eob: Ooh. Good point.

tests/steamship_tests/plugin/integration/test_e2e_prompt_plugin.py
eob: I think so -- if the tag shape is unexpected. Should we softly handle it w/ an empty string result?

PR #264, opened 2023-01-24
`with Steamship.temporary_workspace` helper

eob:


src/steamship/client/steamship.py
```diff
@@ -88,6 +90,32 @@ def embed_and_search(
...
+        client.switch_workspace(temporary_handle)
+
+        # Safety check that we are now working form the new workspace.
+        if client.config.workspace_handle != temporary_handle:
+            raise SteamshipError(
+                message=f"Attempted to switch totemporary workspace {temporary_handle} but the client claimed to be working from {client.config.workspace_handle}"
+            )
+
+        yield client
+
```
dkolas, line 109: Would it be better to store a different handle to the client, so that if the user switched workspaces on it, it could still be deleted?

dkolas (dismissed):
Looks good - one possible improvement

PR #267, opened 2023-01-25
Ask for tags, default prompt api

dkolas:


douglas-reid (approved):
Looks good. I only wish there was some elegant way to check to see if `Prompt API` _should_ be tag_string.

PR #270, opened 2023-01-26
Cleaner error message if no auth from deploy

dkolas:


src/steamship/cli/login.py
```diff
@@ -32,7 +32,7 @@ def login(api_base: str, web_base: str) -> str:  # noqa: C901
 
     if not opened_browser:
         raise SteamshipError(
-            "Could not launch browser for login flow. If you are in Replit or another headless environment, please set your Steamship API key as an environment variable or in steamship.json.  See more here: https://docs.steamship.com/configuration/authentication.html"
+            "Could not launch browser for login flow. If you are in Replit, you must set the STEAMSHIP_API_KEY as a Secret and close and reopen this shell. If you are in another headless environment, please set your Steamship API key as an environment variable or in steamship.json. See more here: https://docs.steamship.com/configuration/authentication.html"
```
eob, line 35: Small bug: `~/.steamship.json`

What if we added some vertical spacing and used colorized output followed by `exit()` rather than throwing an Error? I know from the developer perspective this is an error, so throwing exception fits, but from the user's perspective this is a user interface -- and maybe their first interaction with it.

```
Could not launch browser to log in to Steamship.

If you are in Replit:

1) Get an API key at https://steamship.com/account/api
2) Set the STEAMSHIP_API_KEY Replit Secret
3) Close and reopen this shell so that secrets refresh

If you are in a different headless environment, visit https://docs.steamship.com/configuration/authentication.html
```

eob: I don't think the "other headless environment" needs full details since afaik we don't know of such a situation existing yet

dkolas: @eob This method also happens outside the context of the CLI, so we can't put an exit() here.  The PR catches the error in the CLI context and just outputs the message, so I think the result is the same from the user perspective.

Happy to update the message as you suggest.

eob (approved):
🚢🚢🚢 LGTM!

PR #273, opened 2023-01-30
Shippy plugin support

dkolas:
This PR adds support for deploying plugins from the Python-based CLI.

Along the way:
- Required adding update support for Plugin object
- Add user_id to plugin object (depends on an engine change deployed to staging but not yet to prod)
- Had to move ConfigParameters to avoid the dreaded circular dependency


eob (approved):
LGTM!

EniasCailliau (approved):
LGTM!

PR #276, opened 2023-02-02
add tag consts for functionality used in various extension libraries

douglas-reid:
Adds util method and a few consts related to tags that are used in a few related extension libraries (and plugins).


src/steamship/data/tag_utils.py
```diff
@@ -0,0 +1,12 @@
+from typing import Any, Optional
+
+from steamship import File
+
+
+def first_value(file: File, tag_kind: str, value_key: str) -> Optional[Any]:
```
dkolas, line 6: If adding this, might be nice to have an optional tag_name as well

dkolas (commented):
This feels like a really good flavor!  With this kind of stuff, you could wrap a lang chain in a steamship package super quick and then deploy it statefully.

src/steamship/data/tag_utils.py
douglas-reid: done. wdyt?

douglas-reid:
@dkolas @eob @EniasCailliau PTAL. I think this is "ready".

Open question (for which I would like feedback): Do we like the "flexible" (byop-style) wrappers, or do you think we should be more prescriptive?

Put another way: instead of `SteamshipLLM` which allows swapping in your own plugin (maybe with your own API Key, etc.), should we just start with `SteamshipGPT3` which does not allow different plugin use (for now). Similarly, should we have `SteamshipGoogleSearch` instead of the generic (and pluggable) `SteamshipSearch` for now?

requirements.txt
```diff
@@ -5,4 +5,6 @@ inflection==0.5.1
 fluent-logger==0.10.0
 toml==0.10.2
 click==8.1.3
-semver==2.13.0
\ No newline at end of file
+semver==2.13.0
+langchain~=0.0.76
+transformers~=4.26.0
```
eob, line 10: Adding `langchain` and `transformers` to our requirements.txt makes me slightly nervous -- we're so dependency lean right now, but `langchain` alone is going to pull in:

* tensorflow
* pinecone
* wolfram
* numpy
* elasticsearch
* redis
* etc etc (full list copied below)

Which our plugin system works really hard to make unnecessary as things that get pulled in.

With apologies that this feels like a "thrash-inducing" PR question:

- Are there huge labor / UX downsides to considering the route where there's a separate PyPi module `steamship_langchain` that makes importing all these dependencies optional?


```
python = ">=3.8.1,<4.0"
pydantic = "^1"
SQLAlchemy = "^1"
requests = "^2"
PyYAML = "^6"
numpy = "^1"
faiss-cpu = {version = "^1", optional = true}
wikipedia = {version = "^1", optional = true}
elasticsearch = {version = "^8", optional = true}
redis = {version = "^4", optional = true}
manifest-ml = {version = "^0.0.1", optional = true}
spacy = {version = "^3", optional = true}
nltk = {version = "^3", optional = true}
transformers = {version = "^4", optional = true}
beautifulsoup4 = {version = "^4", optional = true}
torch = {version = "^1", optional = true}
jinja2 = {version = "^3", optional = true}
tiktoken = {version = "^0", optional = true, python="^3.9"}
pinecone-client = {version = "^2", optional = true}
weaviate-client = {version = "^3", optional = true}
google-api-python-client = {version = "2.70.0", optional = true}
wolframalpha = {version = "5.0.0", optional = true}
qdrant-client = {version = "^0.11.7", optional = true}
dataclasses-json = "^0.5.7"
tensorflow-text = {version = "^2.11.0", optional = true, python = "^3.10, <3.12"}
tenacity = "^8.1.0"
```

src/steamship/langchain/cache/cache.py
```diff
@@ -0,0 +1,90 @@
...
+        logging.debug(f"cache lookup: {prompt} in {cache_handle}")
+        try:
+            cache_file = File.get(self.client, handle=cache_handle)
+        except SteamshipError:
+            logging.warning(f"no cache found in workspace with handle: {cache_handle}")
+            return None
+
+        try:
+            # find the Steamship Tag in the cache for the prompt and return the values
+            for file_tag in cache_file.tags:
```
eob, line 39: The KeyValueStore class might actually be really useful here. It's hidden in our utils folder but I find myself using it all the time. It basically acts as an implementation of what you've done here that we can rally around and get test coverage over while we don't have a formal KV-store endpoint.

Here are the unit tests for it which demonstrate how it can be used:

https://github.com/steamship-core/python-client/blob/main/tests/steamship_tests/utils/test_kv_store.py

src/steamship/langchain/llm/llm.py
```diff
@@ -0,0 +1,72 @@
...
+        # not run-time
+        llm_plugin = self.client.use_plugin(
+            plugin_handle=self.plugin_handle,
+            instance_handle=instance_handle,
+            config=llm_config,
+            fetch_if_exists=True,
+        )
+
+        logging.info(f"invoking {self.plugin_handle} with prompt: {prompt}")
+        task = llm_plugin.tag(doc=prompt)
```
eob, line 47: Might be helpful to use the `llm_plugin.generate` call here to synchronously go from prompt to output.

```
plugin = self.client.use_plugin(...)
llm_plugin = cast(PromptGenerationPluginInstance, plugin)
generation = llm_plugin.generate(prompt_text, prompt_variables)
```

src/steamship/langchain/memory/conversation.py
```diff
@@ -0,0 +1,123 @@
...
+def _block_sort_key(block: Block) -> str:
+    """Return a sort key for Blocks based on associated timestamp tags."""
+    return [
+        tag.value.get(TagValueKey.TIMESTAMP_VALUE)
+        for tag in block.tags
+        if tag.kind == TagKind.TIMESTAMP
+    ][0]
+
+
+class SteamshipPersistentConversationMemory(Memory):
```
eob, line 37: This is really cool

eob (changes requested):
This is so, so, very cool.

I think there are a few convenient areas where the existing client has formalisms that you can use to simplify the code here which are worth a quick check into (KeyValueStore & llm.generate) 

And curious on your read about requirements.txt -- it seems like this PR would > 10-20x increase our dependency footprint as a library

requirements.txt
douglas-reid: I was worried about this as well.  I don't think it is a UX issue, really(though curious for other thoughts).  I don't know anything about the complexity of separating PyPi modules, but I'm happy to pursue that.

src/steamship/langchain/llm/llm.py
douglas-reid: Good point. I copy/pasted this from earlier stabs in the dark that predated that. I'll update.

eob:
To this question specifically:

```
Open question (for which I would like feedback): Do we like the "flexible" (byop-style) wrappers, or do you think we should be more prescriptive?

Put another way: instead of SteamshipLLM which allows swapping in your own plugin (maybe with your own API Key, etc.), should we just start with SteamshipGPT3 which does not allow different plugin use (for now). Similarly, should we have SteamshipGoogleSearch instead of the generic (and pluggable) SteamshipSearch for now?
```

This feels like one of those over-engineering responses to respond with, but is it trivial to just do:

```
class SteamshipGPT3(SteamshipLLM):
  self.plugin = 'gpt-3'
```

If it really is just a difference of which plugin you indirect to, there may be power in not removing the option of choice entirely -- shows that we can swap things out easily for you.

douglas-reid:
> To this question specifically:
> 
> ```
> Open question (for which I would like feedback): Do we like the "flexible" (byop-style) wrappers, or do you think we should be more prescriptive?
> 
> Put another way: instead of SteamshipLLM which allows swapping in your own plugin (maybe with your own API Key, etc.), should we just start with SteamshipGPT3 which does not allow different plugin use (for now). Similarly, should we have SteamshipGoogleSearch instead of the generic (and pluggable) SteamshipSearch for now?
> ```
> 
> This feels like one of those over-engineering responses to respond with, but is it trivial to just do:
> 
> ```
> class SteamshipGPT3(SteamshipLLM):
>   self.plugin = 'gpt-3'
> ```
> 
> If it really is just a difference of which plugin you indirect to, there may be power in not removing the option of choice entirely -- shows that we can swap things out easily for you.

It is **just** a difference, but only if your plugin adheres to the _implicit_ contract (must be a tagger + must produce the expected tag set ...). I think it is a quick fix of an iteration where we can make the base class and refactor if the pattern looks sticky. WDYT?

douglas-reid:
@eob removed *all* the langchain stuff to separate into a new repo/module as discussed. PTAL at just the util stuff that remains.

src/steamship/langchain/llm/llm.py
douglas-reid: Looking at this again. This isn't a strictly perfect update as the `prompt` string from langchain is the populated template (vs the template itself). Put another way, we don't have the `prompt_variables` at this point. But maybe that's OK and we just ignore that for now?

PR #278, opened 2023-02-06
[SHIP-544] Add File.append and block index

dkolas:
Prereq for multimedia blocks.

Add file.append_block for nice addition of blocks at the end.  Add Block.index_in_file, for validation outside of the file.blocks order of the index.

Depends on https://github.com/nludb/nludb/pull/432


src/steamship/data/file.py
```diff
@@ -254,6 +254,9 @@ def list(client: Client) -> ListFileResponse:
             expect=ListFileResponse,
         )
 
+    def append_block(self, text: str = None, tags: List[Tag] = None) -> Block:
+        return Block.create(self.client, self.id, text, tags)
```
EniasCailliau, line 258:
```suggestion
        return Block.create(self.client, file_id=self.id, text=text, tags=tags)
```

Would use kwargs here since I didn't understand why we used self.id here.

EniasCailliau (approved):
1 suggestion you can accept without changing code :)

PR #283, opened 2023-02-08
Improve task.wait()

eob:
- Adds increased timeout (3min)
- Adds callback (for status bars; I noticed that in Doug's example code)

### DIFF
--- a/src/steamship/base/tasks.py
+++ b/src/steamship/base/tasks.py
@@ -232,9 +232,17 @@ class Task(GenericCamelModel, Generic[T]):
     ):
         """Polls and blocks until the task has succeeded or failed (or timeout reached).
 
-        - Default wait time is 180s = 3min.
-        - on_each_refresh is an optional callback you can get after each refresh whose callback
-          will be: (refresh #, total_elapsed_time in s, task)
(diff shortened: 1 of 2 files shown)


src/steamship/base/tasks.py
```diff
@@ -224,15 +224,31 @@ def post_update(self, fields: Set[str] = None) -> Task:
...
             TaskState.failed,
         ):
-            self.refresh()
             time.sleep(retry_delay_s)
+            self.refresh()
+            refresh_count += 1
+
+            # Possibly make a callback so the caller knows we've tried again
+            if on_each_refresh:
+                on_each_refresh(refresh_count, time.perf_counter() - t0, self)
```
douglas-reid, line 251: Non-blocking comment: do we care at all about warning people not to make long-running calls in the callback?  what's the Python way?

douglas-reid (approved):

src/steamship/base/tasks.py
eob: :raises fist at python's single threadedness: that's a very good question. 

At the very least I'll update the docs

PR #285, opened 2023-02-09
Update task wait message

dkolas:


src/steamship/base/tasks.py
```diff
@@ -261,7 +261,7 @@ def wait(
         # If the task did not complete within the timeout, throw an error
         if self.state not in (TaskState.succeeded, TaskState.failed):
             raise SteamshipError(
-                message=f"Task {self.task_id} did not complete within requested timeout of {max_timeout_s}s"
+                message=f"Task {self.task_id} did not complete within requested timeout of {max_timeout_s}s. The task is still running on the server."
```
douglas-reid, line 264: @dkolas do we want to suggest something like: "You can retrieve its status via Task.get() or try waiting again with wait()" ?

dkolas: Updated :)

douglas-reid (approved):

PR #288, opened 2023-02-10
Initial version of test-instance command

dkolas:
Per discussion here (thanks @gregwhelan):

https://discordapp.com/channels/927675537390968882/1073669075718389851

This initial version allows creation of a test instance of a package from the command line. Takes config params if your package requires them.

I made it as `test-instance` and not a more generic `create-instance` so that I could make some assumptions about creating / reusing a particular workspace.

Right now I have it outputting a little json dict, so that it's clear that you got an instance of the version you wanted, and you can see the config (if your package has config)
Could add a --url-only option or something to have it only spit out the invocation url if piping through jq feels like too much of a pain for scripting

### DIFF
--- a/src/steamship/cli/cli.py
+++ b/src/steamship/cli/cli.py
@@ -1,12 +1,14 @@
+import json
 import logging
 import sys
 import time
 from os import path
+from typing import List
 
 import click
 
 import steamship
-from steamship import Steamship, SteamshipError
+from steamship import PackageInstance, Steamship, SteamshipError, Workspace
 from steamship.base.configuration import Configuration
 from steamship.cli.deploy import (
     PackageDeployer,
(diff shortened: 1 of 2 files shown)


src/steamship/cli/cli.py
```diff
@@ -116,9 +124,87 @@ def deploy():
...
+@click.command()
+@click.option(
+    "--delete-data",
+    "-d",
+    is_flag=True,
+    default=False,
+    help="Delete all existing data by recreating the workspace.",
+)
+@click.option("--config", "-c", multiple=True)
+def test_instance(delete_data, config: List[str]):  # noqa: C901
```
douglas-reid, line 136: i think one thing that would be good is to somewhere provide the "why" of this command (both here and in docs). mainly i'd like to see something that captures how this differs from `temporary_workspace()` and why that's important.

it touches, iiuc, on the larger (potential) notion of cross-workspace artifact sharing, so capturing the intended use case clearly (near the code) would be good, imho.

dkolas:
Closed in favor of #340

PR #289, opened 2023-02-11
Add configuration  hash to auto-named handles

eob:
Steamship configs are immutable; the following two lines will error because they attempt to load the same package instance (`instanceHandle`) with differing configurations:

```
Steamship.use('packageHandle', 'instanceHandle', config={"name"="Larry"}) # Sometimes works; if not done before with different config
Steamship.use('packageHandle', 'instanceHandle', config={"name"="Curly"}) # Always fails, since we know config differs
```

When the instanceHandle is not provided, a default is provided, equal to the packageHandle/pluginHandle:

```
Steamship.use('packageHandle', config={"name"="Larry"}) # Sometimes works; if not done before with different config
Steamship.use('packageHandle', config={"name"="Curly"}) # Always fails, since we know config differs
```

This is good in that it hard-errors early to enforce the immutable config, but potentially confusing to newcomers who expect more typical construction semantics.

Further, it presents a particular challenge if one **intends to have configuration be dynamic**. An example of this would be an argument to a package method which, itself, parameterizes a plugin which must be used in that method's implementation.

In that scenario, our present handling makes:

* "Anonyous" packages/plugins are impossible (as two different configs would be provided to the default instance handle), and
* "non-anonyous" packages/plugins require the name to encode the config object which parameterizes them

That effectively renders dynamic configuration impossible unless the developer knows to hash the entire config object into the handle, which is very unlikely to be apparent without explanation.

# This PR

This PR solves the problem by doing what the developer would have to do anyway: hashing the config object into the auto-generated handle for them.

When an anonymous package/plugin is used with a non-None config block, the default name is changed from:

```
{packageName}
```

to

```
{packageName}-{md5(config)}
```

This renders the following block of code always runnable:

```
Steamship.use('package', config={"name"="Larry"})  # Always works
Steamship.use('package', config={"name"="Curly"})  # Always works
```

## Not Changed

This PR does not change the semantics of the NAMED instance invocation. Those semantics remain the same: once actively named by the user, the config can not change. 

## Shoutout

Thanks for the catch @ytang07!

### DIFF
--- a/src/steamship/client/steamship.py
+++ b/src/steamship/client/steamship.py
@@ -19,6 +19,7 @@ from steamship.data.plugin.index_plugin_instance import EmbeddingIndexPluginInst
 from steamship.data.plugin.plugin_instance import PluginInstance
 from steamship.data.plugin.prompt_generation_plugin_instance import PromptGenerationPluginInstance
 from steamship.data.workspace import Workspace
+from steamship.utils.metadata import hash_dict
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


EniasCailliau (approved):
This will solve so many headaches! Thanks for adding this :)

douglas-reid (approved):

PR #291, opened 2023-02-14
[wip] feat: add decorator for prompt protection

douglas-reid:
Sketch of a potential approach to provide tooling to users to protect their prompts:
- Fuzz-based matching on inputs (ultimate list can live somewhere separately and be imported)
- Max Words for a string input (another path towards limiting exposure)

The UX needs some work, but I think this could provide value. Current usage:

```python
@post("future_greet")
@validate_input_strings(param_names=["name"], max_words=4)
def future_greet(self, name: str = "Person") -> InvocableResponse[Task]:
    task_1 = self.invoke_later("greet", arguments={"name": name})
    return InvocableResponse(json=task_1)
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,3 +6,7 @@ fluent-logger==0.10.0
 toml==0.10.2
 click==8.1.3
 semver==2.13.0
+thefuzz~=0.19.0
+Sphinx~=4.3.2
+pytest~=7.1.2
+msgpack~=1.0.4
--- a/tests/assets/packages/demo_package.py
+++ b/tests/assets/packages/demo_package.py
@@ -14,6 +14,7 @@ from steamship.invocable import (
     create_handler,
     get,
     post,
+    validate_input_strings,
 )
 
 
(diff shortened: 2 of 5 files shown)


requirements.txt
```diff
@@ -5,4 +5,8 @@ inflection==0.5.1
 fluent-logger==0.10.0
 toml==0.10.2
 click==8.1.3
-semver==2.13.0
\ No newline at end of file
+semver==2.13.0
+thefuzz~=0.19.0
+Sphinx~=4.3.2
+pytest~=7.1.2
```
dkolas, line 11: Is this not in requirements.dev.txt?

tests/assets/packages/demo_package.py
```diff
@@ -70,10 +71,12 @@ def greet1(self, name: str = "Person") -> InvocableResponse[str]:
         return InvocableResponse(string=f"Hello, {name}!")
 
     @post("greet")
+    @validate_input_strings(param_names=["name"])
     def greet2(self, name: str = "Person") -> InvocableResponse[str]:
         return InvocableResponse(string=f"Hello, {name}!")
 
     @post("future_greet")
+    @validate_input_strings(param_names=["name"], max_words=4)
```
dkolas, line 79: I really like this decorator style.  I think there would be huge value in having @check_malicious on prompt params.

dkolas (commented):
I think this is a great model for adding these!!

PR #292, opened 2023-02-16
Explicitly wait on snapshot creation in index_plugin_instance

dkolas:
This isn't really a fix, but possibly a reliability band-aid for vector search


eob (approved):
LGTM!

PR #293, opened 2023-02-17
default to protecting api key in serializations

douglas-reid:
Better safe than sorry. Preventing log leakage with API Keys, etc., seems like the right default stance.

### DIFF
--- a/src/steamship/base/client.py
+++ b/src/steamship/base/client.py
@@ -188,12 +188,28 @@ class Client(CamelModel, ABC):
         # implementations), Pydantic will try to include them by default. So we have to suppress that otherwise
         # downstream serialization into JSON will fail.
         if "exclude" not in kwargs:
-            kwargs["exclude"] = {"use", "use_plugin", "_instance_use", "_instance_use_plugin"}
+            kwargs["exclude"] = {
+                "use": True,
+                "use_plugin": True,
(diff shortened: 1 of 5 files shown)


src/steamship/base/client.py
```diff
@@ -188,12 +188,28 @@ def dict(self, **kwargs) -> dict:
         # implementations), Pydantic will try to include them by default. So we have to suppress that otherwise
         # downstream serialization into JSON will fail.
         if "exclude" not in kwargs:
-            kwargs["exclude"] = {"use", "use_plugin", "_instance_use", "_instance_use_plugin"}
+            kwargs["exclude"] = {
+                "use": True,
+                "use_plugin": True,
+                "_instance_use": True,
+                "_instance_use_plugin": True,
+                "config": {"api_key"},
```
eob, line 196: This seems fine to me -- I can't think of anywhere we depend on this, and arguably nobody should ever want this in the dict serialization

src/steamship/base/configuration.py
```diff
@@ -37,7 +37,7 @@
 
 
 class Configuration(CamelModel):
-    api_key: str
+    api_key: SecretStr
```
eob, line 40: Nice! That's a cool feature

eob (approved):
LGTM

PR #294, opened 2023-02-18
Add optional HTTP timeouts to package invocations

dkolas:
Helpful for applications that might need to bail out quickly if the service doesn't respond (ex: Clockbot) :)


EniasCailliau (approved):
LGTM 🎉

PR #295, opened 2023-02-20
Two Minor test improvements

dkolas:
- It's no longer necessary to wait on deploying things, as the deploy chain includes health check and thus waits engine-side
- Skip the empty API key test when running in an env with a browser (locally) so that it doesn't try to log you in


EniasCailliau (approved):
LGTM ✅

PR #296, opened 2023-02-21
[SHIP-539] Non text blocks

dkolas:
This PR:

- Allows Blocks to be uploaded with non-text content
- Gives Blocks a mime-type for their content
- Updates File.append_block to allow for non-text content

Depends on Engine PR: https://github.com/nludb/nludb/pull/446


src/steamship/data/block.py
```diff
@@ -15,13 +17,26 @@ class BlockQueryRequest(Request):
...
+class BlockUploadType(str, Enum):
+    FILE = "file"  # A file uploaded as bytes or a string
+    BLOCKS = "blocks"  # Blocks are sent to create a file
+    URL = "url"  # content will be fetched from a URL
+    NONE = "none"  # No upload; plain text only
+
+
 class Block(CamelModel):
+    """A Block is a chunk of content within a File. It can be plain text content, image content,
+    video content, etc. If the content is not text, the text value may be the empty string
```
EniasCailliau, line 29: If I read the create() function correctly we also support text=None which is in conflict with having to use an empty string for backwards compatibility.

EniasCailliau (dismissed):
Question: Will this be paired with a new style of tagger? Or will the creator of a tagger now have to decide wether adding a new block vs adding tags is a good decision?

PR #297, opened 2023-02-22
Block large embeddings before insertion (with available override)

dkolas:
This is an additional mitigation towards issues with failed embedding index insertions.  If you attempt to embed a string longer than 5000 characters, it will be rejected within the SDK unless you pass the override flag.

### DIFF
--- a/src/steamship/data/embeddings.py
+++ b/src/steamship/data/embeddings.py
@@ -5,6 +5,7 @@ from typing import Any, Dict, List, Optional, Type, Union
 
 from pydantic import BaseModel, Field
 
+from steamship import SteamshipError
 from steamship.base import Task
 from steamship.base.client import Client
 from steamship.base.model import CamelModel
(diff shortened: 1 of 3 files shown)


src/steamship/data/embeddings.py
```diff
@@ -195,10 +198,33 @@ def insert_file(
             expect=IndexInsertResponse,
         )
 
+    def _check_input(self, request: IndexInsertRequest, allow_long_records: bool):
```
steamship-developers, line 201: do we want to also check MAX_ACCEPTED_TOKENS (or whatever)?  Or is that better homed plugin-side (as the limits may shift per-provider)?

steamship-developers (approved):

src/steamship/data/embeddings.py
dkolas: I think that may be better left plugin-side, as tokenization might be different from plugin to plugin anyway, and is somewhat more expensive than a length check.

PR #303, opened 2023-03-01
fix: ensure expectation is met in lookups/creates

douglas-reid:
Steamship should be a good 🚢 . A good ship should not lie to users. It would be a lie to tell a user that a File was created when the object returned is not a File. Hence, Steamship must enforce expectations.

Without this PR, it is possible, though rare, for the following to happen:

```
f = File.create(client=client, blocks=[Block(text="something filtered by WAF")])
print(f.handle)
```

```
steamship.base.error.SteamshipError: ... | 'str' object has no attribute 'handle'
```

### DIFF
--- a/src/steamship/base/client.py
+++ b/src/steamship/base/client.py
@@ -516,6 +516,15 @@ class Client(CamelModel, ABC):
 
         elif task is not None:
             return task
+        elif data is not None and expect is not None:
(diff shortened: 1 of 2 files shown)


dkolas (approved):
STEAMSHIP IS A GOOD 🚢 
STEAMSHIP IS A GOOD 🚢

PR #306, opened 2023-03-06
Add reset function to index_plugin_instance

EniasCailliau:


eob (changes requested):
Sorry for this nit, but small test request to add to this; I think it'll go a long way toward making sure that the client & engine are 100% synced on semantics of reset:

```
index.insert(tag1, tag2)
r = index.query(tag-1-matching-query)
assert r1 is tag1
index.reset()
index.insert(tag2)
r2 = index.query(tag2-matching-query)
assert r2 == tag2
```

that way we can be sure that it really did reset in the back-end :)

eob:
Also sorry for how long it took me to get to this PR! It's been a crazy two weeks!

PR #307, opened 2023-03-07
feat: add utils for building tag queries

douglas-reid:
This PR is meant to add utilities for quickly creating tag queries for use in Steamship. This is meant to provide an error-free  (and hopefully readable) mechanism for constructing tag queries.

NOTE: I realize it is early days yet for the Tag query system, but I've been writing these more and just couldn't stop myself from writing a set of utility classes.


dkolas (approved):
This looks good to me.  This kind of thing isn't really my cup of tea, but people seem to like it, so :)

If we're going to have it, I'd rather have this version where it serializes to the same string for a wire format, rather than making JSON queries.

douglas-reid:
@EniasCailliau @eob any thoughts on this?

eob:
OK by me to push it in and keep iterating on it! Especially since it's in the `utils` package.

Question:  I don't really understand the distinction between `and_` and `And`, e.g. in:

```
    query = BlockTag().and_(
        TagKind("name").and_(Overlaps(TagKind("foo"))).or_(TagValue("bar", Op.EXISTS))
    )
```

eob:
@dkolas @douglas-reid what do you all think about merging / passing on this PR since it's been hanging out for a while?

douglas-reid:
Let's close it. I wanted it, but I think I'm alone in that regard.

PR #308, opened 2023-03-08
Don't snapshot in index plugin instance; let the engine do it

dkolas:
Now that the engine auto-snapshotting has been validated on staging, this PR should go in to allow the SDK to take advantage


eob (approved):
LGTM!

PR #309, opened 2023-03-09
Remove snapshots and snapshot tests from SDK

dkolas:
This functionality is now an engine implementation detail, and clients should not use it directly.


EniasCailliau (approved):
Yeeey! Removing code 🥇 

Great work on our vector store!

PR #310, opened 2023-03-10
[SHIP-540] python generator plugin interfaces

dkolas:
Initial generator plugin interface, with test generator which reverses text.

Depends on engine changes which are not yet deployed to prod.

Review request to especially make sure the parameters of `GenerateRequest` make sense and are sufficiently commented.


src/steamship/data/operations/generator.py
```diff
@@ -0,0 +1,49 @@
+from __future__ import annotations
+
+from typing import List, Optional
+
+from steamship.base.request import Request
+from steamship.base.response import Response
+from steamship.data.block import Block
+
+
+class GenerateRequest(Request):
```
douglas-reid, line 10: any reason not to include a `kwargs` style extra bit here?  I feel like we need a way to pass runtime args to calls to generate.

douglas-reid: the text example was "stop" keywords

douglas-reid: NOTE TO FUTURE SELF: we will address side-channel info in a separate PR.

src/steamship/data/operations/generator.py
```diff
@@ -0,0 +1,52 @@
...
+    """This class provides the input for a request to a Generator.  There are several ways to specify the input; see below"""
+
+    # Input Specification
+
+    # Must specify plugin instance to use
+    plugin_instance: str = None
```
douglas-reid, line 16: clarity: this is instance handle, correct?

douglas-reid (commented):
Just some questions re: Request surface.

src/steamship/data/operations/generator.py
dkolas: Yeah, I copied the behavior of tag, which is awkward. Can change here if the inconsistency is better than the bad-naming

src/steamship/data/operations/generator.py
```diff
@@ -0,0 +1,52 @@
...
+
+    # May specify blocks by their file_id. If so, may specify start and end index
+    input_file_id: str = None
```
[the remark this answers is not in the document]
dkolas: I think that would be a lovely add-on

src/steamship/data/operations/generator.py
```diff
@@ -0,0 +1,52 @@
...
+    input_file_start_block_index: int = None
+    input_file_end_block_index: Optional[
+        int
+    ] = None  # EXCLUSIVE end index, like most programming languages
+
+    # May specify raw text
+    text: Optional[str] = None
```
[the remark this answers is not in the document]
dkolas: I can add comments to make this more explicit.

src/steamship/data/operations/generator.py
```diff
@@ -0,0 +1,52 @@
...
+
+    # Desired output specification
+
+    # Whether we want the output appended to a file
+    append_output_to_file: bool = False
+
+    # May specify a file to which to append the results.  This may be the same file as
+    # the input or not.  If appendOutputToFile is true but the outputFileId is not set,
+    # create a new file.
+    output_file_id: Optional[str] = None
```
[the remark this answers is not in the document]
dkolas: Yes,  should add-on

PR #316, opened 2023-03-15
Use version while generating an `instance_handle`

EniasCailliau:
Add version to generated instance_handle's so changing the version will result in getting a new instance


eob (approved):
LGTM!

PR #317, opened 2023-03-15
ChatFile interface

dkolas:
See tests for usage example


src/steamship/data/chat.py
```diff
@@ -0,0 +1,176 @@
...
+
+
+CHAT_GENERATOR_INSTANCE_HANDLE = "chat-generator-instance-handle"
+
+
+class Chat:
+    """A Chat is a subclass of a File ideal for ongoing interactions between a user and a virtual assistant."""
+
+    generator_instance_handle: str
+    file: File
```
douglas-reid, line 36: two thoughts here:
1) what if we take some sort of `GeneratorMap` instead so that we could build a subclass of `Chat` that is `MultiModalChat` that could dynamically select differing generators, etc.?  Short of that, I think we should take a list of handles.


GeneratorMap could maybe look something like this rough sketch:
```python
class BaseGenerator(BaseModel):
   handle: str
   mime_types: List[str]
   

class ImageGenerator(BaseGenerator):
   mime_types: ["png", ...]


class GeneratorMap():
  def for_mime_type() -> BaseGenerator:
       pass

  def for_handle() -> BaseGenerator:
      pass
```

2) I think the instinct to wrap `File` is a good one, but what if we wrap it as `ChatHistory`?  And have `Chat` use `ChatHistory` for memory? It would allow differing approaches towards storage in the future (tailored as needed).

PR #319, opened 2023-03-17
dir returns sdk version

dkolas:
This PR allows calls to `__dir__` to include the SDK version being used.  This is a prereq for having engine behavior depend on the SDK version, ex. when we introduce instance initialization.


eob (approved):
LGTM! It looks like some extra code might have come along for the ride.

Also: do we know that the __version__ is being properly set when we do a new release?

dkolas:
> LGTM! It looks like some extra code might have come along for the ride.
> 
> Also: do we know that the **version** is being properly set when we do a new release?

Yes, it's being properly set. You can see it when you run shippy.

PR #323, opened 2023-03-22
Package and Plugin Instance initialization

dkolas:
This PR introduces client-side functionality to automatically call instance initialization methods on Python package and plugin instances.

If a Plugin or Package overrides `instance_init`, this method will be called exactly once on creation of a new Instance.  This method takes no parameters, but has access to both the `config` and `context`, which now includes the Instance's `invocation_url`.

Things to especially note / give feedback on:
- `use` and `use_plugin` now by default wait on initialization.  This implies some extra round-trips the FIRST time they're called on an instance; making the call in a way that returns an existing package/plugin instance will still be one-trip.
- `instance_init` has a default `pass` implementation in `Invocable`.  This means that you merely have to override this one method to be called; the downside of this decision is that an Invocable which does nothing useful in initialization must still be waited upon.

Paired engine PR here: https://github.com/nludb/nludb/pull/495


src/steamship/data/package/package_instance.py
```diff
@@ -109,3 +113,37 @@ def invoke(
...
+
+        Parameters
+        ----------
+        max_timeout_s : int
+            Max timeout in seconds. Default: 180s. After this timeout, an exception will be thrown.
+        retry_delay_s : float
+            Delay between status checks. Default: 1s.
+        """
+        t0 = time.perf_counter()
+        refresh_count = 0
```
EniasCailliau, line 136: Don't think you're doing something with the variable aside from incrementing it. You can probably drop it since you're already using time.perf_counter()

src/steamship/data/plugin/plugin_instance.py
```diff
@@ -182,3 +186,37 @@ def get_training_parameters(
...
+
+        Parameters
+        ----------
+        max_timeout_s : int
+            Max timeout in seconds. Default: 180s. After this timeout, an exception will be thrown.
+        retry_delay_s : float
+            Delay between status checks. Default: 1s.
+        """
+        t0 = time.perf_counter()
+        refresh_count = 0
```
EniasCailliau, line 209: Don't think you're doing something with the variable aside from incrementing it. You can probably drop it since you're already using time.perf_counter()

EniasCailliau (approved):
One small nit. LGTM

src/steamship/data/package/package_instance.py
dkolas: Quite right, thanks!

PR #325, opened 2023-03-28
[SHIP-611] Allow returning usage reports in plugin invocations

dkolas:


eob (approved):
LGTM! Thanks for adding this!

PR #328, opened 2023-04-10
feat: add paging to list operations in client SDK

douglas-reid:
Companion PR to https://github.com/nludb/nludb/pull/483.

This adds support for paging in the python SDK for objects that support a `list` call currently.

### DIFF
--- a/src/steamship/base/request.py
+++ b/src/steamship/base/request.py
@@ -1,3 +1,6 @@
+from enum import Enum
+from typing import Optional
+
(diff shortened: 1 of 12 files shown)


dkolas (dismissed):
This looks good - How hard would it be though to either:

- add some sort of `nextPage()` method to ListResponse, and tuck the relevant info in it, so that the user could just easily loop through the pages?

- or make ListResponse iterable in a way that would yield the individual items and then fetch the next page as needed?

PR #329, opened 2023-04-12
feat: add owner and workspace handles to invocation context

douglas-reid:
Companion PR to: https://github.com/nludb/nludb/pull/535

The idea is to allow packages to dynamically provide manifests that use their domain-specific invocation URL.

Concerns:
- Exposes user workspace handles to owners
- Exposes owner handles to users


dkolas (approved):
Looks good to me

PR #334, opened 2023-04-14
Add `delay_ms` to invocable.invoke_later

eob:


src/steamship/base/client.py
```diff
@@ -268,23 +270,40 @@ def _headers(  # noqa: C901
...
+        if task_delay_ms and task_delay_ms < 0:
+            raise SteamshipError(
+                message=f"Unable to wait a negative duration of time (task_delay_ms={task_delay_ms})"
+            )
+
+        if wait_on_tasks or (task_delay_ms and task_delay_ms > 0):
             # Will result in the engine persisting the inbound HTTP request as a Task for deferred
             # execution. Additionally, the task will be scheduled to first wait on the other tasks
             # provided in the list of IDs. Accepts a list of EITHER Task objects OR task_id strings.
             as_background_task = True
```
EniasCailliau, line 282: Am I correct that this is the only shared piece of code between the branches `wait_on_tasks` and `(task_delay_ms and task_delay_ms > 0)`

If yes I would just do 

`as_background_task = wait_on_tasks or (task_delay_ms and task_delay_ms > 0)`

Will improve readability.

EniasCailliau: Or even: 

```
if wait_on_tasks or (task_delay_ms and task_delay_ms > 0) :
   headers["X-Task-Background"] = "true" 
```

EniasCailliau (approved):
1 nit for readability.

PR #336, opened 2023-04-18
[Experimental] Easily create a file ALWAYS returning a file id

eob:
**First Experimental Code!!!**

NOTE: This PR is actually a subset of something larger I wrote. I'm trying to piecemeal things to be a good citizen :)

One thing I've noticed is that packages spend a lot of time tip-toeing around the fact that there are a few ways to import a file... and some of them return a File with ID, while others return an async task.

This function provides a universal "Easy Create" that:

- Takes a URL and tries to do "the right thing"
- Always returns a File object, even if that means awaiting the completion of an Importer plugin

That means folks can just say:

```python
file = scrape(client, url)
```

and continue working


dkolas (dismissed):
No major objections, but you might expand the comment on the main `scrape` method - instead of saying "no matter what", actually spell out that it may way ton an asynchronous task.

PR #337, opened 2023-04-18
[Experimental] easy.blockify

eob:
Builds upon easy.scrape.

Provides a best-effort blockification of a file based on mime type, always returning a task or throwing if blockification was not possible.


dkolas (dismissed):
Looks good to me - I actually rather prefer this sort of thing in the SDK where it's inspectable rather than hidden in engine code.

PR #338, opened 2023-04-18
[Experimental] move context window counting to chatfile

dkolas:
This brings token context window counting from the package level to the ChatFile wrapper, simplifying package code for things like Telegram Buddy.


eob (dismissed):
LGTM

PR #340, opened 2023-04-20
feat: add 'ship use' command for easy instance creation

douglas-reid:
This adds a basic command to the CLI to allow for rapid instance creation. It must be run from a directory containing a Steamship manifest file.

Configuration may be passed as a JSON string, or as a path to JSON file.

NOTE: At the moment, this must be run after 'ship deploy', as no attempt is made to create a package/plugin before creation of an instance. This could be addressed, if desired.

### DIFF
--- a/src/steamship/cli/create_instance.py
+++ b/src/steamship/cli/create_instance.py
@@ -1,0 +1,119 @@
+import json
+from os import path
+from typing import Any, Dict, Optional
+
+import click
+
+from steamship import Steamship, SteamshipError
+from steamship.cli.ship_spinner import ship_spinner
+from steamship.data.manifest import DeployableType, Manifest
+
+
+@click.command(name="use")
+@click.option(
+    "--workspace",
+    "-w",
+    required=False,
+    type=str,
+    help="Workspace handle. The new instance will be created in this workspace. If not specified, "
+    "the default workspace will be used.",
+)
+@click.option(
+    "--instance",
+    "-i",
+    type=str,
+    required=False,
+    help="Instance handle. Used to name the instance (if it needs to be created). If not specified, "
+    "a generated instance name will be used.",
+)
+@click.option(
+    "--config",
+    "-c",
+    type=str,
+    required=False,
+    help="Instance configuration. May be inline JSON or a path to a file. If not specified, "
+    "an empty configuration dictionary will be passed to the instance.",
+)
+def create_instance(
+    workspace: Optional[str] = None, instance: Optional[str] = None, config: Optional[str] = None
+):
+    """Create an instance of your package/plugin for use.
+
+    Must be run from a directory containing a Steamship manifest."""
+    global client
+    global manifest
+
+    try:
+        client = Steamship(workspace=workspace)
+    except SteamshipError as e:
+        click.secho(e.message, fg="red")
+        click.get_current_context().abort()
+
+    if path.exists("steamship.json"):
+        manifest = Manifest.load_manifest()
+    else:
+        click.secho("No manifest found for instance creation.", fg="red")
+        click.secho("Please try again from a directory with a package or plugin manifest.")
+        click.get_current_context().abort()
+
+    if manifest is None:
+        click.secho("Steamship manifest failed to load.", fg="red")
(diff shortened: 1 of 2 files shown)


src/steamship/cli/create_instance.py
```diff
@@ -0,0 +1,119 @@
...
+    instance_fn = client.use
+    if manifest.type == DeployableType.PLUGIN:
+        instance_fn = client.use_plugin
+
+    try:
+        click.echo("Creating a new instance for usage: ", nl=False)
+        with ship_spinner():
+            instance = instance_fn(
+                manifest.handle,
+                instance_handle=instance,
```
dkolas, line 74: I think this might be an issue, or at least unexpected - 

If you pass the same instance handle twice, or don't pass it twice, this will fail, right?  because the version won't match the second time? 

If I'm right, I would be in favor of allowing the user to easily confirm deletion of the old one, or maybe it's a command line flag, or maybe that's the default behavior.

src/steamship/cli/create_instance.py
```diff
@@ -0,0 +1,119 @@
...
+    workspace: Optional[str] = None, instance: Optional[str] = None, config: Optional[str] = None
+):
+    """Create an instance of your package/plugin for use.
+
+    Must be run from a directory containing a Steamship manifest."""
+    global client
+    global manifest
+
+    try:
+        client = Steamship(workspace=workspace)
```
dkolas, line 47: Instead of defaulting to the default workspace, should we default to one named after the package/plugin?  Multiple packages ending up in the same data space seems like a path to a lot of confusion

dkolas (changes requested):
This is great! I have a couple questions about default behaviors that I think we should decide on before merge.

src/steamship/cli/create_instance.py
douglas-reid: This sounds like a reasonable plan to me (and I think matches the web-app UX?).  Will update.

src/steamship/cli/create_instance.py
douglas-reid: there are any number of cases where the instance collisions could be an issue, i suppose. you could imagine "random" workspaces for each invocation dealing with that. but we could also add a flag. do you have a pref?

src/steamship/cli/create_instance.py
```diff
@@ -0,0 +1,119 @@
...
+                manifest.handle,
+                instance_handle=instance,
+                version=manifest.version,
+                config=invocable_config,
+                wait_for_init=True,
+            )
+            if instance:
+                click.secho(
+                    f"\nSuccess! New instance '{instance.handle}' for {manifest.type} '{manifest.handle}' created "
+                    f"in workspace '{client.get_workspace().handle}'.",
```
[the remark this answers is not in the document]
douglas-reid: Refactored a bit. Hope it retains the spirit of what you liked. Will add details below.

douglas-reid:
@dkolas PTAL. Refactored a bit.

Example uses:

```bash
$ ship use 
Using values:
- Workspace: space-9472efd8e00d29fc28878b40c32d3284
- Package: telegram-bot-with-tools
- Version: 0.2.0-rc.1
- Instance Handle: telegram-bot-with-tools-2f6178179138fb261feefe8e2bc5ec12
- Configuration: {}
Creating new (or fetching existing) instance: 🚢
Failed to create instance: Configuration for this PackageInstance is invalid.
Aborted!
```

```bash
$ ship use -c config.json  
Using values:
- Workspace: space-d489fda7fd78e8fd3f05087d4fec2fb0
- Package: telegram-bot-with-tools
- Version: 0.2.0-rc.1
- Instance Handle: telegram-bot-with-tools-11decc2b314e674c33a4a1b2f5624d66
- Configuration: {"bot_token": "Asd"}
Creating new (or fetching existing) instance: 🚢
Success!
Instance URL: <snip>/space-d489fda7fd78e8fd3f05087d4fec2fb0/telegram-bot-with-tools-11decc2b314e674c33a4a1b2f5624d66/
```

dkolas (approved):

PR #341, opened 2023-04-21
Proposal - Experimental Telegram Helpers

eob:
Proposal:

We're about to have >5 packages which all replicate a variety of telegram hooks, and we're already seeing reports in Discord of our Telegram interactions going awry and SMS-bombing folks.

I think there's a lot of safety / value in pulling today's sketch of the Telegram code into a managed wrapper and then relying on that wrapper.

It will change, but we just manage that by updating it here and then bumping the Steamship version, so I think this is a strict safety-adder. 

Folks generally OK w/ this?

### DIFF
--- a/src/steamship/experimental/transports/telegram.py
+++ b/src/steamship/experimental/transports/telegram.py
@@ -1,8 +1,11 @@
 import logging
+import tempfile
+from typing import Optional
 
 import requests
 
-from steamship import Block, SteamshipError
+from steamship import SteamshipError
+from steamship.experimental.transports.chat import ChatMessage
 from steamship.experimental.transports.transport import Transport
 
 API_BASE = "https://api.telegram.org/bot"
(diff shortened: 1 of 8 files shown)


src/steamship/experimental/transports/telegram.py
```diff
@@ -0,0 +1,111 @@
...
+                    f"Telegram transport unable to send a block of MimeType {block.mime_type}"
+                )
+
+    def _info(self) -> dict:
+        """Fetches info about this bot."""
+        resp = requests.get(f"{self.api_root}/getMe").json()
+        logging.info(f"/info: {resp}")
+        return {"telegram": resp.get("result")}
+
+    def _parse_inbound(self, payload: dict, context: Optional[dict] = None) -> ChatMessage:
```
douglas-reid, line 80: should we have an option for `update_id` and caching the last N? of them so that we avoid duplications?

src/steamship/experimental/transports/telegram.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def _send(self, blocks: [ChatMessage]):
+        """Send a response to the Telegram chat."""
+        for block in blocks:
+            chat_id = block.get_chat_id()
+            if block.is_text() or block.text:
+                params = {"chat_id": int(chat_id), "text": block.text}
+                requests.get(f"{self.api_root}/sendMessage", params=params)
+            elif block.is_image():
+                _bytes = block.raw()
```
douglas-reid, line 58: note: there is a size limit for Telegram uploads. we should probably add handling for any sort of common util for not trying to send anything too large.

src/steamship/experimental/transports/telegram.py
eob: 100%. I think the next round of work on this could start to work with the `ChatFile` class that @dkolas has been working on to make sure we don't duplicate inbound or outbound.

Does the LLM caching work you've been doing related to this? Or is that a separate set of code?

src/steamship/experimental/transports/telegram.py
eob: Ah, good to know. I'll fiddle with it and push up an update

PR #343, opened 2023-04-22
feat: add `ship serve` command for local development

eob:
Adds a `ship serve` command to the CLI.

```
ship serve
```

## To be figured out later

- What the base path should be (e.g. is it /method? Is it /workspace/instance/method?)

## Future PRs

The following code exists the telegram-buddy project and can be ported in as follow-ups:

- NGrok support for packages that need to register public endpoints.
- Overloading of `invoke_later` for a shimmed simulation of most async tasks


src/steamship/cli/local_server/server.py
```diff
@@ -0,0 +1,78 @@
...
+    To shut down, call:
+
+       server.stop()
+
+    """
+
+    invocable: Type[Invocable]
+    port: int
+    server: TCPServer
+    default_api_key: Optional[str] = (None,)
```
dkolas, line 26: Is this tuple syntax necessary for some reason?  Not a big deal, I just want to make sure there's not something here I don't understand

src/steamship/invocable/invocable_localhost.py
dkolas, line 1: Should this empty file be here?

src/steamship/cli/cli.py
```diff
@@ -65,6 +69,78 @@ def ships():
...
+    api_module = get_api_module(path)
+    invocable_class = get_class_from_module(api_module)
+    click.secho(f"Found Invocable: {invocable_class.__name__}")
+
+    server = SteamshipHTTPServer(
+        invocable_class,
+        port=port,
+        invocable_handle=invocable_handle,
+        invocable_version_handle=invocable_version_handle,
+        invocable_instance_handle=invocable_instance_handle,
```
dkolas, line 127: Would it make more sense to default these from the manifest rather than `None`?  I can't think of a reason it would matter to the debugging flow though

src/steamship/cli/local_server/handler.py
```diff
@@ -0,0 +1,311 @@
...
+                exception=ex,
+            )
+
+    try:
+        invocable = invocable_cls_func()(
+            client=client, config=request.invocation.config, context=invocation_context
+        )
+        # TODO: We don't want to run this every time, but for now we are.
+        logging.info("Running __instance_init__")
+        invocable.instance_init()
```
dkolas, line 75: Is this the only difference from the version of this method in `lambda_handler`?  If so, would it not be better to just call the instance init after you've gotten the handler so that you could just use `lambda_handler.py`'s version of this method and not have to copy/paste the rest? The lambda_handler code is a little confusing / ugly already, so avoiding replicating it seems worthwhile if possible.

src/steamship/cli/local_server/handler.py
```diff
@@ -0,0 +1,311 @@
...
+    except Exception as ex:
+        logging.exception(ex)
+        return InvocableResponse.error(
+            code=HTTPStatus.INTERNAL_SERVER_ERROR,
+            prefix=error_prefix,
+            exception=ex,
+        )
+
+
+def handler(bound_internal_handler, event: Dict, _: Dict = None) -> dict:  # noqa: C901
```
dkolas, line 111: Is there a way to refactor this method in `lambda_handler` such that so much of this code doesn't have to be copy/pasted here?  Same logic as above comment.

dkolas (approved):
This is awesome that this works now. Some annoying requests below.

PR #351, opened 2023-05-08
Experimental Telegram Bot Superclass

dkolas:
This PR takes the most recent error-handling changes to TelegramBuddy and extracts out a superclass.

Encapsulates:

- receiving messages
- sending messages, 
- instance init for setting webhook
- getting webhook status, 
- disconnecting webhook
- post route for auto-web-ui

Let me know if you feel like the content is too much / too little.

https://github.com/steamship-packages/telegram-buddy/pull/8

shows what's left in TelegramBuddy


src/steamship/experimental/package_starters/telegram_bot.py
```diff
@@ -0,0 +1,136 @@
...
+        return TelegramBotConfig
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.api_root = f"https://api.telegram.org/bot{self.config.bot_token}"
+
+    def instance_init(self):
+        """This instance init method is called automatically when an instance of this package is created. It registers the URL of the instance as the Telegram webhook for messages."""
+        webhook_url = self.context.invocable_url + "respond"
+        response = requests.get(
```
eob, line 32: It might even be worth it to go one level MORE error-tolerant and wrap this whole thing in a try/catch that just complains but otherwise continues.

That way we can confidently support folks who never enter a telegram key (my thinking: this is probably about to become the *de facto* base class for ALL AGENTS until we have something to replace it with, so planning for the case where someone doesn't want to involve telegram is probably good)

src/steamship/experimental/package_starters/telegram_bot.py
```diff
@@ -0,0 +1,136 @@
...
+            response = self.response_for_exception(e)
+
+        return {
+            "answer": response,
+            "sources": [],
+            "is_plausible": True,
+        }
+
+    @post("respond", public=True)
+    def respond(self, update_id: int, **kwargs) -> InvocableResponse[str]:
```
douglas-reid, line 72: @dkolas wdyt about wrapping the whole thing in a try/except?  and remove the nested handling?

src/steamship/experimental/package_starters/telegram_bot.py
```diff
@@ -0,0 +1,136 @@
...
+        message_id = 0
+
+        try:
+            response = self.respond_to_text(question, chat_session_numeric_id, message_id)
+        except SteamshipError as e:
+            response = self.response_for_exception(e)
+
+        return {
+            "answer": response,
+            "sources": [],
```
eob, line 67: This prevents us from delivering sources with an Ask My XYZ bot, which for that class of Agent is a feature that folks use

One option without going full experimenta.transport style is to have respond_to_text return a tuple with optional sources array as the second item

eob (changes requested):
Looks awesome; a few questions / suggestions to make sure all of our current usage scenarios (no telegram; yes sources) are supported

eob (changes requested):
Looks awesome; a few questions / suggestions to make sure all of our current usage scenarios (no telegram; yes sources) are supported

src/steamship/experimental/package_starters/telegram_bot.py
```diff
@@ -0,0 +1,136 @@
...
+            params={"url": webhook_url, "allowed_updates": ["message"]},
+        )
+        if not response.ok:
+            raise SteamshipError(
+                f"Could not set webhook for bot. Webhook URL was {webhook_url}. Telegram response message: {response.text}"
+            )
+        logging.info(f"Initialized webhook with URL {webhook_url}")
+
+    @abstractmethod
+    def respond_to_text(self, message_text: str, chat_id: int, message_id: int) -> Optional[str]:
```
douglas-reid, line 43: i think this is OK for now, but I'm wondering why not pass through the other args as **kwargs here too (in case there is something like group info or user info needed)?

def respond_to_text(self, message_text: str, chat_id: int, message_id: int, **kwargs) -> Optional[str]:

and

response = self.respond_to_text(message_text, chat_id, message_id, **kwargs)

It allows a backdoor abuse channel for the contract, but provides a decent amount of flexibility.

src/steamship/experimental/package_starters/telegram_bot.py
dkolas: I'm not sure I follow this one. If we're in the context of inheriting from `TelegramBot` (not `Agent` or anything more general) then I would argue that an instance that doesn't provide a valid telegram bot key is most definitely an error.

src/steamship/experimental/package_starters/telegram_bot.py
dkolas: Confused a little about the intent here as well - I thought that the "answer" wrapper for telegram bots was just to make it so that the telegram bot could also be invoked by the web ui?  

If, for now, because this is a superclass and not a transport mix-in, the point of anything that inherits from it is to deploy to Telegram, and Telegram doesn't have any special handling for a source list, then what would be the point of delivering a source array?  It would be better if you were an implementor (again, with a focus on Telegram) to pre-format the sources into the returned text, right?

src/steamship/experimental/package_starters/telegram_bot.py
dkolas: Sounds like a good idea to me

src/steamship/experimental/package_starters/telegram_bot.py
dkolas: Sure!

PR #354, opened 2023-05-10
strawman: sketch of agent bits of SDK

douglas-reid:
**DO NOT SUBMIT**

This PR exists **solely** for trying to facilitate SDK design. It is not meant NOR intended to be a finalized approach that is consumed by python SDK users.

Questions for reviewers:
- is this "style" the one we want to pursue?
- do we need fuller worked examples here?
- is this approximately the right level of opinionated for the SDK?  Should it be heavier?  Lighter?
- do we think something like `AgentContext` is valuable?  What role should it play?
- Is it useful to separate `Action` from `Tool`?  Is there a better concept there (`ToolExecution`? other?)
- Do we believe `Block` + `Tag` is the right way to pass information between `Tool`s and between `Tool` and `Agent` ?

**DO NOT SUBMIT**

### DIFF
--- a/src/steamship/agents/agents.py
+++ b/src/steamship/agents/agents.py
@@ -1,0 +1,187 @@
+import uuid
+from abc import ABC, abstractmethod
+from typing import Any, List, Optional
+
+from pydantic import BaseModel
+
+from steamship import Block, File, Steamship, Tag, Task
+from steamship.invocable import InvocableResponse, PackageService, post
+from steamship.utils.kv_store import KeyValueStore
+
+
+class Action(BaseModel):
+    name: str
+    data: Any
+
+
+class FinishAction(Action):
+    pass
+
+
+class AgentContext(BaseModel, ABC):
+    # todo: define what should be in this context
+    # some set of state-related to agent (maybe request ids, callbacks, etc.)
+    pass
+
+    @abstractmethod
+    def update_blocks(self, blocks: List[Block]):
+        pass
+
+    @abstractmethod
+    def append_log(self, message: str):
+        pass
+
+
+class Tool(BaseModel, ABC):
+    name: str
+    human_description: str
+    ai_description: str
+
+    @abstractmethod
+    def run(self, tool_input: List[Block], context: AgentContext) -> List[Block]:
+        pass
+
+
+class KnockKnockTool(Tool):
+    name = "KnockKnockTool"
+    human_description = "Starts Knock-Knock Jokes"
+    ai_description = ("Used to begin the telling of a joke.",)
+
+    def run(self, tool_input: List[Block], context: AgentContext) -> List[Block]:
+        context.append_log("starting knock-knock joke...")
+        return [Block(text="Knock-Knock")]
+
+
+class WorkspaceTool(Tool, ABC):
+    client: Steamship
+
+
+class TaskScheduler(WorkspaceTool):
+    name = "TaskScheduler"
+    human_description = "Schedule async calls to a package"
+    ai_description = ("Used to schedule reminders for the user at a future point in time.",)
+
+    def __init__(self, client: Steamship, instance_handle: str, method: str):
+        super().__init__(client=client)
+        self._handle = instance_handle
+        self._method = method
+
+    def run(self, tool_input: List[Block], context: AgentContext) -> List[Block]:
+        # parse and call _schedule_task
+        task = self._schedule_task("some-time-in-future", {"arg1": "foo"})
+        return [
+            Block(
+                text="Your task has been scheduled.",
+                tags=[Tag(kind="Task ID", value={"task_id", task.task_id})],
+            )
+        ]
+
+    def _schedule_task(self, time: str, task_kwargs: dict) -> Task:
+        url = self.client._url(operation="package/instance/invoke")
+        payload = {
+            "instanceHandle": self._handle,
+            "payload": {
+                "httpVerb": "POST",
+                "invocationPath": self._method,
+                "arguments": task_kwargs,
+            },
+        }
+        return self._call(url, payload, time)
+
+    def _call(self, url, payload, schedule_time) -> Task:
+        # does the steamship-appropriate calling function.
+        pass
+
+
+class ExampleCacheTool(WorkspaceTool):
+    name = "CacheTool"
+    human_description = "In-Memory Key-Value Store"
+    ai_description = ("Used to store and lookup values by a known key.",)
+    store: KeyValueStore
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.store = KeyValueStore(client=self.client, store_identifier=f"cache-{uuid.uuid4()}")
+
+    def run(self, tool_input: List[Block], context: AgentContext) -> List[Block]:
+        pass
+
+
+class Agent(BaseModel, ABC):
+    tools: Optional[Tool] = []
+
+    @abstractmethod
+    def next_action(self, context: AgentContext) -> Action:
+        pass
+
+    @abstractmethod
+    def execute_action(self, action: Action, context: AgentContext):
+        pass
+
+    @abstractmethod
+    def run(self, agent_input: List[Block], context: AgentContext) -> List[Block]:
+        pass
+
+
+REACT_PROMPT = "<insert prompt that works here>."
+
+
+class ReACTAgent(Agent):
+    client: Steamship
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self._llm = self.client.use_plugin("gpt-4")
+
+    def _extract_action(self, blocks: [Block]) -> Action:
+        # whatever format matches the prompt...
+        pass
+
+    def next_action(self, context: AgentContext) -> Action:
+        # assuming most recent block is stored in context ??
+        action_task = self._llm.generate(...)
+        action_task.wait()
+        return self._extract_action(action_task.output.blocks)
+
+    def execute_action(self, action: Action, context: AgentContext):
+        tool_name = action.name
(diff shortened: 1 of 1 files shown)


src/steamship/agents/agents.py
```diff
@@ -0,0 +1,187 @@
...
+            self.execute_action(action=action, context=context)
+
+
+class ChatHistory(BaseModel):
+    pass
+
+
+class ChatReACTAgent(ReACTAgent):
+    client: Steamship
+    chat_history: ChatHistory
```
douglas-reid, line 165: Is this better represented in `AgentContext`?

src/steamship/agents/agents.py
```diff
@@ -0,0 +1,187 @@
...
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.agent = ReACTAgent(client=self.client)
+
+    @post("run_agent")
+    def run_agent(self, query: str) -> InvocableResponse:
+        query_block = Block.create(
+            client=self.client, text=query, tags=[Tag(kind="user-input", value={...})]
+        )
+        output_blocks = self.agent.run(agent_input=[query_block], context=AgentContext())
```
douglas-reid, line 185: should we focus on convenience methods for input?

src/steamship/agents/agents.py
```diff
@@ -0,0 +1,187 @@
...
+    data: Any
+
+
+class FinishAction(Action):
+    pass
+
+
+class AgentContext(BaseModel, ABC):
+    # todo: define what should be in this context
+    # some set of state-related to agent (maybe request ids, callbacks, etc.)
```
douglas-reid, line 23: should this include a "transaction" log of sorts?

src/steamship/agents/agents.py
```diff
@@ -0,0 +1,187 @@
...
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.store = KeyValueStore(client=self.client, store_identifier=f"cache-{uuid.uuid4()}")
+
+    def run(self, tool_input: List[Block], context: AgentContext) -> List[Block]:
+        pass
+
+
+class Agent(BaseModel, ABC):
```
douglas-reid, line 110: do these feel like the bits we want to expose to implementation?  Others?  Slightly different mix?

src/steamship/agents/agents.py
```diff
@@ -0,0 +1,187 @@
...
+        pass
+
+
+class Tool(BaseModel, ABC):
+    name: str
+    human_description: str
+    ai_description: str
+
+    @abstractmethod
+    def run(self, tool_input: List[Block], context: AgentContext) -> List[Block]:
```
douglas-reid, line 41: i wonder if we want wrapper base methods that do things like report start/stop, etc.

src/steamship/agents/agents.py
```diff
@@ -0,0 +1,187 @@
...
+class ChatReACTAgent(ReACTAgent):
+    client: Steamship
+    chat_history: ChatHistory
+
+    def __init__(self, chat_id: str, **kwargs):
+        super().__init__(**kwargs)
+        self.chat_history = File.create(client=self.client, handle=f"history-{chat_id}")
+
+
+class ExampleService(PackageService):
```
eob, line 172: I think I'm pretty sold on the idea that Agent owned by Package but not Package. And this general flavor of how they interact feels pretty good to me. 

Some high level thinking out loud:

- If agents want to schedule future work, they'll need the RequestContext and some notion of "invoke_later". That has the potential to turn into a weird encapsulation leak... but I not so concerned that it feels like a reason to make Agent = Package. 
- I think some standard notion of both input and output might make a ton of sense at the package-level. I'm not sure if it's `conversational_input` or `agent_input` or what -- but just to homogenize the comms to the point of plug-and-play telegram / debugging / agent wiring / Block marshalling, etc.

.. basically this direction here feels super comfortable to me.

src/steamship/agents/agents.py
eob: This feels like a weak argument for "yes", but I do know that in the Engine, wrapping things into a Context object made it miles easier to then add new things to that context when we realized they were important.

Also I suppose it provides slightly stronger typing than relying on a kwargs list that might change over time. 

The fact that frameworks like Langchain so heavily prefer kwarg-plumbing makes me want to understand the thinking there more, though, too.. I guess it provides for easier ad-hoc overriding?

src/steamship/agents/agents.py
```diff
@@ -0,0 +1,187 @@
...
+from typing import Any, List, Optional
+
+from pydantic import BaseModel
+
+from steamship import Block, File, Steamship, Tag, Task
+from steamship.invocable import InvocableResponse, PackageService, post
+from steamship.utils.kv_store import KeyValueStore
+
+
+class Action(BaseModel):
```
eob, line 12: Is this the "Noun which represents the doing of a thing?"

E.g. --- All agent behavior is composed as [a1, a2... an] in the presence of some AgentContext?

The ReAct reasoner, or python code, exists to enqueue future Actions. And future Actions only take the output of prior Actions as input?

src/steamship/agents/agents.py
```diff
@@ -0,0 +1,187 @@
...
+    name: str
+    data: Any
+
+
+class FinishAction(Action):
+    pass
```
eob, line 18: Is this the Action which represents the intent to stop processing?

src/steamship/agents/agents.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+class AgentContext(BaseModel, ABC):
+    # todo: define what should be in this context
+    # some set of state-related to agent (maybe request ids, callbacks, etc.)
+    pass
+
+    @abstractmethod
+    def update_blocks(self, blocks: List[Block]):
```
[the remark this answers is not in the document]
eob: - I feel like a little magic could go a long way
- I think providing some really strong handling of everything from inner thoughts to eventual output is going to help the monitoring story 100x.

src/steamship/agents/agents.py
eob: Yeah -- and I think callbacks into the agent or python too.

Since a lot of times the user of a tool might not be the developer of the tool (or in the ReACt case.. might not even be the direct invoker!)

So that the wrapping agent or even packageService could know when a tool was about to execute, did execute, etc.

eob (commented):
🚀🚀🚀🚀🚀🚀

eob (commented):
🚀🚀🚀🚀🚀🚀

eob:
Looking at HF Agents/Tools, another very neat design property they seem to have, that I think we could replicate, is the following:

An Agent/Tool in copy-pasted Python works...

But if you were to ship deploy agent.py or tool.py, we could just "do the right thing" even without steamship.json (since most of that is inferred from the python file anyway) and now it's a remoted Agent/Tool.

And if you wanted, you could then just have a RemoteAgent("handle") or RemoteTool("handle") that you used instead.

It probably makes sense for each instance to always pin itself to the latest version unless you specifically mention the version. Which is slightly different than current semantics of it pinning itself to the latest version upon first invocation, and then not auto-rolling-forward unless you update the version & your code. That way we still support versioning, but default to $LATEST at all times.

That also makes us compatible with HF agents -- you could have SteamshipTool("handle") and SteamshipAgent("handle") that were wrapped in a way that lets us help the HF community by providing agents with memory attached

src/steamship/agents/agents.py
dkolas: @eob No, I think Files are the obvious place to be storing this context.

I think this context object is a very useful abstraction.  I wonder if remembering things (memory) is really a part of a context object and not a separate Tool.

src/steamship/agents/agents.py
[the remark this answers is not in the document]
dkolas: Agree that layering some structure (and potentially indices) on top of the blocks would be very useful. This feels like the place to have an opinion about what needs to be recorded during the interactions of an agent.

src/steamship/agents/agents.py
```diff
@@ -0,0 +1,187 @@
...
+        context.append_log("starting knock-knock joke...")
+        return [Block(text="Knock-Knock")]
+
+
+class WorkspaceTool(Tool, ABC):
+    client: Steamship
+
+
+class TaskScheduler(WorkspaceTool):
+    name = "TaskScheduler"
```
dkolas, line 60: Task Scheduler doesn't feel like a Tool to me, it feels like an aspect of when a tool would execute. Would love to find a good way to slide this laterally into a different part.

src/steamship/agents/agents.py
```diff
@@ -0,0 +1,187 @@
...
+        }
+        return self._call(url, payload, time)
+
+    def _call(self, url, payload, schedule_time) -> Task:
+        # does the steamship-appropriate calling function.
+        pass
+
+
+class ExampleCacheTool(WorkspaceTool):
+    name = "CacheTool"
```
dkolas, line 97: Should Caches be part of the Context instead of a separate Tool?

dkolas (commented):
To me the major missing piece here is some notion of AgentExecutor, which is an object that handles the mechanics of how to invoke the next Tool or Action in the process.

This is a bit of a forward ref to @eob 's musings on async, but a LocalAgentExecutor can just syncrhonously wait on async things to finish. A PackageAgentExecutor knows how to sleep, make async calls to the task system, and make package callbacks when those steps finish.

Moreover this creates the natural place for all of the juicy monitoring hooks on when things start / stop, how long each step takes, etc etc.

eob:
@douglas-reid  OK if we close this one to clean up the open PRs view?

PR #356, opened 2023-05-10
Make parse inbound returns optional

dkolas:
If the telegram bot joins a group, or receives an image, it gets a message with no text value.  Rather than this throwing an error in parse_inbound, we return None, so that the bot can decide not to reply without try/except.


eob (approved):
LGTM!

PR #357, opened 2023-05-11
Fix web-bot

EniasCailliau:
This PR makes the response format of the webbot compatible with what is currently supported by the front end.  

image

I also added the field `who` to ChatMessage so we wouldn't need a hacky way to include that field in our response (by default we ignore fields that aren't defined in the spec). For context, here's how the hackathon project added the who field: 

image
 

🚨 This PR temporarily removes support for source attribution. This will be added back in a follow-up PR to decide if and how sources fit into ChatMessage responses.

### DIFF
--- a/src/steamship/experimental/package_starters/web_bot.py
+++ b/src/steamship/experimental/package_starters/web_bot.py
@@ -1,14 +1,23 @@
 from abc import ABC, abstractmethod
-from typing import Any, Dict, List, Optional
+from typing import List, Optional
 
-from steamship import Block
 from steamship.experimental.transports.chat import ChatMessage
 from steamship.experimental.transports.steamship_widget import SteamshipWidgetTransport
 from steamship.invocable import PackageService, post
 
 
-class SteamshipWidgetBot(PackageService, ABC):
+def response_for_exception(e: Optional[Exception], chat_id: Optional[str] = None) -> ChatMessage:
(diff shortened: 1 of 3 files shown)


src/steamship/experimental/transports/chat.py
```diff
@@ -6,6 +6,9 @@
 
 
 class ChatMessage(Block):
+
+    who: Optional[str] = "bot"
```
dkolas, line 10: Was this supposed to be in this pr?

src/steamship/experimental/package_starters/web_bot.py
```diff
@@ -20,28 +29,13 @@ def create_response(self, incoming_message: ChatMessage) -> Optional[List[ChatMe
         raise NotImplementedError
 
     @post("answer", public=True)
-    def answer(self, **payload) -> Dict[str, Any]:
+    def answer(self, **payload) -> List[ChatMessage]:
```
dkolas, line 32: Sorry I botched this. Thanks for catching

dkolas (approved):
Looks good - there may be one stray unrelated line in the PR. Thanks for fixing.

src/steamship/experimental/transports/chat.py
EniasCailliau: It was yes, without it the front-end thinks the message was written by the user.

PR #358, opened 2023-05-11
Proposal to add sources to ChatMessage

EniasCailliau:
This PR adds sources to ChatMessage so the Q&A agent can include sources in its response. 

I introduced a definition of Document which is compatible with LangChain so we can easily upgrade our existing q&a agents. 

We can easily extend/change the definition of a source document once we have a better view on our agents and docstore.


dkolas (approved):
Question below about who / defaults

dkolas:
Closing, per slack discussion, since role functionality is now on Block directly.

PR #359, opened 2023-05-15
Experiments in tool implementations

eob:


eob:
Closing.
* The synchronous bits of this are in main
* The async bits are not; we're going to pursue a few other options before circling back to evaluate.

PR #360, opened 2023-05-15
AgentContext WIP

dkolas:


src/steamship/agents/context.py
```diff
@@ -0,0 +1,64 @@
+from abc import ABC, abstractmethod
+from typing import List, Optional
+
+from pydantic import BaseModel
+
+from steamship import Block
+from steamship.data.tags.tag_constants import RoleTag
+
+
+class AgentContext(BaseModel, ABC):
```
douglas-reid, line 10: @dkolas where, in your estimation, do callbacks and monitoring hooks belong in this in this interface?

douglas-reid (commented):
One thought: This appears to tie `AgentContext` directly to wrapping memory, vs as something that holds and provide access to all sorts of things, including memory. Is that, potentially, going to be limiting?

In other words, do we think:

```python
class AgentContext():

   @staticmethod
   def with_value(parent AgentContext, key: str, value: Any) -> AgentContext:
       pass
```

or even:

```python
class AgentContext():

   def message_history() -> MessageHistory:
      pass

   def key_value_store() -> KeyValueStore:
       pass

   def vector_store(...) -> VectorStore:
       pass
```

is not opinionated enough compared to:

```python
class AgentContext():
   def add_memory(...):
       pass

   def get_memory_by_key(...):
       pass
```

src/steamship/agents/context.py
```diff
@@ -0,0 +1,64 @@
...
+    #   if we don't like this we can split it out
+
+    @abstractmethod
+    @staticmethod
+    def get_or_create_context(user_id: str) -> "AgentContext":
+        """Get or create a new context for a particular end-user"""
+        pass
+
+    # Chat history (short term memory)
+    # This could be stored in a File object
```
eob, line 27: I also wonder if we want to parse out and save the Agent's scratch pads in file objects for structured logging of ReAct, tool invocation, etc.

src/steamship/agents/context.py
```diff
@@ -0,0 +1,64 @@
...
+        """Add a bit of conversation to the history."""
+        pass
+
+    def add_user_message(self, message: Block):
+        self.add_message(message, RoleTag.USER)
+
+    def add_system_message(self, message: Block):
+        self.add_message(message, RoleTag.SYSTEM)
+
+    def add_agent_message(self, message: Block):
```
eob, line 45: We might want to extend to support multiple agents (e.g. things like AutoGPT where multiple agents are working together on something)

src/steamship/agents/context.py
eob: It feels pretty comfortable for me to imagine them being in this object. In the good old days of wordpress, IIRC they just had a big flat namespace of hooks that you could even override even though some were "system" hooks.

We could have something similar -- a hook registry that the owning packageservice could even add hooks to -> or even hooks that called between agents of a team!

dkolas:
Closing per coalescing around other branch

PR #361, opened 2023-05-15
DO NOT MERGE: Playing with Agents and Packages

douglas-reid:
DO NOT MERGE

This modifies the `agent.py` to explore one possible way of using agents with context in a Telegram Bot.

DO NOT MERGE

### DIFF
--- a/src/steamship/agents/agents.py
+++ b/src/steamship/agents/agents.py
@@ -1,12 +1,11 @@
-import uuid
 from abc import ABC, abstractmethod
 from typing import Any, List, Optional
 
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 
-from steamship import Block, File, Steamship, Tag, Task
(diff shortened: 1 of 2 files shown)


src/steamship/agents/agents.py
```diff
@@ -151,37 +119,175 @@ def execute_action(self, action: Action, context: AgentContext):
...
+                        block,
+                        chat_id=incoming_message.get_chat_id(),
+                        message_id=incoming_message.get_message_id(),
+                    )
+                )
+
+        return responses
+
+
+class DalleTool(Tool):
```
eob, line 250: @douglas-reid before you implement sketched of these these, take a look at my open PR in case you want to borrow code!

eob:
👍 🚢 🚀  I'm digging this style of putting the pieces together. It feels really flexible, but also without too much overhead.

PR #362, opened 2023-05-17
wip: the most wip of all wips that's ever been wip'd

douglas-reid:


src/steamship/experimental/agent_service/agent_service.py
```diff
@@ -0,0 +1,382 @@
...
+
+class ImageTool(SyncTool, ABC):
+    pass
+
+
+class ScheduleMessageTool(SyncTool, ABC):
+    pass
+
+
+class Action:
```
eob, line 127: I really love this metaphor, and it's something that my branch completely lacks and wish it had.

eob: @douglas-reid am I taking this in a different direction than you're intending it when I start to think of a ChatHistory as having blocks tagged with `Action` in addition to `Message`?

The `Action` block have the JSON-serialized version of this object?

douglas-reid: @eob i couldn't decide if `Action` persists, or not, but I don't think that's off base. that's something i started to think about after i sorta got things ripped apart: a la if you reuse a ChatHistory concept, *just for localized agent history/scratchpad*, can we dump the history via blocks with tags via something like `Action`?

src/steamship/experimental/agent_service/agent_service.py
```diff
@@ -0,0 +1,522 @@
...
+            )
+        else:
+            tool_binding = action.tool_binding
+            output_blocks = tool_binding.tool.run(tool_binding.inputs, context)
+            step = (tool_binding, output_blocks)
+            context.completed_steps.append(step)
+            self.upsert_context(context)
+            return action
+
+    @post("_steamship/run")
```
eob, line 358: +1 so many times to what you're implying with `_steamship/run`

PR #363, opened 2023-05-17
AGENTS! (but really just Tools)

dkolas:
Here's a first PR on agents.

This PR introduces the v0 version of Tools, with a run method and a post-processing method.

AgentContext is just a stub.


eob:
@dkolas @dkolas @EniasCailliau So so excited for the work going on right now 🙌. To first commit! 

image

PR #364, opened 2023-05-17
[WIP] Unit Tests to experiment with some desired tasking properties AgentService may require

eob:
## Running Suggestion

Run the unit test in `test_proto_agent_service.py` and then view the link below after giving ES a few seconds to get the latest logs.

https://steamship.kb.us-east-2.aws.elastic-cloud.com:9243/app/logs/stream?flyoutOptions=(flyoutId:!n,flyoutVisibility:hidden,surroundingLogsId:!n)&logPosition=(end:now,position:(tiebreaker:149866,time:1684379794910),start:now-1d,streamLive:!f)&logFilter=(language:kuery,query:ToolThatDynamicallyPlansMoreAsyncWork)

You'll see the `ToolThatDynamicallyPlansMoreAsyncWork` having planned a dynamic workflow --- after execution began ---that involves multiple tools in which data is ferried from one to the other.

### .. But the unit test fails

The unit test fails because async package invocations Base64 encode their output, but I haven't base64 DECODED properly everywhere it's consumed. But I'm pausing here because it's observable in the Elastic logs that data is flowing as we would hope it would. If folks think this approach is worth pursuing, we can address these encoding issues.

### .. So how does it work?

I tried to put tons of comments in the files in the PR. It's an evolution of the same design in the "ToolPipeline" PR but it focuses less on the whole Tool-side of things and more on:

- What nouns we might consider (`ToolAction` and `ToolData`)
- How PackageServices might work (`enqueue_action`, `run_action`)
- How tools might request future work (`context.run_tool`)
- How we might, Tensorflow style, have a `ToolData` class which can equally be ACTUAL DATA or a PROMISE OF DATA, enabling the developer to write code which uses either, and the system makes sure the right thing happens.


tests/assets/packages/agent_service_prototype.py
```diff
@@ -0,0 +1,290 @@
...
+
+        # Note how the ToolData object also has a carve-out for a file_id. This isn't yet pursued in the unit
+        # tests on this branch but added to illustrate the map-reduce scenario: in which the initiator of the map
+        # operation synchronously (1) creates a file. and (2) spawns TONS of tasks. then (3) writes those task IDs
+        # to the file. THEN it schedules a future ToolAction that depends on those tasks, and takes as input the ToolData
+        # defined by that file's ID. That means that next ToolAction (1) only happens when the map operations are done, and
+        # (2) is passed, automatically, as its input, a set of blocks containing the UUIDs of the tasks with the results.
+        dymamic_plan_step_2 = ToolAction(
+            tool_name="ToolThatAlwaysResponseAsynchronouslyAndDynamically",
+            input=dyamic_plan_step_1_output.output,
```
douglas-reid, line 92: this style _appears_ to assume the following:
- the entirety of a `Tool`s input in contained in the output of preceding tool.
- we must know the mapping of inputs -> outputs between `Tools` up front.

Is that correct?

If so, how would we work around those limitations?

Context: the advantage of a _distinct_ planner is that it can use the whole execution context to decide on what the input for the next tool in the chain should be.

douglas-reid (commented):
@eob Thanks for putting this together! I appreciate your willingness and flexibility to humor my concerns.

I have some questions based on a quick read of this code. I've added them as comments. They may/may not be well-formed.

At a high-level, I'm concerned, upon reading it, that this feels like it diverges (somewhat afield) from the understanding of `Tools` and `Agents` that I have distilled from AI chatter. I'll try to outline my thinking here.

- Neither `LangChain` nor `HuggingFace` have `Tools` that call other `Tools` directly (again, as I've internalized their implementations). Saying that `Tools` perform a distinct unit of work, but that work is not about planning more work (at the same "level of execution", at least) is a constraining simplification, but it has advantages (conceptual complexity reduction, easier testing/debugging, etc.). I worry here that we are offering something that will be perceived as funky and different and not what people are asking for / building on.

- How do tools that use other tools configure them separately from still other tools?  The lookups are by name, right? It feels like I'm missing something there in my quick read. If tools can not be configured separately per their usage, I worry a bit about conflicts and/or explosion of objects (bad example: can i dynamically change image size in DALL-E?). 

- I don't see a direct way to hook in any sort of open-ended LLM-driven planning (there doesn't seem to be a pluggable control loop -- have I missed it?). Is the way to do this to require that  a `ReACTTool` *always* add as a final step in its `run()` an invocation of itself a Task running on the output?  That seems potentially problematic -- is it possible (as long as there is a terminating case)?

To push on an LLM-driven loop, what would that look like:

```python

class ReActTool(Tool):
   # assumes existence of the following tools:
   # - "openai-tool" (an LLM tool, with access to tool registry, and an output parser for blockification in post_process)
   # - "tool-executor" (something that gets a Tool name from a block and then executes it)

   def run(self, inputs: List[Block], context: AgentContext):
     if is_final(inputs):
         return inputs

      prompt = prompt_from_blocks(inputs)
      get_tool_and_input_strings = ToolAction("openai-tool", ToolData(prompt))
      llm_future = context.invoke_action(get_tool_and_input_strings)

      tool_executor = ToolAction("tool-executor", ToolData(llm_future.output))
      tool_future = context.invoke_action(tool_executor)

      react = ToolAction("react", ToolData(tool_future.output))
      react_future = context.invoke_action(react)

      return react_future.task
```

Is that right?  If so, doesn't the need for `tool_executor` Tool feel a bit like design smell?  If not, how would this work?

---

I think I'd like to clearly lay out the advantages of the proposed approach vs saying that:
- `AgentService` plans, schedules, and executes `Tool`s using a `Planner`
- `Tools` cannot use other `Tools` (directly -- must be mediated via an Agent/Package).
- If you wish to have a hybrid (or more complex) execution plan, then you *must* use separate agent instances.
- Agent instances can be `Tools`, but importantly, all scheduling, sub-tool management occurs entirely within their scope -- and unknown to the Agent using them as a Tool.

In this way, to support a podcast audio-summarizer and a dalle-generator in the same `TelegramBot`, I might have *two* AgentServices:

1. Root Agent, with an LLM planner that can schedule two types of tools (podcast-audio-summarizer-tool and dalle).
1. Child Agent, with a Pipeline planner that can schedule four different type of tools (search, podcast downloader, speech-to-text, summarizer)

Given a "summarize podcasts" prompt, RootAgent's LLM planner suggests using "podcase-audio-summarizer-tool. RootAgent calls `run()` on that Tool. That tool executes a `client.use().invoke() on the Package for the Child Agent` and returns the `Task`.  The RootAgent sees that it got a `Task`, calls `invoke_later` on itself saying "wake me when this child task is done". 

ChildAgent then gets invoked, begins its plan, selects the Search tool, calls `run`, gets a `Task`, and updates its running task with the new dependency and suspends until the grandchild `Task` completes.  _I think_ that is do-able, but I'm not clear on adding deps to a running task and suspending (it seems within the realm of possibilities).

eob:
Closing for the time being; this PR primarily is just an interesting convo / design reference going forward so trying to clean up the open PRs view

PR #365, opened 2023-05-18
Basic example of LLM-based control flow. Useful to ask ourselves: What's our +1 iteration from LangChain. 

EniasCailliau:


src/steamship/agents/llm.py
```diff
@@ -0,0 +1,36 @@
+from abc import ABC, abstractmethod
```
eob, line 1: A few thoughts on the `LLM` abstraction above merely `use_plugin`:

* I think it'll serve us really well in the Agent & Tools platform because it will help us curate a more "it always works" collection of options over-top the underlying plugin infra.
* I even think we might find ourselves (downstream..) wanting to have the defaults stored in some `settings` file in the Workspace that the user can override via API / Web Interface. E.g. you should be able to call the `switch_llm()` method that changes from gpt-3.5 to 4 to cohere, etc etc
* It also seems like a good way to enforce some kind of tracing / agent-specific logging.

Basically a long-winded way to say: two thumbs up to this direction from my end!

src/steamship/agents/context.py
```diff
@@ -0,0 +1,39 @@
...
+
+from pydantic import BaseModel
+
+from steamship import Steamship
+from steamship.experimental.transports.chat import ChatMessage
+
+
+class ChatHistory(BaseModel):
+    chat_id: str = ""  # chat history is scoped by a chat_id. chat_id is unique across user_id's
+    messages: List[str] = []  # Represents a database that can select & sort messages
```
eob, line 11: I think it's worth it to make `Block` the one true message type. There's so much that it instantly resolves:

- How we store it
- How we support multiple mime types
- How we annotate things atop it
- Etc etc

For the rest of the world that doesn't speak in blocks (e.g. steamship-langchain) we just do whatever conversion is necessary in the adapter library.

dkolas (commented):
I think this is a great base for our first agent release.

src/steamship/agents/context.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def get_history(self, message: ChatMessage, context):
```
eob, line 13: nit (I know this is just an interface sketch) but I think we probably want a `k: Int = 20` var or something

src/steamship/agents/context.py
```diff
@@ -0,0 +1,39 @@
...
+    def clear(self):
+        self.messages = []
+
+
+class AgentContext(BaseModel):
+    client: Steamship
+    chat_history: Optional[ChatHistory] = ChatHistory()
+    tracing: Optional[Dict[str, Any]] = []
+    scratchpad: Optional[List[str]] = []
+    new_message: ChatMessage
```
dkolas, line 31: Do we want a vector store here?  Is it separate from ChatHistory, or a different form thereof?

src/steamship/agents/context.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class ChatHistory(BaseModel):
+    chat_id: str = ""  # chat history is scoped by a chat_id. chat_id is unique across user_id's
+    messages: List[str] = []  # Represents a database that can select & sort messages
+
+    def get_history(self, message: ChatMessage, context):
+        return self.messages
+
+    def add_user_message(self, message: ChatMessage):
```
eob, line 16: nit, but maybe broader design principle -- I think we might want to always work in `List[Block]` instead of just `Block` just so that we avoid having to switch back and forth in different places in the code. This feels like on the spectrum of aesthetic preference though

src/steamship/agents/context.py
```diff
@@ -0,0 +1,39 @@
+from typing import Dict, Any, List, Optional
+
+from pydantic import BaseModel
+
+from steamship import Steamship
+from steamship.experimental.transports.chat import ChatMessage
+
+
+class ChatHistory(BaseModel):
```
eob, line 9: Is the assumption that this is backed by a File? Just thinking out loud:

It feels like there's value in this standing alone as a (non-abstract?) base class, but maybe having a subclass that's FileTrackedChatHistory, etc?

That might support the idea that there are different chat histories. E.g. the VectorStoreBackedChatHistory, FileChatHistory, etc.

The ChatHistory might have a set of child chatHistories it pushes any updates to

src/steamship/agents/context.py
eob, line 31: What's new_message?

src/steamship/agents/agent.py
```diff
@@ -0,0 +1,138 @@
...
+            new_message=context.new_message.text,
+            chat_history_str="\n".join(context.chat_history.get_history(message=context.new_message, context=context)),
+            scratchpad_str="\n".join(context.scratchpad))
+
+        blocks = self.llm.complete(prompt)
+
+        return self._extract_action(blocks, context)
+
+    def should_continue(self, context: AgentContext) -> bool:
+        return context.tracing.get("n_iterations", 0) < 10 and context.tracing.get("time_elapsed", 0) < 100
```
eob, line 70: I really like how this is connected to tracing as the way to implement it --- some notion of "developer-extensible data plane" feels like an awesome abstraction.

src/steamship/agents/context.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+class AgentContext(BaseModel):
+    client: Steamship
+    chat_history: Optional[ChatHistory] = ChatHistory()
+    tracing: Optional[Dict[str, Any]] = []
+    scratchpad: Optional[List[str]] = []
+    new_message: ChatMessage
+
+    def emit(self, messages: List[ChatMessage]):
```
eob, line 33: Unsure if this is specifically what you were suggesting here, but FWIW I like the idea that there's an `emit` function (on the context?) that allows -- in theory -- anything to emit a message any time.

I also think that some standard contract of "Tools run, the agent observes their results, the agent responds" is useful, but just permitting the possibility that intermediate messages may be appropriate feels like a really great door to keep open.

src/steamship/agents/context.py
EniasCailliau: Yea, that's why I've included ChatHistory so we could have the flexibility to add a vector store in there. Not sure if ChatHistory contains a vector store by default, or if that's a subclass of ChatHistory.

src/steamship/agents/context.py
EniasCailliau: +1. str was just a shortcut for me to get it working. Going to copy code from Doug now to get multi-modal working in this PR.

src/steamship/agents/context.py
EniasCailliau: It's the last message. Removed it for now :)

src/steamship/agents/context.py
EniasCailliau: I agree. As a dev I felt comfortable with the notion that a Tool could just send a message to close the loop without having to return to the Agent. 

Flow that is now possible: 

Agent decides to execute long, asyc task -> Task gets executed -> Task ends by sending a message to user. 

VS 

Agent decides to execute long, asyc task -> Task gets executed -> Task ends by returning to Agent -> Agent has to know what to send to the user based on its context.

PR #366, opened 2023-05-18
TOOLS! 🧰🛠️🔧

eob:
# Tools & ToolREPL

## Tools

This PR introduces a livery of tools collected form the whole team's prior prototyping efforts. 

A few notes on simplifications that have been made and corners that have been cut:

* Every tool runs synchronously at the moment. Big comment markers were added at the cut-points to re-add async running.
* Any tool that was a "Tool of Tools" has been excluded. For now we're going to see if we can accomplish things like Pipelines using Planners instead of Tools.
* Any tool that was a "Invoke a Package" or "Schedule a future Task" tool has been excluded. There are amazing. But probably represent more complexity that feels like an easy merge at the current moment. 

## Steamship REPL

This is based on the REPL @douglas-reid contributed to the hackathon-stater project.

In any `my_tool.py` file, one can simply write:

```python
if __name__ == "__main__":
    ToolREPL(MyTool()).run()
```

To get a REPL that lets them use the tool.


dkolas (approved):
Looks useful!

PR #367, opened 2023-05-19
feat: AgentService & Action Interface Stubs

eob:
This PR collects a few bare-bones common elements of @EniasCailliau and @douglas-reid's agent WIP PRs. 

The goal is to nail down an incredibly simplistic set of base classes that appear to be in common across all threads of prototyping, so that we increase the surface area of shared dependency.

There are two decisions in this PR which might not be shared that I'll point out here; proposal is for us to Replit-hack together and then merge this PR in:

- `List[Block]` as the I/O format (as opposed to `str` or `Any` or `Block`)
- `SpeakAction` as a distinct action that's possible, even though `FinishAction` also has a `messages` field

### DIFF
--- a/src/steamship/agents/action.py
+++ b/src/steamship/agents/action.py
@@ -1,0 +1,36 @@
+import abc
+from typing import List
+
+from pydantic import BaseModel
+
+from steamship import Block
+
+
+@abc.ABC
+class Action(BaseModel):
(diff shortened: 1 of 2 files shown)


eob:
Closing; unnecessary since after merge of Doug's branch! 🚀

PR #370, opened 2023-05-19
New parser design

EniasCailliau:
This PR changes the interface of the parser and attempts to simplify its structure. The new interface makes the parser responsible for parsing LLM output into either an Action or a FinalAction. 

1. Parser gets tools at init time vs at runtime 
2. Parser produces Action or FinalAction by itself 
3. Actions contain tools instead of tool references 

Other bits in this PR: 

- More classes inherit from BaseModel now
- Rename OpenAIReACTPlanner to ReACTPlanner. I will be pushing a PR with an LLM interface soon. I think this reactplanner will work across other LLM vendors, even if we decide an LLM interface is a bad idea.
- pre-commit linting 😇


src/steamship/agents/parsers/llm.py
```diff
@@ -1,37 +1,33 @@
...
-
-        print(f"Action tool: {tool or 'no tool'}")
-        return tool, blocks
+    tools_lookup_dict: Optional[Dict[str, BaseTool]] = None
+
+    def __init__(self, **kwargs):
+        tools_lookup_dict = {tool.name: tool for tool in kwargs.pop("tools", None)}
+        super().__init__(tools_lookup_dict=tools_lookup_dict, **kwargs)
+
+    def parse(self, text: str, context: AgentContext) -> Union[Action, FinishAction]:
```
eob, line 17: Not a blocking comment:

I've been wondering about `Union[Action, FinishAction]`

For future discussion -- do folks se the world as only having two types of actions from the python class perspective, or a ton of base classes and `Action` just stands in as the one base-class to represent them all?

eob (approved):
LGTM.

All of this seems fine to me; FWIW I don't rate myself as having a high-value opinion on this PR though -- I haven't written a React parser before. @douglas-reid's thoughts far better than mine here.

PR #372, opened 2023-05-19
feat: Chat Message (moving from experimental)

eob:
This it the class that the existing suite of Telegram bots have been using to communicate with Telegram.

It's a wrapper around block with helper functions for recording message_id and chat_id in a standardized way. It uses tags to do this.


eob:
From convo in slack re `who` field:

It was there because the chat widget wants to see it; else doesn't know who the message is from.

- LC uses `role`
- WhatApp uses `from`

cc @EniasCailliau @dkolas

PR #373, opened 2023-05-22
Add LLM interface

EniasCailliau:
This PR: 

* Adds a LLM interface 
* Move scratchpad construction to a private function


src/steamship/agents/example_agents/my_assistant.py
```diff
@@ -15,11 +16,13 @@
 class MyAssistant(AgentService):
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
-        self.tools = [
-            SearchTool(),
-            GenerateImageTool(),
-        ]
-        self.planner = OpenAIReACTPlanner()
+        self.planner = ReACTPlanner(
```
eob, line 19: 🚀🚀🚀

src/steamship/agents/llm/openai.py
```diff
@@ -0,0 +1,19 @@
...
+
+from steamship import Block, PluginInstance, Steamship
+from steamship.agents.base import LLM
+
+
+class OpenAI(LLM):
+    generator: PluginInstance
+    client: Steamship
+
+    def __init__(self, client, *args, **kwargs):
```
eob, line 11: We can leave this as a follow-on, but I think it's hugely valuable to permit plumbing in the model name so that folks can select 3.5-turbo if they want

src/steamship/agents/planner/react.py
```diff
@@ -87,12 +78,18 @@ def plan(self, context: AgentContext) -> Action:
...
-        task = gpt4.generate(text=prompt, options={"stop": "Observation:"})
-        task.wait()
-        # here, we assume that the response will always be a single block of text.
-        # print(f"LLM output: {task.output.blocks} \n")
-        return self.output_parser.parse(task.output.blocks[0].text, context)
+    def _construct_scratchpad(self, context):
+        steps = []
+        for action in context.completed_steps:
+            steps.append(
+                "Thought: Do I need to use a tool? Yes"
```
eob, line 88: It would be so cool to start logging this in a structured way right form V1 -- that way we can visualize the thought process.

eob: Also do  we need \n chars here?

PR #374, opened 2023-05-22
Context and ChatHistory

dkolas:
This PR makes ChatHistory part of Context, backed by File objects.

- ChatFile is moved from experimental to agent and renamed ChatHistory
- Borrowed liberally from #372 in adding key tags to blocks in a repeatable manner
- Added helpers to Block to eliminate the need for ChatMessage
- Yanked context cache out of AgentService, now managed by context itself

Could really use some help on:

- @eob I don't understand if / why all the 'upsert context' calls were necessary in `AgentService` - maybe they're not now that it's File-backed, but this may need discussion
- Can't figure out how to cleanly resolve the circular dependency between `AgentContext` and `Action` (see line 75 in AgentContext)


src/steamship/agents/context/chathistory.py
```diff
@@ -0,0 +1,225 @@
...
+        tags: List[Tag] = None,
+        content: Union[str, bytes] = None,
+        url: Optional[str] = None,
+        mime_type: Optional[MimeTypes] = None,
+    ) -> Block:
+        """Append a new block to this with content provided by the end-user."""
+        tags = tags or []
+        tags.append(
+            Tag(
+                kind=TagKind.CHAT, name=ChatTag.ROLE, value={TagValueKey.STRING_VALUE: RoleTag.USER}
```
eob, line 86: This suggestion feels like slight early-optimization, but I wonder if we want to additionally hard-code in -- right from the start -- the ability to say WHICH user or WHICH system/agent. Since eventually we might end up finding we've got a ChatHistory that's in:

* a chat room with multiple humans
* an AgentService with multiple agents

src/steamship/agents/context/chathistory.py
```diff
@@ -0,0 +1,225 @@
...
+
+    @property
+    def last_system_message(self) -> Optional[Block]:
+        for block in self.file.blocks[::-1]:
+            if block.chat_role == RoleTag.SYSTEM:
+                return block
+        return None
+
+    @property
+    def last_agent_message(self) -> Optional[Block]:
```
eob, line 150: These helpers feel super useful. 

I think for tool -> tool -> tool pipelines, we might want one that's `last_message` which loads any of a set of roles? E.g. the tool's input might be the last tool's output OR last user input, whichever is later?

src/steamship/agents/context/chathistory.py
```diff
@@ -0,0 +1,225 @@
...
+    # create a GPT-3 encoder instance
+    enc = tiktoken.get_encoding(tiktoken_encoder)
+    # encode the text using the GPT-3 encoder
+    tokenized_text = enc.encode(block.text)
+    # calculate the number of tokens in the encoded text
+    return len(tokenized_text)
+
+
+# TODO: abstract this into one of N possible strategies
+def filter_blocks_for_prompt_length(max_tokens: int, blocks: List[Block]) -> List[int]:
```
eob, line 190: Ditto re utils package

src/steamship/agents/context/context.py
```diff
@@ -0,0 +1,100 @@
...
+
+    # User<->Package chat history (NOT Agent<-->Tool history)
+    # NB: This is distinct from any sort of history related to agent execution
+    # Primarily used to ferry prompts into agent executions AND
+    # to allow for some sort of referential lookup of context as input to the agent
+    chat_history: ChatHistory
+
+    # Provides the overall goal of the agent
+    @property
+    def initial_prompt(self) -> Optional[Block]:
```
eob, line 57: +1 to this style of plumbing.

src/steamship/agents/context/context.py
```diff
@@ -0,0 +1,100 @@
...
+        return self.chat_history.initial_system_prompt
+
+    # instead of saving the execution history via File as text, this chooses to represent
```
eob, line 60: I know we've got a TON of different ideas on how to best accomplish this, so rather than jump in advocating for any particular one of them... I wanted to jump in and say:

- I think we're within the threshold of: "Any we've been discussing sound fine, and we can change it later -- even if it breaks backward compatibility." 

Feels like one of those totally OK "cost of iteration / growing" things.

src/steamship/data/block.py
```diff
@@ -188,6 +199,59 @@ def is_video(self):
         """Return whether this is a video Block."""
         return self.mime_type in [MimeTypes.MP4_VIDEO, MimeTypes.WEBM_VIDEO]
 
+    @property
```
eob, line 202: Just want to actively +1 that I'm totally OK w/ us adding these kinds of helpers directly to Block

eob (approved):
LGTM! I think this is a huge step forward! 

I know there are still some open design threads in here, but fwiw I think this is a fantastic base to build upon, even if we don't resolve those threads in this PR 🚀

eob (dismissed):
LGTM! I think this is a huge step forward! 

I know there are still some open design threads in here, but fwiw I think this is a fantastic base to build upon, even if we don't resolve those threads in this PR 🚀

src/steamship/agents/context/chathistory.py
```diff
@@ -0,0 +1,225 @@
...
+                tags=tags,
+            )
+        return ChatHistory(file)
+
+    def append_user_message(
+        self,
+        text: str = None,
+        tags: List[Tag] = None,
+        content: Union[str, bytes] = None,
+        url: Optional[str] = None,
```
EniasCailliau, line 79: Would this be covered by replacing text by `message: Union[str|block]`?

src/steamship/agents/context/chathistory.py
```diff
@@ -0,0 +1,225 @@
...
+        """Append a new block to this with content provided by the agent, i.e., results from the assistant."""
+        tags = tags or []
+        tags.append(
+            Tag(
+                kind=TagKind.CHAT,
+                name=ChatTag.ROLE,
+                value={TagValueKey.STRING_VALUE: RoleTag.ASSISTANT},
+            )
+        )
+        return self.file.append_block(
```
EniasCailliau, line 131: nit: some code duplication here. Would wrap this into _append_message(role="") method

src/steamship/agents/context/chathistory.py
```diff
@@ -0,0 +1,225 @@
...
+            text=text, tags=tags, content=content, url=url, mime_type=mime_type
+        )
+
+    @property
+    def last_user_message(self) -> Optional[Block]:
```
EniasCailliau, line 136: The implementation does not match my expectations when I read the function title

This return the last user message IF it matches the RoleTag. 

I expected it to return the last user message, pseudocode: 

```
[block for block in block.chat_role == RoleTag.USER][::-1]
```

src/steamship/agents/context/context.py
```diff
@@ -0,0 +1,100 @@
...
+    # for each of these bits, we probably need to save only IDs, and rehydrate properly;
+    # likely requires an override of `dict()` and `parse_obj()`
+    completed_steps: AgentSteps = []
+
+    # This supports parallel tool execution. But I think that should
+    # be a future iteration, as it complicates things quite a bit.
+    #
+    # as above this likely could be "discovered" from `agent_history: ChatFile` and tags
+    # dynamically, and be more "Steamship-native". Perhaps in a second pass.
+    in_progress: List[
```
EniasCailliau, line 74: Does this need to be backed by a File too?

EniasCailliau (approved):

src/steamship/agents/context/chathistory.py
```diff
@@ -0,0 +1,225 @@
...
+from steamship import File, MimeTypes, SteamshipError
+from steamship.base.client import Client
+from steamship.data import TagKind
+from steamship.data.block import Block
+from steamship.data.tags import Tag
+from steamship.data.tags.tag_constants import ChatTag, DocTag, RoleTag, TagValueKey
+
+
+class ChatHistory:
+    class Config:
```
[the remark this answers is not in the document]
dkolas: I've just removed the Config for now.

src/steamship/agents/context/chathistory.py
dkolas: I think my code is doing this, right?  It iterates through all the blocks from last to first and returns the first one it encounters that has the role?

src/steamship/agents/context/context.py
dkolas: I think it will when tool chains are async, but can be not right now.  Do you agree?

src/steamship/agents/context/chathistory.py
```diff
@@ -0,0 +1,225 @@
...
+            raise SteamshipError(
+                "Multiple ChatHistory objects have been created in this workspace with these context keys."
+            )
+
+    @staticmethod
+    def get_or_create(
+        client: Client,
+        context_keys: Dict[str, str],
+        tags: List[Tag] = None,
+        initial_system_prompt: Optional[str] = None,
```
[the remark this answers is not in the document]
dkolas: Removed it for now

src/steamship/agents/context/chathistory.py
douglas-reid: @dkolas i would think about this method (and others) potential taking a param that controls the number of messages.  a la: `last_user_messages(self, n: int = 1) -> Optional[Block]`.

src/steamship/agents/context/chathistory.py
```diff
@@ -0,0 +1,161 @@
...
+            )
+        )
+        return self.file.append_block(
+            text=text, tags=tags, content=content, url=url, mime_type=mime_type
+        )
+
+    @property
+    def last_user_message(self) -> Optional[Block]:
+        for block in self.file.blocks[::-1]:
+            if block.chat_role == RoleTag.USER:
```
douglas-reid, line 123: it doesn't matter much, but i'd think about refactoring these three methods to use a `last_messages(self, role_tag: str, n: int = 1) -> Optional[Block]` method, so that:

```
def last_system_messages(self, n: int = 1) -> Optional[Block]:
  return self.last_messages(role_tag: RoleTag.SYSTEM, n=n)
```

douglas-reid (commented):
Some initial comments.

PR #377, opened 2023-05-23
Public Block and File data

dkolas:
This PR provides capability to make Blocks and Files's byte data publicly accessible.  You can:

- Create a File or Block with `public_data=True`, allowing the content to be fetched at `<api>/[block|file]/<id>/raw`
- View the `public_data` status of a File or Block
- Flag a generate request so that the output is automatically made public
- convenient access to the full URL of the raw data

Not yet present:
- ability to update `public_data` on an existing File or Block (this is present on the API side)


eob (approved):
LGTM!

PR #378, opened 2023-05-23
Multi inheritance tests

dkolas:
Tests for how multi inheritance works for packages.

😄 - DIR operation finds decorated endpoint routes from both parents.
😦 - Instance init only calls the first parent's implementation.

Workarounds demonstrated in the tests.


eob:
Dave these are some nice tests -- if you rebase on `main` I think they'll start passing; Maybe we keep this PR just to add extra robustness around PackageService rather than close it w/o merge.

PR #381, opened 2023-05-23
Create TelegramAgent; remove ChatMessage

dkolas:
Rather than do multiple inheritance between AgentService and TelegramBot, this PR introduces TelegramAgentService which inherits from SteamshipWidgetAgentService which inherits from AgentService.  

With this, implementing a Telegram Agent should be defining the Agent itself.


src/steamship/experimental/package_starters/telegram_agent.py
```diff
@@ -0,0 +1,84 @@
...
+    def respond(self, **kwargs) -> InvocableResponse[str]:
+        """Endpoint implementing the Telegram WebHook contract. This is a PUBLIC endpoint since Telegram cannot pass a Bearer token."""
+
+        # TODO: must reject things not from the package
+        message = kwargs.get("message", {})
+        chat_id = message.get("chat", {}).get("id")
+        try:
+            incoming_message = self.telegram_transport.parse_inbound(message)
+            if incoming_message is not None:
+                context = AgentContext.get_or_create(self.client, context_keys={"chat_id": chat_id})
```
eob, line 49: I don't think this is something we need to do in this PR / by Friday, but I wonder if all transport-specific endpoints eventually route to the same common place where things like AgentContext gets instantiated?

src/steamship/experimental/package_starters/telegram_agent.py
```diff
@@ -0,0 +1,84 @@
...
+                context.chat_history.append_user_message(text=incoming_message.text)
+                if len(context.emit_funcs) == 0:
+                    context.emit_funcs.append(self.telegram_transport.send)
```
eob, line 52: This is really neat

eob (approved):
LGTM! One small "thinking about V2 comment"

PR #384, opened 2023-05-23
refactor: cleanup agent class objects

douglas-reid:
This PR represents an attempt to cleanup, document, and rationalize the existing set of classes in `agent` hierarchy. I also wanted to remove as much vestigial and unused surface (primarily related to async execution) in an attempt to simplify and reduce surface area for implementors.

As part of this PR, I updated the example assistant to show how synchronous agent responses could be done using `emitFns`. This then led to an update of the REPL to allow flexible identification of `AgentService` methods. I think we need more thought there.

Because we've introduced an `LLM` class, I took the `get_llm()` method that created ad-hoc ones from context off of the context and added helpers to create contexts with LLM instances. This was an attempt to rationalize usage within Tools, etc. I'm open to other approaches there.


src/steamship/data/block.py
```diff
@@ -252,6 +252,12 @@ def _one_time_set_tag(self, tag_kind: str, tag_name: str, string_value: str):
 
         self.tags.append(tag)
 
+    def as_llm_input(self) -> str:
```
eob, line 255: This is an awesome idea

src/steamship/agents/utils.py
```diff
@@ -0,0 +1,19 @@
+from typing import Optional
+
+from steamship.agents.schema import AgentContext
+from steamship.agents.schema.llm import LLM
+
+_LLM_KEY = "llm"
+
+
+def with_llm(llm: LLM, context: Optional[AgentContext] = None) -> AgentContext:
```
eob, line 9: From the PR message it sounds like this felt mostly like an odd shim to write, but I actually think this style may prove super useful as a way to think....

If things like "Which LLM am I using" become dynamic settings of an instance-in-progress (e.g. 'let's try Cohere out'), then this kind of chaining could let us pile on user settings atop reasonable defaults in a pretty elegant way

src/steamship/agents/service/agent_service.py
```diff
@@ -1,85 +1,46 @@
...
-
-        # at this stage in run_agent, we have no more running tasks. we have 0 or more completed steps.
-        # we must now select the next step
-        action = self.take_action(context)
+    def run_agent(self, agent: Agent, context: AgentContext):
+        action = agent.next_action(context=context)
         while not isinstance(action, FinishAction):
-            action = self.take_action(context)
-
+            # TODO: logging?
```
eob, line 37: Check out my weird, but I think workable (?), logging PR -- it proposes using structured logging to let us lift different views about agent/tool activity out of the standard logs.

src/steamship/agents/schema/tool.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class AgentContext(BaseModel):
+    """Placeholder to avoid circular dependency."""
+
+    # TODO(doug): refactor so this is not necessary
+    pass
+
+
+class Tool(BaseModel):
```
eob, line 16: Does this need to include ABC since below has a @abstractmethod?

src/steamship/agents/schema/action.py
```diff
@@ -0,0 +1,38 @@
...
+    """Tool used by this action."""
+
+    input: List[Block]
+    """Data provided directly to the Tool (full context is passed in run)."""
+
+    output: Optional[List[Block]] = []
+    """Any direct output produced by the Tool."""
+
+
+class AgentTool(Tool):
```
eob, line 25: nit/readability - I was confused by `AgentTool` as a name.

I think the idea is that we want `Tool` to be an ABC, but we also want something that suffices as a short-cut for `NoOp`?

Maybe we just call it `NoOpTool` (if that's what it is)?

src/steamship/agents/schema/action.py
```diff
@@ -0,0 +1,38 @@
+from typing import Any, List, Optional, Union
+
+from pydantic import BaseModel
+
+from steamship import Block, Task
+from steamship.agents.schema.tool import AgentContext, Tool
+
+
+class Action(BaseModel):
```
eob, line 9: Not important now, but thought pin to wonder if eventually `input` and `output` will benefit from being a proxy object which represents: "The ability to load blocks -- now or in the future".  Such that both bindings can be a persistence-friendly future on blocks.

eob (commented):
Looks really solid. A few minor comments within

src/steamship/agents/schema/action.py
dkolas: or FinishTool

dkolas (approved):
You are the hero we need, not the hero we deserve

PR #388, opened 2023-05-24
add: inputs to logging for tool

douglas-reid:
Added some more for context on what was being sent to individual tools.

### DIFF
--- a/src/steamship/agents/service/agent_service.py
+++ b/src/steamship/agents/service/agent_service.py
@@ -36,8 +36,9 @@ class AgentService(PackageService):
         action = agent.next_action(context=context)
         while not isinstance(action, FinishAction):
             # TODO: Arrive at a solid design for the details of this structured log object
+            inputs = ",".join([f'"{b.as_llm_input()}"' for b in action.input])
             logging.info(
-                f"Running Tool {action.tool.name}",
+                f"Running Tool {action.tool.name} ({inputs})",
                 extra={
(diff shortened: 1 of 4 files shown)


eob (approved):
These all look like good changes to me

src/steamship/agents/service/agent_service.py
```diff
@@ -29,15 +29,26 @@ def run_action(self, action: Action, context: AgentContext):
...
             context.completed_steps.append(action)
 
     def run_agent(self, agent: Agent, context: AgentContext):
         action = agent.next_action(context=context)
         while not isinstance(action, FinishAction):
             # TODO: Arrive at a solid design for the details of this structured log object
+            inputs = ",".join([f"{b.as_llm_input()}" for b in action.input])
             logging.info(
-                f"Running Tool {action.tool.name}",
+                f"Running Tool {action.tool.name} ({inputs})",
```
dkolas, line 51: Does this push user data to our log system?

dkolas (approved):
Great fixes! One thing - we may not want to log inputs / outputs at least at the `info` level, as that will store them in Elastic. We can set those to `debug` or come up with another strategy.

src/steamship/agents/service/agent_service.py
douglas-reid: potentially. the text of whatever the agent looks at, at least. so, when a tool is called with user input directly, then yes?

PR #389, opened 2023-05-25
CLI fix for python 3.11

dkolas:
In Python 3.11, the enum value string formatting is different, so you get output messages like this:

```
Deployment was successful. View and share your new DeployableType.PACKAGE here:

https://app.staging.steamship.com/DeployableType.PACKAGEs/telegram-agent-test-dk
```

This PR explicitly calls `.value`, which works in both Python 3.11 and below.

Note - Using anything beyond python 3.8, which is our current deployment target, could lead to other compatibility issues.  Nonetheless fixing/changing this seems like an easy win to me.


dkolas:
now this seems to be working in 3.11 without this fix?  I am confused.  closing this PR

eob:
@dkolas @douglas-reid I propose we consider this a green check and separately address the failed test.

It's basically testing something inherently probabilistic (embedding of various words + AKNN lookup) -- the "A" in "AKNN" being the element of stochastic behavior

douglas-reid:
> @dkolas @douglas-reid I propose we consider this a green check and separately address the failed test.
> 
> It's basically testing something inherently probabilistic (embedding of various words + AKNN lookup) -- the "A" in "AKNN" being the element of stochastic behavior

Filed issue: #423 to track. OK with ignoring and moving forward.

PR #392, opened 2023-05-26
JSON Object Generator Tool

eob:
This tool became super useful while building the podcast generator. It’s a demo of how to perform the task “Make me an object with this schema."


dkolas (dismissed):
Code looks good! Did you have an intended use case in mind?  I haven't figured it out quite yet, but that may just be my lack of imagination :)

eob:
@dkolas here's a "notes from the rough and tumble world of prompt engineering": this approach gets TERRIBLE results when I try it on a bunch of different use cases 😂

It's clean and elegant in theory, but the attention head doesn't seem to be able to track the implications of the header row of CSV or TSV well enough for it to behave the way that's desired.

I've rewritten it (in another project) to be JSON-based, in which the schema is repeated for every example to remind the LLM what fields are desired, and that seems to be working really well.

Will port that over and update here.

The example use case is:

"I want an example of how to generate a structured object"

src/steamship/agents/tools/text_generation/json_object_generator.py
```diff
@@ -0,0 +1,170 @@
...
+        clause = f'"{key}": "{value}"'
+        return clause
+
+    def object_json(self, schema: List[str], values: List[str]):
+        """Render a CSV-style header row and value list into a JSON object."""
+        clauses = []
+        for field, value in zip(schema, values):
+            clauses.append(self.kv_clause(field, value))
+
+        return "{" + ", ".join(clauses) + "}"
```
dkolas, line 98: Why are you rolling your own mini-encoder instead of creating a dict of strings and then calling `json.dumps`?

dkolas (approved):
Looks good, but one suggested simplification below

src/steamship/agents/tools/text_generation/tsv_row_generator_tool.py
```diff
@@ -0,0 +1,101 @@
...
+
+        Output
+        ------
+        output: List[Blocks]
+            A single block containing a new row of the table described by the tool's configuration.
+        """
+
+        # Compile the prompt based entirely on the tool configuration.
+        random.shuffle(self.example_rows)
+        tsv_rows = ["\t".join([str(field) for field in row]) for row in self.example_rows]
```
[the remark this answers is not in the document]
eob: This is somewhat overcome by the JSON replacement, but there's still relevant (quote-escaping) code in there so wanted to make sure to respond ---

This is 100% me trying to anticipate what's hard/easy for the LLM. I didn't do experiments to see if there was a difference

src/steamship/agents/tools/text_generation/tsv_row_generator_tool.py
```diff
@@ -0,0 +1,101 @@
...
+
+        prompt = self.rewrite_prompt.format(
+            table_description=self.table_description,
+            header_fields="\t".join(self.header_fields),
+            example_rows=tsv_block,
+            new_row_prefix_fields=new_row_prefix,
+        )
+
+        llm = get_llm(context)
+        return llm.complete(prompt, stop="\n")
```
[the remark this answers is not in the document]
eob: Yeah -- this was the huge problem with this version. It only had the right number of columns about 30% of the time.

The JSON re-implementation, at least on my test cases (podcast-related generation) works every time.

src/steamship/agents/tools/text_generation/json_object_generator.py
eob: I was worried about stable ordering. 

I don't want to just alphabetize it, because that takes away the ability to hard-code some of the fields for the new generation.

E.g. key order has to be [ HARD_CODED_FIELDS ] + [ GENERATED_FIELDS ]

Could def be missing an easier way to do it that I'm not seeing

eob: Hard-coded case is, e.g.

Generate me the object `{ story_genre, story_title, story_summary }` but with the `story_genre` fixed to `COMEDY`

So in the case, you need the "Generation row" to be this:

```
{"story_genre": "COMEDY"
```

rather than just

```
{
```

And you also need all of the examples to (I think?) be in the same order too to help the LLM see patterns

PR #394, opened 2023-05-30
Allow setting of public_data from block and file

dkolas:
The previous public_data PR did not include a method for altering the `public_data` status of a `Block` or `File`.  

This PR introduces `set_public_data` on both.


eob (approved):
LGTM!

PR #396, opened 2023-05-30
Package Mixins

dkolas:
This PR builds on the dynamic route registration of #393  and sketches the concept of an explicit Package Mixin, which is added to a PackageService in its init method, and adds routes to the package.

The goal is that things like Telegram Transport and Ask My X could be mixed into an agent without having to resort to multiple inheritance.

Right now, the mixins get to see the PackageService's config during instance_init.  This would mean that the PackageService's config would need to include the union of parameters for itself and all its mixins. I can't yet think of a better way.

### DIFF
--- a/src/steamship/experimental/transports/telegram.py
+++ b/src/steamship/experimental/transports/telegram.py
@@ -135,12 +135,13 @@ class TelegramTransport(Transport):
         if video_or_voice := (payload.get("voice") or payload.get("video_note")):
             file_id = video_or_voice.get("file_id")
             file_url = self._get_file_url(file_id)
-            return Block(
+            block = Block(
                 text=payload.get("text"),
                 url=file_url,
-                chat_id=str(chat_id),
-                message_id=str(message_id),
             )
+            block.set_chat_id(str(chat_id))
+            block.set_message_id(str(message_id))
+            return block
 
         # Some incoming messages (like the group join message) don't have message text.
         # Rather than throw an error, we just don't return a Block.


src/steamship/invocable/package_mixin.py
```diff
@@ -0,0 +1,9 @@
+from abc import ABC
+
+from steamship.invocable import Config
+
+
+class PackageMixin(ABC):
+    def instance_init(self, config: Config):
```
eob, line 7: Oooh yeah `instance_init` is a really nice example of why formalizing mixins is going to be really nice

tests/assets/packages/package_with_mixins.py
```diff
@@ -0,0 +1,30 @@
...
+    def test_mixin_route(self) -> InvocableResponse:
+        return InvocableResponse(data=f"mixin {self.suffix}")
+
+
+class PackageWithMixin(PackageService):
+    def __init__(self, **kwargs):
+        super(PackageWithMixin, self).__init__(**kwargs)
+        self.add_mixin(
+            TestMixin("yo")
+        )  # We add instances of mixins, not the class, in case we need to parameterize their inits
```
eob, line 26: yeah that makes sense to me too

tests/steamship_tests/app/unit/test_mixins.py
```diff
@@ -0,0 +1,45 @@
...
+    package = PackageWithMixin()
+    package.invocable_instance_init()
+
+    assert package.mixins[0].instance_init_called
+
+
+def test_two_conflicting_mixins_raises_error():
+    package = PackageWithMixin()
+    with pytest.raises(SteamshipError):
+        package.add_mixin(TestMixin("ahoy"))
```
eob, line 45: How do you feel about plumbing the `permit_overwrite_of_existing` variable all the way up to `add_mixin` but defaulting it to `False`? I don't think I feel strongly, but seemed like an obvious question to ask

eob (approved):
LGTM

tests/steamship_tests/app/unit/test_mixins.py
dkolas: I'm not sure how it would work.  The mixin will have its own internal state, which may be affected by the calling of its routes.  It assumes a pairing of that internal state and which routes it defines.  Permitting overwrite means being allowed to be overwritten, in which case bad things almost certainly start happening, right?

If you want to extend or change the behavior of a mixin, I think you'd want to subclass the mixin, and then mix THAT in :)

douglas-reid:
@dkolas thanks for tackling - sure beats multiple inheritance.

A few questions this triggered in my head:
- do you imagine a world in which mixin selection would be dynamic (say, based on environment, or otherwise)? 
- what guidance can we provide for (potential) mixin collisions?  if i add two mixins with overlapping routes, what is the expected outcome?

tests/steamship_tests/app/unit/test_mixins.py
eob: Yeah I think there's def enough murkiness there for the default to be: "you can't clobber unless you know what you're doing so much you can figure out how to clobber manually" :)

eob:
@douglas-reid @dkolas responding -- responding to @douglas-reid 's comment --

one thing that this makes me start imaging is (eventual) web control of transports.

[ wayne's world dream sequence ]

E.g. the agent instance page could have a "Connect to Chat" tab that lets you add and remove mixins -> this could just be setting values on some KVStore in the agent which drives which mix-ins to expose.

dkolas:
> @dkolas thanks for tackling - sure beats multiple inheritance.
> 
> A few questions this triggered in my head:
> 
> * do you imagine a world in which mixin selection would be dynamic (say, based on environment, or otherwise)?
> * what guidance can we provide for (potential) mixin collisions?  if i add two mixins with overlapping routes, what is the expected outcome?

I think it could already be dynamic based on package config. The init method has access to config, and could use that to decide whether to mix in a mixin.

Mixing in mixins with overlapping routes should throw an error. I've tested that it's an error, but I think there's room to make the error more specific ("route x from mixin y conflicts with route x from mixin z")

dkolas:
> > @dkolas thanks for tackling - sure beats multiple inheritance.
> > A few questions this triggered in my head:
> > 
> > * do you imagine a world in which mixin selection would be dynamic (say, based on environment, or otherwise)?
> > * what guidance can we provide for (potential) mixin collisions?  if i add two mixins with overlapping routes, what is the expected outcome?
> 
> I think it could already be dynamic based on package config. The init method has access to config, and could use that to decide whether to mix in a mixin.
> 
> Mixing in mixins with overlapping routes should throw an error. I've tested that it's an error, but I think there's room to make the error more specific ("route x from mixin y conflicts with route x from mixin z")

I did this now.  The error looks like:

```
"When attempting to add mixin TestMixin2, route POST test_mixin_route conflicted with already added route POST test_mixin_route on class TestMixin'"
```

PR #398, opened 2023-05-31
Vector Fact Learner

eob:
This PR:

1) Adds the `VectorSearchTool` base class
2) Refactors the `VectorSearchQATool` atop it
3) Introduces the `VectorSearchLearnerTool` tool, which learns facts
4) Introduces the `FactLearner` example agent which persistently learns facts and can answer questions about them from a vector store


src/steamship/agents/examples/fact_learner.py
```diff
@@ -0,0 +1,75 @@
...
+        self._agent = ReACTAgent(
+            tools=[
+                VectorSearchLearnerTool(),
+                VectorSearchQATool(),
+            ],
+            llm=OpenAI(self.client),
+        )
+
+    @post("prompt")
+    def prompt(self, prompt: str) -> str:
```
douglas-reid, line 36: one minor comment here: when playing with history/memory, i quickly discovered the utility of being able to take "context id" as a param in the example method. we should be sure to add whatever mechanism we agree on in that PR to this bit.

src/steamship/agents/examples/fact_learner.py
```diff
@@ -0,0 +1,75 @@
...
+        # AgentContexts serve to allow the AgentService to run agents
+        # with appropriate information about the desired tasking.
+        # Here, we create a new context on each prompt, and append the
+        # prompt to the message history stored in the context.
+        context_id = uuid.uuid4()
+        context = AgentContext.get_or_create(self.client, {"id": f"{context_id}"})
+        context.chat_history.append_user_message(prompt)
+
+        # Add the LLM
+        context = with_llm(context=context, llm=OpenAI(client=self.client))
```
douglas-reid, line 48: question: would this be a good place to play with different temp settings?

src/steamship/agents/tools/question_answering/vector_search_learner_tool.py
```diff
@@ -0,0 +1,63 @@
...
+        ------
+        output: List[Blocks]
+            A lit of blocks containing the answers.
+        """
+
+        output = []
+        for input_block in tool_input:
+            if input_block.is_text():
+                self.learn_sentence(input_block.text, context=context)
+                output.append(Block(text=f"I'll remember: {input_block.text}"))
```
douglas-reid, line 52: aside: is more pythonic to have `Block(text=...)` vs `Block.text(...)` or even `TextBlock(...)` ?  I don't have a good feel for that yet.

douglas-reid (approved):

src/steamship/agents/examples/fact_learner.py
eob: Yeah I really liked that element of those other PRs -- depending on the ordering of merge, I'll back-port it in!

src/steamship/agents/examples/fact_learner.py
eob: Unsure -- what would the nature of the playing be in this case? To show how different temps alter the way in which it answers the questions?

src/steamship/agents/tools/question_answering/vector_search_learner_tool.py
eob: I'm also unsure and have sort of drifted from style to style in different parts of the codebase. If you feel like you start to build a sense of which feels better, lmk. 

`Block.from_text(...)` feels kind of nice to me.

PR #399, opened 2023-05-31
Make metadata optional 

EniasCailliau:
Context: TelegramBot does not pass in metadata to the transport's `send` function. This PR explores the option to make metadata optional as we have 5 spots where we pass empty dicts as metadata.


src/steamship/experimental/transports/transport.py
```diff
@@ -63,7 +63,8 @@ def instance_deinit(self, *args, **kwargs):
         end = time.time()
         logging.info(f"Transport deinitialized in {end - start} seconds: {self.__class__.__name__}")
 
-    def send(self, blocks: List[Block], metadata: Dict[str, Any]):
+    def send(self, blocks: List[Block], metadata: Metadata):
+        metadata = metadata or {}
```
dkolas, line 67: Maybe I haven't had enough coffee yet, but I don't understand how this makes it optional.  Metadata is still defined as Dict[str, Any], right, not Optional[Dict[str, Any]]?

dkolas (dismissed):
The reason I initially didn't reference the Metadata class from Transport is that I didn't know if we wanted it to have dependencies on the Agent classes. 

Also see below - there's something I'm not understanding about how this makes it optional?

PR #401, opened 2023-05-31
Ask My Docs Mixin

eob:
This set of Mixins adds the Ask My Docs functionality into our codebase, along with unit tests (🚀).

It's implemented as a series of mixins:

- FileImporterMixin
- BlockifierMixin
- IndexerMixin

That are all tied together with:

- IndexerPipelineMixin


src/steamship/invocable/mixins/blockifier_mixin.py
```diff
@@ -19,6 +20,7 @@ def blockify(
         file_id: str,
         mime_type: Optional[MimeTypes] = None,
         blockifier_handle: Optional[str] = None,
+        after_task_id: Optional[str] = None,
     ) -> Task:
         """Blockify the file `file_id` using a curated set of defaults for the provided `mime_type`.
```
dkolas, line 25: This feels like a very thin wrapper around the normal call to file.blockify.  Does this content just belong in the main SDK rather than as a mixin?  Is there value in exposing the _route_ `blockify_file`, or is this just a step in the larger process?

src/steamship/invocable/mixins/indexer_mixin.py
```diff
@@ -0,0 +1,141 @@
...
+        file = File.get(self.client, _id=file_id)
+        update_file_status(self.client, file, "Indexing")
+
+        for block in file.blocks or []:
+            self._index_block(block, metadata=metadata, index_handle=index_handle)
+
+        update_file_status(self.client, file, "Indexed")
+        return True
+
+    @post("/search_index")
```
dkolas, line 134: Is this a route, or just a utility method?

src/steamship/invocable/mixins/indexer_mixin.py
```diff
@@ -0,0 +1,141 @@
...
+    def search_index(
+        self, query: str, index_handle: Optional[str] = None, k: int = 5
+    ) -> SearchResults:
+        index = self._get_index(index_handle)
+        task = index.search(query, k)
+        task.wait()
+        return task.output
```
dkolas, line 141: `task.wait()` now returns task.output, so this can just be `return task.wait()`

src/steamship/invocable/mixins/indexer_pipeline_mixin.py
```diff
@@ -0,0 +1,89 @@
...
+        self.importer_mixin = FileImporterMixin(client)
+        self.invocable.add_mixin(self.importer_mixin)
+
+        self.blockifier_mixin = BlockifierMixin(client)
+        self.invocable.add_mixin(self.blockifier_mixin)
+
+        self.indexer_mixin = IndexerMixin(client)
+        self.invocable.add_mixin(self.indexer_mixin)
+
+    @post("/set_file_status")
```
dkolas, line 41: Making this a route definitely feels like an encapsulation leak

dkolas (approved):
Some comments below, mostly re: design.  Given the timeframe feel free to pick and choose what to address and what to punt for future iterations.

src/steamship/invocable/mixins/indexer_pipeline_mixin.py
eob: Yeah I agree. I think there are a set of things this sprint uncovered that feel like they want to be a broader Notion design convo.

Re: the encapsulation leak side of things:

The unit of "async scheduleable task" is "public (but authed) endpoint" for us. But some async tasks are very much implementation details, like this one. I haven't yet been able to find a clean way to deal with that.

src/steamship/agents/examples/document_qa_agent.py
```diff
@@ -0,0 +1,97 @@
...
+from steamship.agents.schema import AgentContext
+from steamship.agents.schema.context import Metadata
+from steamship.agents.service.agent_service import AgentService
+from steamship.agents.tools.question_answering import VectorSearchQATool
+from steamship.invocable import post
+from steamship.invocable.mixins.indexer_pipeline_mixin import IndexerPipelineMixin
+from steamship.utils.repl import AgentREPL
+
+
+class DocumentQAService(AgentService):
```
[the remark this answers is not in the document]
eob: Sounds good -- will do (and can adopt that style or other examples)

src/steamship/agents/examples/document_qa_agent.py
```diff
@@ -0,0 +1,97 @@
...
+        super().__init__(**kwargs)
+
+        # This Mixin provides HTTP endpoints that coordinate the learning of documents.
+        #
+        # It adds the `/learn_url` endpoint which will:
+        #    1) Download the provided URL (PDF, YouTube URL, etc)
+        #    2) Convert that URL into text
+        #    3) Store the text in a vector index
+        #
+        # That vector index is then available to the question answering tool, below.
```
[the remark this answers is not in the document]
eob: I suspect I may not have plumbed this 100% of the way through; I'll go back and make sure that it's plumbed through end-to-end. 

But in general, the intention was that `index_name` be an optional arg on all the exposed HTTP endpoints, and also in the QA tool constructor, causing them to select a particular name if one was provided.

src/steamship/agents/examples/document_qa_agent.py
```diff
@@ -0,0 +1,97 @@
...
+        def sync_emit(blocks: List[Block], meta: Metadata):
+            nonlocal output
+            block_text = "\n".join(
+                [b.text if b.is_text() else f"({b.mime_type}: {b.id})" for b in blocks]
+            )
+            output += block_text
+
+        context.emit_funcs.append(sync_emit)
+        self.run_agent(self._agent, context)
+        return output
```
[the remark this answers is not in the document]
eob: Yeah I think this might be a good idea; would definitely clean things up in example code -- and still be overridable from the AgentService one is writing, which is one of the properties that @douglas-reid was trying to build in with it (e.g. if you want your REPL harness to include a particular temperature of LLM, or a particular choice of LLM)

src/steamship/invocable/mixins/blockifier_mixin.py
eob: I got the read folks were a bit hesitant to have a `auto-blockify` function in the SDK proper since it has specific knowledge of plugins we've deployed.. whereas here it feels somehow more OK to have specific knowledge of plugins we've deployed.

Also: it feels easier to put things like the update-file-status here than the SDK, since they're really UI niceties that aren't actually needed (until we decide to officiate them) at the sdk level.

I'm agnostic/unsure

src/steamship/invocable/mixins/indexer_mixin.py
eob: ooh! I didn't know that!!

src/steamship/utils/file_tags.py
```diff
@@ -0,0 +1,14 @@
...
+    file = file.refresh()
+    status_tags = [tag for tag in file.tags if tag.kind == "status"]
+    for status_tag in status_tags:
+        try:
+            status_tag.client = client
+            status_tag.delete()
+        except SteamshipError:
+            pass
+
+    Tag.create(client, file_id=file.id, kind="status", name=status)
```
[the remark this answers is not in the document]
eob: Yeah I agree. This code just ports what `Ask My *` does.. but it leaves me wondering if we want some sort of operation atop tag that replaces if exist, and adds otherwise.

That still doesn't solve the ordering problem you raise here, but it would be a step better than we've got now

src/steamship/agents/examples/document_qa_agent.py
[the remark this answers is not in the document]
eob: Following up: just went through and looked and it appears to be plumbed all the way through.

src/steamship/invocable/mixins/indexer_mixin.py
eob: Route

PR #404, opened 2023-05-31
feat(agents): add basic memory for usage

douglas-reid:
WIP

Sketch of how we could add memory to our agents without too much additional modification. There are probably other ways to do this, etc., but this gives us something to respond to and decide what works (or doesn't).

### DIFF
--- a/src/steamship/agents/examples/my_assistant.py
+++ b/src/steamship/agents/examples/my_assistant.py
@@ -1,11 +1,12 @@
 import uuid
-from typing import List
+from typing import List, Optional
 
 from steamship import Block
 from steamship.agents.llms.openai import OpenAI
 from steamship.agents.react import ReACTAgent
 from steamship.agents.schema import AgentContext
 from steamship.agents.schema.context import Metadata
+from steamship.agents.schema.conversation_memory import MessageWindowMemory
 from steamship.agents.service.agent_service import AgentService
 from steamship.agents.tools.image_generation import DalleTool
 from steamship.agents.tools.search import SearchTool
@@ -26,18 +27,20 @@ class MyAssistant(AgentService):
                 SearchTool(),
                 DalleTool(),
             ],
-            llm=OpenAI(self.client),
+            llm=OpenAI(self.client, temperature=0),
+            conversation_memory=MessageWindowMemory(k=2),
         )
 
     @post("prompt")
-    def prompt(self, prompt: str) -> str:
+    def prompt(self, prompt: str, context_id: Optional[uuid.UUID] = None) -> str:
         """Run an agent with the provided text as the input."""
 
         # AgentContexts serve to allow the AgentService to run agents
         # with appropriate information about the desired tasking.
-        # Here, we create a new context on each prompt, and append the
-        # prompt to the message history stored in the context.
-        context_id = uuid.uuid4()
+        # Here, we use the passed in context (or a new context) for the prompt,
+        # and append the prompt to the message history stored in the context.
+        if not context_id:
+            context_id = uuid.uuid4()
         context = AgentContext.get_or_create(self.client, {"id": f"{context_id}"})
         context.chat_history.append_user_message(prompt)
 
@@ -59,6 +62,8 @@ class MyAssistant(AgentService):
 
         context.emit_funcs.append(sync_emit)
         self.run_agent(self._agent, context)
+        # TODO: is this right?
+        context.chat_history.append_assistant_message(output)
         return output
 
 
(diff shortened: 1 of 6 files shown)


src/steamship/agents/react/react.py
```diff
@@ -71,8 +76,13 @@ def next_action(self, context: AgentContext) -> Action:
             tool_index=tool_index,
             tool_names=tool_names,
             scratchpad=scratchpad,
+            chat_history=self.conversation_memory.messages_as_string(
```
EniasCailliau, line 79: Could this be part of AgentContext?

src/steamship/agents/examples/my_assistant.py
```diff
@@ -26,18 +27,20 @@ def __init__(self, **kwargs):
...
                 DalleTool(),
             ],
-            llm=OpenAI(self.client),
+            llm=OpenAI(self.client, temperature=0),
+            conversation_memory=MessageWindowMemory(k=2),
         )
 
     @post("prompt")
-    def prompt(self, prompt: str) -> str:
+    def prompt(self, prompt: str, context_id: Optional[uuid.UUID] = None) -> str:
```
eob, line 35: I haven't yet read ahead, but in general I continue to really like your way of thinking about context

src/steamship/agents/examples/my_assistant.py
```diff
@@ -59,11 +62,13 @@ def sync_emit(blocks: List[Block], meta: Metadata):
 
         context.emit_funcs.append(sync_emit)
         self.run_agent(self._agent, context)
+        # TODO: is this right?
+        context.chat_history.append_assistant_message(output)
```
eob, line 66: This feels like the right direction to me. High level reactions to see if you're thinking the same way:

- The fact that `/prompt` is in user code, rather than a base class, is sort of an acknowledgement that there's some optionality here we want to keep the user in control of...
- Along with that is: what agent gets run? in what context? how are messages returned? do we remember that we said it?

It feels to me like appending to the chat history is something that we might decide is "pushed into the agent" later.. but for now it doesn't strike me that there's a downside of putting it here. 

It sort of educates the end-developer about what's actually happening, too, which can be good

src/steamship/agents/memory/chathistory.py
```diff
@@ -97,7 +97,7 @@ def append_system_message(
             text=text, tags=tags, content=content, url=url, mime_type=mime_type
         )
 
-    def append_agent_message(
+    def append_assistant_message(
```
eob, line 100: We might want to start curating (in the docs) some common terminology, even if we know it'll change.

E.g. "we tend to use `assistant` to mean....", "we tend to use `xyz` to mean ..."

src/steamship/agents/react/react.py
```diff
@@ -71,8 +76,13 @@ def next_action(self, context: AgentContext) -> Action:
...
+                chat_history=context.chat_history
+            ),
         )
 
+        print(f"prompt: {prompt}")
```
eob, line 84: no biggie but flagging so you see it

src/steamship/agents/schema/agent.py
```diff
@@ -20,6 +21,9 @@ class Agent(BaseModel, ABC):
     tools: List[Tool]
     """Tools that can be used by the Agent in selecting the next Action."""
 
+    conversation_memory: ConversationMemory = NoMemory()
+    """Conversation memory to use when running agent."""
```
eob, line 25: btw I had no idea this was a way to document variables in python and ever since seeing you do it have adopted it -- thanks!!

src/steamship/agents/schema/conversation_memory.py
```diff
@@ -0,0 +1,104 @@
...
+        self, chat_history: ChatHistory, user_prefix: str = "User", assistant_prefix: str = "AI"
+    ) -> str:
+        messages = self.messages(chat_history)
+        print(f"messages: {messages}")
+        as_strings = []
+        for block in messages:
+            role = block.chat_role
+            if role == RoleTag.USER:
+                as_strings.append(f"{user_prefix}: {block.text}")
+            elif role == RoleTag.ASSISTANT:
```
eob, line 27: I get super easily confused at the difference between "assistant" and "system" and "agent" so if others are like me, a quick inline reminder might be helpful

src/steamship/agents/schema/conversation_memory.py
```diff
@@ -0,0 +1,104 @@
...
+                as_strings.append(f"{assistant_prefix}: {block.text}")
+            elif role == RoleTag.SYSTEM:
+                as_strings.append(f"System: {block.text}")
+            elif role == RoleTag.AGENT:
+                as_strings.append(f"Agent: {block.text}")
+        return "\n".join(as_strings)
+
+
+class NoMemory(ConversationMemory):
```
eob, line 36: +1 to this idea - it's great.

If I'm reading this right: the way to not have memory is to actively chose the memory type NoMemory?

eob (commented):
Love this. Added some comments here and there that I hope are useful!

eob (commented):
Love this. Added some comments here and there that I hope are useful!

douglas-reid:
Closing in favor of style from Riff 2.

PR #407, opened 2023-06-02
feat: File Importer Mixin

eob:
This adds a Mixin which provides official support for the notion of importing a file.

- An HTTP method is provided that just returns the (eventually content-laden) file object
- A python object method is provide which also returns the task (if there is one) upon which to wait for the content

This is a building block to making Ask My functionality something that can just be mixed-in to an AgentService.

Note:

The end-game I'm going for is a `DocumentQA` mixin which takes the service as an init arg and then applies these component-level mixins, along with a coordinating endpoint that unites them all.


dkolas (approved):
Looks good to me - I think this mixin style is going to be very powerful

dkolas (approved):
Looks good to me - I think this mixin style is going to be very powerful

PR #409, opened 2023-06-02
Emit Bug Fix [Disconcerting - explanation required]

dkolas:
This is a fix for a bug where the first message to a web bot works, but subsequent calls do not.

The issue appears to be that SOMEHOW, the emit_func from the previous run is still present in memory.  This does not make any sense to me, nor does the "existing emit functions" logging support that.

AND YET - The other logging, of which emit function is being used, supports the existing-functions theory.  Before the change to overwrite the list, if you sent a message to the Web client, then a message to Telegram, the telegram call would attempt to use the web client's emit.  Also, changing the code to always overwrite the emit function list fixes the issue.

### DIFF
--- a/src/steamship/agents/service/agent_service.py
+++ b/src/steamship/agents/service/agent_service.py
@@ -77,4 +77,5 @@ class AgentService(PackageService):
             f"Completed agent run. Result: {len(action.output or [])} blocks. {output_text_length} total text length. Emitting on {len(context.emit_funcs)} functions."
         )
         for func in context.emit_funcs:
+            logging.info(f"Emitting via function: {func.__name__}")
             func(action.output, context.metadata)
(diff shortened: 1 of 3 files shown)


src/steamship/agents/service/agent_service.py
```diff
@@ -70,5 +70,12 @@ def run_agent(self, agent: Agent, context: AgentContext):
             )
 
         context.completed_steps.append(action)
+        output_text_length = 0
+        if action.output is not None:
+            output_text_length = sum([len(block.text or "") for block in action.output])
+        logging.info(
+            f"Completed agent run. Result: {len(action.output or [])} blocks. {output_text_length} total text length. Emitting on {len(context.emit_funcs)} functions."
```
douglas-reid, line 77: In the future, we should probably make these debugs?

douglas-reid (dismissed):
Looks good. I'd also suggest in the Agent context.get_or_create() that you clear emitFn to be 100% sure.

PR #411, opened 2023-06-05
Transports as Mixins

dkolas:
This PR refactors Transports to be Package Mixins.  This allows them to be additive aspects to a an Agent Package without requiring subclassing or conflicting with one another.

- Web Widget and Telegram transports promoted from Experimental to agents package.
- Functionality from TelegramAgent and SteamshipWeb agent merged into the transports to make them mixins.
- Old versions of Bots, Agents, and Transports removed to reduce surface area.
- Introduce mock Telegram service and first actual SDK tests for web widget to hopefully reduce regression issues.

Other intentionally removed functionality:

- Transport as a mixin superclass can't provide any routes, because they would always conflict as soon as you had multiple Transports.  This removes the prototype `deinit` functionality.  This could / should be readded with Engine support so that there is a guarantee it is called.


src/steamship/agents/mixins/transports/transport.py
```diff
@@ -0,0 +1,87 @@
...
+
+from steamship import Block, Steamship
+from steamship.agents.schema import AgentContext, Metadata
+from steamship.agents.tools.audio_transcription.assembly_speech_to_text_tool import (
+    AssemblySpeechToTextTool,
+)
+from steamship.invocable.package_mixin import PackageMixin
+
+
+class Transport(PackageMixin, ABC):
```
eob, line 14: `TransportMixin`? Unsure if I have a strong opinion

src/steamship/data/block.py
```diff
@@ -207,7 +207,12 @@ def is_image(self):
 
     def is_audio(self):
         """Return whether this is an audio Block."""
-        return self.mime_type in [MimeTypes.MP3, MimeTypes.MP4_AUDIO, MimeTypes.WEBM_AUDIO]
+        return self.mime_type in [
```
eob, line 210: I think we might want to sneak `OGG` in since Telegram apparently uses that

eob (approved):
LGTM! 

Clicking approve but I think we might want to slightly alter some of the endpoints to carve out a namespace just for telegram. Up to you since you're closer to the code. 

REALLY excited that we've got test coverage of this now 🚀🚀🚀🚀🚀🚀

src/steamship/agents/mixins/transports/telegram.py
```diff
@@ -46,7 +59,12 @@ def _instance_init(self, *args, **kwargs):
                 f"Could not set webhook for bot. Webhook URL was {webhook_url}. Telegram response message: {response.text}"
             )
 
-    def _instance_deinit(self, *args, **kwargs):
+    @post("webhook_info")
+    def webhook_info(self) -> dict:
+        return requests.get(self.api_root + "/getWebhookInfo").json()
+
+    @post("disconnect_webhook")
```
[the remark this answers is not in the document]
dkolas: Yeah, I meant to do this and forgot.  Done now.

src/steamship/agents/mixins/transports/telegram.py
```diff
@@ -153,3 +165,42 @@ def _parse_inbound(self, payload: dict, context: Optional[dict] = None) -> Optio
...
+
+    def build_emit_func(self, chat_id: str) -> EmitFunc:
+        def new_emit_func(blocks: List[Block], metadata: Metadata):
+            for block in blocks:
+                block.set_chat_id(chat_id)
+            return self.send(blocks, metadata)
+
+        return new_emit_func
+
+    @post("respond", public=True)
```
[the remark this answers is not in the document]
dkolas: 👍

src/steamship/data/block.py
dkolas: Added OGG_AUDIO and OGG_VIDEO.

douglas-reid: do we want some sort of catch-all (like `starts_with("audio/")` ?

PR #418, opened 2023-06-08
Fix python insanity

dkolas:
This explains the previously unexplainable #409 .

```
>>> class A:
...    a_list=[]
...    def __init__(self):
...       self.a_list.append("a")
...
>>> a = A()
>>> a.a_list
['a']
>>> b = A()
>>> b.a_list
['a', 'a']
The gist being that a_list is a static class var, at least until you would overwrite it with an instance-level version:
>>> a.a_list = ['b']
>>> a.a_list
['b']
>>> b.a_list
['a', 'a']
```

### DIFF
--- a/src/steamship/agents/schema/context.py
+++ b/src/steamship/agents/schema/context.py
@@ -19,7 +19,7 @@ class AgentContext:
     def id(self) -> str:
         return self.chat_history.file.id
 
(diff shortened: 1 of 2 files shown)


eob (approved):
LGTM - nice catch

PR #425, opened 2023-06-09
Vector memory integration with ChatHistory

dkolas:
This PR provides a quick, rough sketch of adding vector memory to `ChatHistory`.

This approach:
- Makes indexing of all blocks in the conversation automatic
- provides a `search` method on `ChatHistory` backed by EmbeddingIndexPluginInstance
- Provides a simple interface for text splitting [and one very simple implementation]

Open Questions:
- Is it better for the vector memory to be totally separate from the linear history?
- Should tools access this memory rather than building their own?


src/steamship/agents/memory/chathistory.py
```diff
@@ -1,30 +1,37 @@
...
 class ChatHistory:
-    """A ChatHistory is a wrapper of a File ideal for ongoing interactions between a user and a virtual assistant."""
+    """A ChatHistory is a wrapper of a File ideal for ongoing interactions between a user and a virtual assistant.
+    It also includes vector-backed storage for similarity-based retrieval."""
 
     file: File
+    embedding_index: EmbeddingIndexPluginInstance
 
-    def __init__(self, file: File):
+    def __init__(self, file: File, embedding_index: EmbeddingIndexPluginInstance):
```
eob, line 23: (unsure of confidence-level):

We could consider permitting this to be `None`, which would indicate that vector-memory was not desired to play a role.

src/steamship/agents/memory/chathistory.py
```diff
@@ -1,30 +1,37 @@
...
         self.file = file
+        self.embedding_index = embedding_index
 
     @staticmethod
     def _get_existing_file(client: Client, context_keys: Dict[str, str]) -> Optional[File]:
         """Find an existing File object whose memory Tag matches the passed memory keys"""
-        file_query = f'kind "{TagKind.CHAT}" and name "{ChatTag.CONTEXT_KEYS}" and ' + " and ".join(
-            [f'value("{key}") = "{value}"' for key, value in context_keys.items()]
+        file_query = (
+            f'kind "{TagKind.CHAT}" and name "{ChatTag.CONTEXT_KEYS}"'
```
eob, line 32: The query system really comes back to pay us dividends in so many small and awesome ways

src/steamship/agents/memory/chathistory.py
```diff
@@ -36,9 +43,35 @@ def _get_existing_file(client: Client, context_keys: Dict[str, str]) -> Optional
...
 
+    @staticmethod
+    def _get_index_handle_from_file(file: File) -> str:
+        for tag in file.tags:
+            if tag.kind == TagKind.CHAT and tag.name == ChatTag.INDEX_HANDLE:
+                return tag.value[TagValueKey.STRING_VALUE]
+        raise SteamshipError(f"Could not find index handle on file with id {file.id}")
+
+    @staticmethod
+    def _get_embedding_index(client: Steamship, index_handle: str) -> EmbeddingIndexPluginInstance:
```
eob, line 54: Aside: variations of this method are in a few places in the codebase (I've added variations of this to a few spots in the Tools & Mixins). At some point we may want to extract it into a util method.. or maybe even a static method on EmbeddingIndexPluginIndex itself.. like `get_default` or something

src/steamship/agents/memory/chathistory.py
```diff
@@ -47,20 +80,54 @@ def get_or_create(
...
             )
-        return ChatHistory(file)
+        else:
+            index_handle = ChatHistory._get_index_handle_from_file(file)
+
+        embedding_index = ChatHistory._get_embedding_index(client, index_handle)
+
+        return ChatHistory(file, embedding_index)
+
+    def _chunk_text(self, text: str) -> List[Tag]:
```
eob, line 107: Thinking out loud:

You know... the VectorSearchLearnerTool [1] actually does similar things... Would we want to leverage that as a reusable code asset in addition to an Agent-usable Tool? That might be an interesting way to consolidate the implementation. (At the cost of potentially trying to force two things to be the same when they might want to be 10% different)



[1] https://github.com/steamship-core/python-client/blob/main/src/steamship/agents/tools/question_answering/vector_search_learner_tool.py

src/steamship/invocable/mixins/indexer_mixin.py
```diff
@@ -10,7 +10,7 @@
 DEFAULT_EMBEDDING_INDEX_CONFIG = {
     "embedder": {
         "plugin_handle": "openai-embedder",
-        "plugin_instance-handle": "text-embedding-ada-002",
+        "plugin_instance_handle": "text-embedding-ada-002",
```
eob, line 13: OMG. Nice catch <breaks into cold sweat..>

eob (commented):
Unsure since WIP if you were looking for comments or a LGTM, but tried to add some comments that might be useful. It's really exciting that this looks like it's so straightforward to incorporate

douglas-reid (commented):
Very cool. 

I was _almost_ expecting something different -- like a `VectorChatHistory` which provided overrides of the `append` methods, etc., but this seems workable too.

One thing I was thinking about is extension points (how easy/hard is it to replace Steamship with Pinecone, etc.). But I think leaning into "Steamship-native" is something I'm growing to appreciate (in the same way messages are `Block`s).

One concern is cost-management: we probably need a way to opt-out of embedding every message for users.

A smaller, secondary concern -- which is outside the scope of this PR -- is transferability. As a user, I'd hate to have to re-embed for new instances, etc.

src/steamship/agents/memory/chathistory.py
```diff
@@ -47,20 +80,54 @@ def get_or_create(
...
+        if text is not None and text != "":
```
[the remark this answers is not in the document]
dkolas: 💯 .  I just wanted to throw the dumb version in first so we could react to the rest.  I'll add an interface for this today.

src/steamship/agents/memory/chathistory.py
```diff
@@ -1,30 +1,37 @@
...
 from steamship.data.block import Block
+from steamship.data.plugin.index_plugin_instance import EmbeddingIndexPluginInstance, SearchResults
 from steamship.data.tags import Tag
 from steamship.data.tags.tag_constants import ChatTag, DocTag, RoleTag, TagValueKey
 
 
 class ChatHistory:
-    """A ChatHistory is a wrapper of a File ideal for ongoing interactions between a user and a virtual assistant."""
+    """A ChatHistory is a wrapper of a File ideal for ongoing interactions between a user and a virtual assistant.
+    It also includes vector-backed storage for similarity-based retrieval."""
```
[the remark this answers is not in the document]
dkolas: That's the gist.  I really liked the idea of it being totally automatic.  Agree with other note though that there needs to be a way to turn it off for cost purposes (or maybe it should be off by default since it costs $).

src/steamship/agents/memory/chathistory.py
dkolas: 👍

src/steamship/agents/schema/chathistory.py
```diff
@@ -1,30 +1,48 @@
...
     file: File
+    embedding_index: EmbeddingIndexPluginInstance
+    text_splitter: TextSplitter
 
-    def __init__(self, file: File):
+    def __init__(
+        self,
+        file: File,
+        embedding_index: Optional[EmbeddingIndexPluginInstance],
+        text_splitter: TextSplitter = None,
```
douglas-reid, line 29: @dkolas `text_splitter` stuff looks good to me.

i'll be curious to see what we find out about chat history messages (typical sizes, etc.). _Might_ be worth `logging.info` with an `extra` field related to number of chunks for new messages or something that would allow us to have some insights (as a platform) and to provide those insights to our customers. Unless we think DB tag queries should suffice moving forward as an alternative (for platform usages)?

PR #428, opened 2023-06-12
cli: Add user interface to development-mode agents

eob:
This adds a user interface to development-mode agents.

```
ship serve --ui
```

Will generate a URL that will bring up a web browser that connects to the local agent.

Requires a corresponding 

https://github.com/steamship-core/python-client/assets/63262/64c9fb1b-ba4f-421f-84bf-9b7307e7e3f2

### DIFF
--- a/src/steamship/cli/cli.py
+++ b/src/steamship/cli/cli.py
@@ -10,7 +10,7 @@ import click
 
 import steamship
 from steamship import Steamship, SteamshipError
-from steamship.base.configuration import Configuration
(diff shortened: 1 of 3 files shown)


dkolas (approved):
This is neat, and code looks good.

One small (10%) eyebrow raise - I have a tiny worry about where we will draw the line on what type of developers we will support, and whether it will be before we get to a full no-code system.  Given that we can't be everything to everyone, we may need to write down what our core developer persona is.

PR #432, opened 2023-06-14
Add Vector memory to Agents

dkolas:
This PR creates a parallel structure for formatting search results for prompts, and inserts them into the React agent.

There is some back-and-forth between the SDK and the engine when creating chunk tags that we may wish to optimize at some point.

I did not implement it as a `MessageSelector`, but rather as a separate thing, because it did not fit the interface.  I'm not sure if that decision was correct.


src/steamship/agents/react/react.py
```diff
@@ -79,6 +82,9 @@ def next_action(self, context: AgentContext) -> Action:
             chat_history=self.messages_to_prompt_history(
                 messages=context.chat_history.select_messages(self.message_selector)
             ),
+            search_history=self.search_to_search_history(
```
EniasCailliau, line 85: Could we use the same interface as `messages_to_prompt_history`?

src/steamship/agents/react/react.py
```diff
@@ -54,6 +54,9 @@ class ReACTAgent(LLMAgent):
 Previous conversation history:
 {chat_history}
 
+Other relevant previous conversation:
+{search_history}
```
EniasCailliau, line 58:
```suggestion
{relevant_history}
```

EniasCailliau: Not sure if this is better. But search implies limits, not sure we want that.

src/steamship/agents/schema/agent.py
```diff
@@ -57,3 +59,11 @@ def messages_to_prompt_history(self, messages: List[Block]) -> str:
             elif role == RoleTag.AGENT:
                 as_strings.append(f"Agent: {block.text}")
         return "\n".join(as_strings)
+
+    @staticmethod
+    def search_to_search_history(search_results: SearchResults) -> str:
+        blocks = []
```
EniasCailliau, line 65: Could SearchResults override __iter__ to encapsulate this behavior? 

If yes, we would not need the search_to_search_history wrapper. Instead we could use messages_to_prompt_history directly.

src/steamship/agents/schema/chathistory.py
```diff
@@ -140,8 +140,22 @@ def append_message_with_role(
             chunk_tags = self.text_splitter.chunk_text_to_tags(
                 text, kind=TagKind.CHAT, name=ChatTag.CHUNK
             )
-            block.tags.extend(chunk_tags)
-            self.embedding_index.insert(chunk_tags)
+            saved_tags = []
```
EniasCailliau, line 143: Can the chunk_text_to_tags take care of this? 

Another question: Do we need a text_splitter here? One could assume chat messages are small enough to fit in 1 embedding?

EniasCailliau (changes requested):
I think I see a way for us to only use `messages_to_prompt_history`. 

Worth it for me because it limits the number of interfaces I would be using to set up memory & context for my bot.

src/steamship/agents/react/react.py
dkolas: 👍

src/steamship/agents/schema/chathistory.py
dkolas: Good suggestion, cleaned that up I think.

I added text_splitter because I couldn't answer the question "should we assume chat messages are small enough to fit in an embedding".  To me having them chunked to preserve some sensible embedding performance seemed like a safer bet.

src/steamship/agents/schema/agent.py
dkolas: I did this slightly differently, by moving the content of the method to SearchResults, but I think still achieving the desired effect.  I didn't make it the iter because iter on SearchResults would mean SearchResult, not Block, to me.

src/steamship/agents/react/react.py
dkolas: Done

src/steamship/agents/react/react.py
```diff
@@ -54,6 +54,9 @@ class ReACTAgent(LLMAgent):
 Previous conversation history:
 {chat_history}
 
+Other relevant previous conversation:
```
douglas-reid, line 57: do we expect that we will need both at the same time?  I can't convince myself that we need both `Previous conversation history:` and `Other relevant previous conversation:` to be optional somehow. When empty, they don't contribute too many tokens, I suppose.

src/steamship/agents/react/react.py
```diff
@@ -79,6 +82,11 @@ def next_action(self, context: AgentContext) -> Action:
             chat_history=self.messages_to_prompt_history(
                 messages=context.chat_history.select_messages(self.message_selector)
             ),
+            relevant_history=self.messages_to_prompt_history(
+                context.chat_history.search(context.chat_history.last_user_message.text, k=3)
+                .wait()
```
douglas-reid, line 87: what happens if `wait()` fails?  should we try this in a recoverable fashion?

src/steamship/agents/schema/text_splitters.py
```diff
@@ -10,22 +10,22 @@ def split_text(self, text: str) -> List[str]:
         """Split the incoming text into strings"""
         raise NotImplementedError()
 
-    def chunk_text_to_tags(self, text: str, kind: str, name: str = None) -> List[Tag]:
+    def chunk_text_to_tags(self, block: Block, kind: str, name: str = None) -> List[Tag]:
         """Split the incoming text into strings, and then wrap those strings in Tags"""
-        if text is not None and text != "":
-            text_splits = self.split_text(text)
+        if block.text is not None and block.text != "":
```
douglas-reid, line 15: do you need to check `is_text()` at all?

src/steamship/data/plugin/index_plugin_instance.py
```diff
@@ -73,10 +76,19 @@ class SearchResults(CamelModel):
...
+        items = [
+            SearchResult.from_query_result(qr, client=client) for qr in query_results.items or []
+        ]
         return SearchResults(items=items)
 
+    def to_ranked_blocks(self) -> List[Block]:
+        blocks = []
+        for result in self.items:
+            tag = result.tag
+            blocks.append(Block.get(tag.client, tag.block_id))
```
douglas-reid, line 89: if `get()` raises an exception, should the entire operation fail?  do we want a graceful handling?

tests/steamship_tests/agents/test_message_formatting.py
```diff
@@ -0,0 +1,37 @@
...
+    )
+    chat_history = ChatHistory.get_or_create(client, context_keys)
+
+    block = chat_history.append_user_message(text="Hi. My birthday is today")
+    chunk_tag = [tag for tag in block.tags if tag.name == ChatTag.CHUNK][0]
+    assert chunk_tag.start_idx == 0
+    assert chunk_tag.end_idx == 24
+
+    chat_history.append_user_message("I like bananas")
+    formatted_result = agent.search_to_search_history(chat_history.search("cake", k=1).wait())
```
douglas-reid, line 30: does `search_to_search_history` still exist?

dkolas: It does not.  That was a mistake, now corrected.

src/steamship/agents/react/react.py
dkolas: I'm honestly not sure why `search` is a task to begin with.  Searching is not a class of operation I expect to be asynchronous.

src/steamship/agents/schema/text_splitters.py
dkolas: 👍

src/steamship/agents/react/react.py
dkolas: I was imagining a long conversation with an agent.  The Previous conversation history might include the most recent N messages, and the 'other relevant' bit might include a few snippets that most closely matched the last input.  

I'm very open to any other / better suggestions on how to incorporate this into the prompt.

PR #435, opened 2023-06-16
feat(functions): add support for mapping tools to openai functions

douglas-reid:
In order to build `Agent`s that can exploit OpenAI's `function_call` interface, we will likely want a way to directly map Steamship Tools to OpenAI function objects.

This PR proposes a way to accomplish this task. It is a parallel PR to https://github.com/steamship-plugins/gpt4/pull/8 (which validates ~equivalent input will work with OpenAI prompting).

With the default scheme, any tool that accepts as input either:
- a single text string
- a single non-text block
- a text string and a non-text block

should be usable in OpenAI `function_call` based prompting.

More work will be necessary to support multiple inputs of the same type (text or uuid). Submitting parameters of type `array` was causing errors returned by OpenAI, so this may be an unimplemented feature on their end that we need to work around.

No attempt here was made to add built-in Tag support for tags need for `name` tags and `function` messages. That should be done in a subsequent PR.


src/steamship/agents/schema/tool.py
```diff
@@ -47,3 +53,25 @@ def run(self, tool_input: List[Block], context: AgentContext) -> Union[List[Bloc
     def post_process(self, async_task: Task, context: AgentContext) -> List[Block]:
         """Transforms Task output into a List[Block]."""
         pass
+
+    def as_openai_function(self) -> OpenAIFunction:
```
dkolas, line 57: Minor nit: Does this belong in Tool, or in some OpenAI specific subclass of tool?

dkolas (approved):
Looks good! One minor question.

src/steamship/agents/schema/tool.py
douglas-reid: My initial reaction is that all Tools should be able to be translated. This is really just a schema thing. It _feels_ like having to go back and decide which tools to make OpenAI tools introduces cognitive (and other) overhead.

I could see having this be on the OpenAIFunction class, maybe. Like, `OpenAIFunction.from_tool(Tool)` maybe.

The nice thing about having it be on Tool however is that individual Tools could decide to override the generation in a tailored fashion.

The other bit I've been thinking about is whether or not this should return a list of OpenAIFunctions, so that a tool could potentially represent a range of functions. For the moment, I figure that can be handled with an enum parameter though.

EniasCailliau: +1 to make this available for all tools

EniasCailliau (approved):
LGTM

src/steamship/agents/schema/tool.py
dkolas: Fine by me :)

PR #441, opened 2023-06-27
When using ngrok, don't append the local port to `invocable_url`

eob:
The `invocable_url` on the context represents the externally-visible URL of the instance.

When running on localhost, this should include the port that the local server is running on.

But when running with NGROK involved, the local port is within the ngrok pipe, and the ngrok url should be used without modification.


douglas-reid (approved):

eob:
Failures were timeout-related & local testing of the feature works, so merging.

PR #442, opened 2023-06-27
Add ability to delete packages plugins and their versions

dkolas:
Depends on https://github.com/nludb/nludb/pull/579

Provide delete methods for Package, PackageVersion, Plugin, and PluginVersion.

These will fail server-side if the object is public and another user has created an instance of it.


tests/steamship_tests/utils/deployables.py
```diff
@@ -150,6 +150,8 @@ def _check_user(client, instance):
 
 def _delete_deployable(instance, version, deployable):
     instance.delete()
+    version.delete()
```
EniasCailliau, line 153: From the description, I understand that `deployable.delete()` cascades to versions. 

Would consider adding a test case that only calls `deployable.delete()` without a call to `version.delete()`

EniasCailliau (approved):
Would add a test case for `package.delte()` without `version.delete()`

PR #443, opened 2023-06-28
docs: add info on calling plugins over HTTP

douglas-reid:
User emailed requesting HTTP instructions for invoking plugins. I've taken that info and added to our external docs based on the similar information for packages.


dkolas (commented):
Is this a thing we actually want to support?  I would vote no.

Currently the engine mediates calls to and from plugins.  For `generators`, this means extracting the appropriate content of existing files, resolving queries, etc.  This means that the API interface to a plugin is aligned with internal expectations of both the engine and the plugin helper shim.  We may wish to change these in the future.  

Further, using us in this way effectively makes us an AI model reseller. This seems to go against who we think our target audience is.

At the very least, I think we should mark this as experimental / unsupported / beta / likely to break or something.

douglas-reid:
> Is this a thing we actually want to support? I would vote no.
> 
> Currently the engine mediates calls to and from plugins. For `generators`, this means extracting the appropriate content of existing files, resolving queries, etc. This means that the API interface to a plugin is aligned with internal expectations of both the engine and the plugin helper shim. We may wish to change these in the future.
> 
> At the very least, I think we should mark this as experimental / unsupported / beta / likely to break or something.

I was fairly torn on this. Having people use our SDK is definitely the preferred path. However, we've had multiple users recently asking to use HTTP directly (and even attempting to figure it out on their own). That, to me, indicates that desire is out-pacing our intentions for plugins. I'm fine adding a bunch of disclaimers and other language (or even an answer that reads "write a simple package that wraps the plugin and use that package. here's example code..."), but I think we need to have _something_ that addresses the need.

Thoughts?

dkolas:
I apologize, I definitely misunderstood what was being put on offer here.  I'm totally fine exposing the engine API for plugins in this way.

douglas-reid:
PTAL. Added some language to attempt to indicate SDK is preferred, and clarify the API surface being invoked.

Issue #440, opened 2023-06-27
bug：chinese encoding error


wunglee:
from steamship import Steamship

instance = Steamship.use("PACKAGE_HANDLE", "INSTANCE_HANDLE", config={
    "default_name": "王某"  
})

print(instance.invoke("greet")      #######.  chinese encoding error

dkolas:
Thank you for reporting!  I've been able to reproduce locally.  Will work on a fix.

dkolas:
Hi again - just reporting that I have a backend fix in the works.  Will close this issue once it's deployed to production.

PR #449, opened 2023-06-28
Encoding Bug (Validation Test)

dkolas:
This PR adds to a test with a case that validates that encodings work properly in results within the SDK.

It will not pass until the backend fix is deployed.

Validation of fix for #440


EniasCailliau (approved):
🚢

PR #453, opened 2023-06-29
Allow mixin subclasses to override the routes of their parent classes

dkolas:


douglas-reid (commented):
This looks reasonable to me. I'll let @EniasCailliau provide the authoritative review however, as he seems to have experienced the pain this solves directly.

EniasCailliau (approved):

PR #454, opened 2023-06-29
feat(agents): add functions-based agent to SDK

douglas-reid:
This is meant to provide a basic implementation of an agent that works with tools using OpenAI's functions extension as the mechanism for tools selection. This is an initial implementation that will likely need refinement as use cases are explored.

### DIFF
--- a/src/steamship/agents/examples/my_functions_based_assistant.py
+++ b/src/steamship/agents/examples/my_functions_based_assistant.py
@@ -1,0 +1,76 @@
+import uuid
+from typing import List, Optional
+
+from steamship import Block
+from steamship.agents.functional import FunctionsBasedAgent
+from steamship.agents.llms.openai import ChatOpenAI
+from steamship.agents.schema import AgentContext
+from steamship.agents.schema.context import Metadata
+from steamship.agents.schema.message_selectors import MessageWindowMessageSelector
+from steamship.agents.service.agent_service import AgentService
+from steamship.agents.tools.image_generation import DalleTool
+from steamship.agents.tools.search import SearchTool
+from steamship.invocable import post
+from steamship.utils.repl import AgentREPL
+
+
+class MyFunctionsBasedAssistant(AgentService):
+    """MyFunctionsBasedAssistant is an example AgentService that exposes a single test endpoint
+    for trying out Agent-based invocations. It is configured with two simple Tools
+    to provide an overview of the types of tasks it can accomplish (here, search
+    and image generation)."""
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self._agent = FunctionsBasedAgent(
+            tools=[
+                SearchTool(),
+                DalleTool(),
+            ],
+            llm=ChatOpenAI(self.client, temperature=0),
+            conversation_memory=MessageWindowMessageSelector(k=2),
+        )
+
+    @post("prompt")
+    def prompt(self, prompt: str, context_id: Optional[uuid.UUID] = None) -> str:
+        """Run an agent with the provided text as the input."""
+
+        # AgentContexts serve to allow the AgentService to run agents
+        # with appropriate information about the desired tasking.
+        # Here, we use the passed in context (or a new context) for the prompt,
+        # and append the prompt to the message history stored in the context.
+        if not context_id:
+            context_id = uuid.uuid4()
+        context = AgentContext.get_or_create(self.client, {"id": f"{context_id}"})
+        context.chat_history.append_user_message(prompt)
+
+        # AgentServices provide an emit function hook to access the output of running
+        # agents and tools. The emit functions fire at after the supplied agent emits
+        # a "FinishAction".
+        #
+        # Here, we show one way of accessing the output in a synchronous fashion. An
+        # alternative way would be to access the final Action in the `context.completed_steps`
+        # after the call to `run_agent()`.
+        output = ""
+
+        def sync_emit(blocks: List[Block], meta: Metadata):
+            nonlocal output
+            block_text = "\n".join(
+                [b.text if b.is_text() else f"({b.mime_type}: {b.id})" for b in blocks]
+            )
+            output += block_text
+
+        context.emit_funcs.append(sync_emit)
+        self.run_agent(self._agent, context)
+        # TODO: is this right?
+        context.chat_history.append_assistant_message(output)
+        return output
+
+
+if __name__ == "__main__":
+    # AgentREPL provides a mechanism for local execution of an AgentService method.
+    # This is used for simplified debugging as agents and tools are developed and
+    # added.
+    AgentREPL(MyFunctionsBasedAssistant, "prompt", agent_package_config={}).run(
+        context_id=uuid.uuid4()
+    )
--- a/src/steamship/agents/examples/my_functions_based_assistant.py
+++ b/src/steamship/agents/examples/my_functions_based_assistant.py
@@ -62,7 +62,6 @@ class MyFunctionsBasedAssistant(AgentService):
 
         context.emit_funcs.append(sync_emit)
         self.run_agent(self._agent, context)
-        # TODO: is this right?
         context.chat_history.append_assistant_message(output)
         return output
 
--- a/src/steamship/agents/functional/functions_based.py
+++ b/src/steamship/agents/functional/functions_based.py
@@ -1,18 +1,12 @@
 from typing import List
 
 from steamship import Block
-from steamship.agents.openai.output_parser import FunctionsBasedOutputParser
-from steamship.agents.schema import (
-    Action,
-    AgentContext,
-    ConversationalLLM,
-    ConversationalLLMAgent,
-    Tool,
-)
+from steamship.agents.functional.output_parser import FunctionsBasedOutputParser
+from steamship.agents.schema import Action, AgentContext, ChatAgent, ChatLLM, Tool
 from steamship.data.tags.tag_constants import RoleTag
 
 
-class FunctionsBasedAgent(ConversationalLLMAgent):
+class FunctionsBasedAgent(ChatAgent):
     """Selects actions for AgentService based on OpenAI Function style LLM Prompting."""
 
     PROMPT = """You are a helpful AI assistant. Only use the functions you have been provided with.
@@ -24,13 +18,12 @@ video, or audio as follows: `Block(UUID for the block)`.
 Example response for a request that generated an image:
 Here is the image you requested: Block(288A2CA1-4753-4298-9716-53C1E42B726B)."""
 
-    def __init__(self, tools: List[Tool], llm: ConversationalLLM, **kwargs):
+    def __init__(self, tools: List[Tool], llm: ChatLLM, **kwargs):
         super().__init__(
             output_parser=FunctionsBasedOutputParser(tools=tools), llm=llm, tools=tools, **kwargs
         )
 
     def next_action(self, context: AgentContext) -> Action:
-
         messages = []
 
         # get system messsage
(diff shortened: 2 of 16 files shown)


src/steamship/agents/schema/agent.py
```diff
@@ -57,4 +57,20 @@ def messages_to_prompt_history(messages: List[Block]) -> str:
...
+class ConversationalLLMAgent(Agent):
+    """ConversationalLLMAgents choose next actions for an AgentService based on chat-based interactions with an LLM."""
+
+    llm: ConversationalLLM
+
+    output_parser: OutputParser
+    """Utility responsible for converting LLM output into Actions"""
+
+    @abstractmethod
+    def next_action(self, context: AgentContext) -> Action:
```
dkolas, line 75: Does this redefinition of the abstract method do something?

src/steamship/agents/schema/action.py
```diff
@@ -21,6 +22,26 @@ class Action(BaseModel):
     output: Optional[List[Block]] = []
     """Any direct output produced by the Tool."""
 
+    def to_chat_messages(self) -> List[Block]:
+        tags = [
+            Tag(kind=TagKind.ROLE, name="function"),
```
dkolas, line 27: nit: you added FUNCTION to RoleTag, but didn't use it here

src/steamship/agents/schema/llm.py
```diff
@@ -15,3 +16,15 @@ class LLM(BaseModel, ABC):
     def complete(self, prompt: str, stop: Optional[str] = None) -> List[Block]:
         """Completes the provided prompt, stopping when the stop sequeunce is found."""
         pass
+
+
+# TODO(dougreid): should LLM and ConversationalLLM share a common parent?
```
dkolas, line 21: I could go either way on this.

src/steamship/agents/schema/llm.py
```diff
@@ -15,3 +16,15 @@ class LLM(BaseModel, ABC):
...
+class ConversationalLLM(BaseModel, ABC):
+    """ConversationalLLM wraps large language model-based backends that use a chat completion style interation.
+
+    They may be used with LLMAgents in Action selection, or for direct prompt completion."""
```
dkolas, line 25: This comment seems to imply that ConversationalLLM should inherit from LLM though

src/steamship/agents/schema/agent.py
```diff
@@ -57,4 +57,20 @@ def messages_to_prompt_history(messages: List[Block]) -> str:
                 as_strings.append(f"System: {block.text}")
             elif role == RoleTag.AGENT:
                 as_strings.append(f"Agent: {block.text}")
+            elif role == RoleTag.FUNCTION:
+                as_strings.append(f"Function: {block.text}")
         return "\n".join(as_strings)
+
+
+# TODO(dougreid): does this need to be a separate thing? or should we just use LLM agent, but with a ConversationalLLM.
```
dkolas, line 65: I vote for separate thing if it takes a specialization of LLM

src/steamship/agents/functional/functions_based.py
```diff
@@ -0,0 +1,60 @@
...
+    """Selects actions for AgentService based on OpenAI Function style LLM Prompting."""
+
+    PROMPT = """You are a helpful AI assistant. Only use the functions you have been provided with.
+
+NOTE: Some functions return images, video, and audio files. These multimedia files will be represented in messages as
+UUIDs for Steamship Blocks. When responding directly to a user, you SHOULD print the Steamship Blocks for the images,
+video, or audio as follows: `Block(UUID for the block)`.
+
+Example response for a request that generated an image:
+Here is the image you requested: Block(288A2CA1-4753-4298-9716-53C1E42B726B)."""
```
dkolas, line 25: Wow this is so much shorter

src/steamship/agents/examples/my_functions_based_assistant.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self._agent = FunctionsBasedAgent(
+            tools=[
+                SearchTool(),
+                DalleTool(),
+            ],
+            llm=ChatOpenAI(self.client, temperature=0),
+            conversation_memory=MessageWindowMessageSelector(k=2),
```
dkolas, line 31: This line doesn't do anything, right?  Should we remove?

dkolas (approved):
Looks good to me.  It seems like the function stuff makes everything a lot cleaner.  

The biggest awkwardness that I see is in having to go from chat history through window / search selection back into a temp file to run through the generator.  It seems like there should be a shortcut for this somehow.

src/steamship/agents/functional/functions_based.py
EniasCailliau: 🔥

src/steamship/agents/functional/functions_based.py
```diff
@@ -0,0 +1,60 @@
...
+        # TODO(dougreid): only use the intersection of the two sets to reduce duplication?
+        messages.extend(context.chat_history.select_messages(self.message_selector))
+
+        # get completed steps
+        actions = context.completed_steps
+        for action in actions:
+            messages.extend(action.to_chat_messages())
+
+        # call converse()
+        output_blocks = self.llm.converse(messages=messages, tools=self.tools)
```
EniasCailliau, line 58: Why not `.chat` looks like the market has converged to `chat` and `complete` as methods to call into the LLM

src/steamship/agents/functional/output_parser.py
```diff
@@ -0,0 +1,89 @@
...
+from steamship.data.tags.tag_constants import RoleTag
+
+
+# TODO(dougreid): extract shared bits from this and the ReACT output parser into a utility?
+class FunctionsBasedOutputParser(OutputParser):
+
+    tools_lookup_dict: Optional[Dict[str, Tool]] = None
+
+    def __init__(self, **kwargs):
+        tools_lookup_dict = {tool.name: tool for tool in kwargs.pop("tools", None)}
```
EniasCailliau, line 16:
```suggestion
        tools_lookup_dict = {tool.name: tool for tool in kwargs.pop("tools", [])}
```

defaulting to None will cause an exception to be raised since it's not iterable. Unsure if that's intentional. If it isn't we can default to an empty array.

src/steamship/agents/functional/output_parser.py
```diff
@@ -0,0 +1,89 @@
...
+            raise RuntimeError(
+                f"Could not find tool from function call: `{name}`. Known tools: {self.tools_lookup_dict.keys()}"
+            )
+
+        input_blocks = []
+        arguments = fc.get("arguments", "")
+        args = json.loads(arguments)
+        # TODO(dougreid): validation and error handling?
+
+        if args.get("text", None):
```
EniasCailliau, line 34:
```suggestion
        if text := args.get("text"):
```

1. We can use walrus operator to avoid 2x .get 
2. .get defaults to None

EniasCailliau: walrus operator is available for 3.8+ which is the minimum Python version for our SDK.

src/steamship/agents/llms/openai.py
```diff
@@ -38,3 +38,52 @@ def complete(self, prompt: str, stop: Optional[str] = None) -> List[Block]:
...
+
+    NOTE: By default, this LLM uses the `gpt-3.5-turbo` model. Valid model
+    choices are `gpt-3.5-turbo` and `gpt-4`.
+    """
+
+    generator: PluginInstance
+    client: Steamship
+
+    def __init__(
+        self, client, model_name: str = "gpt-4-0613", temperature: float = 0.4, *args, **kwargs
```
EniasCailliau, line 55:
```suggestion
        self, client, model_name: str = "gpt-4", temperature: float = 0.4, *args, **kwargs
```

`gpt-4-0613` is now the default.

EniasCailliau (approved):

src/steamship/agents/functional/functions_based.py
eob: I'm super curious to hear how variance-tolerant you've found the prompt to be -- I think we've all wrestled with the ReAct one in the past.

src/steamship/agents/functional/functions_based.py
```diff
@@ -0,0 +1,60 @@
...
+        # get prior conversations
+        messages.extend(
+            context.chat_history.search(context.chat_history.last_user_message.text, k=3)
+            .wait()
+            .to_ranked_blocks()
+        )
+
+        # get most recent context
+        # TODO(dougreid): only use the intersection of the two sets to reduce duplication?
+        messages.extend(context.chat_history.select_messages(self.message_selector))
```
eob, line 50: Is this sort of like the long-term memory component? (note: I haven't read ahead yet in case the answer lies there)

src/steamship/agents/llms/openai.py
eob: I think version-pegging it anyway is a good idea, since it makes it more clear to users what they're getting.

eob: Also looks like we'll need to update the comment on line 61

src/steamship/agents/llms/openai.py
```diff
@@ -38,3 +38,52 @@ def complete(self, prompt: str, stop: Optional[str] = None) -> List[Block]:
...
+        generator = client.use_plugin(
+            PLUGIN_HANDLE, config={"model": model_name, "temperature": temperature}
+        )
+        super().__init__(client=client, generator=generator, *args, **kwargs)
+
+    def complete(self, prompt: str, stop: Optional[str] = None) -> List[Block]:
+        options = {}
+        if stop:
+            options["stop"] = stop
+        action_task = self.generator.generate(text=prompt, options=options)
```
eob, line 73: generate_task?  (Or does the returned task represent a tool action?)

src/steamship/agents/llms/openai.py
```diff
@@ -38,3 +38,52 @@ def complete(self, prompt: str, stop: Optional[str] = None) -> List[Block]:
...
+        action_task.wait()
+        return action_task.output.blocks
+
+    def converse(self, messages: List[Block], tools: Optional[List[Tool]]) -> List[Block]:
+        # TODO(dougreid): this feels icky. find a better way?
```
eob, line 78: I get the ickyness here but fwiw can't think of anything better. To rubber duck it to GH in case it's useful:

1. The available tools is part of the data plane -- not prompt plane -- of this OAI model.
2. That means it needs to be passed in the `options` block to the `generate` method call.
3. To further the special-casey-ness, the thing passed to the options block also needs to be a special OAI-specific object.

The best suggestion I think I have would be to move `as_openai_function` into this class, since this class is related to OpenAI specifically, where as the tool base class isn't. But that doesn't seem to me something screaming to be done either.

src/steamship/agents/schema/action.py
```diff
@@ -21,6 +22,26 @@ class Action(BaseModel):
     output: Optional[List[Block]] = []
     """Any direct output produced by the Tool."""
 
+    def to_chat_messages(self) -> List[Block]:
```
eob, line 25: I'm really psyched that actions are becoming chat messages

eob (approved):
LGTM! I know there are a lot of small design decisions all over that feel somewhat arbitrary to glue together, but FWIW this feels like a really good way to think about this to me

src/steamship/agents/llms/openai.py
```diff
@@ -38,3 +38,52 @@ def complete(self, prompt: str, stop: Optional[str] = None) -> List[Block]:
...
+        temp_file = File.create(client=self.client, blocks=messages)
+        options = {}
+        if len(tools) > 0:
+            functions = []
+            for tool in tools:
+                functions.append(tool.as_openai_function())
+            options["functions"] = functions
+
+        action_task = self.generator.generate(input_file_id=temp_file.id, options=options)
```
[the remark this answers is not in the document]
douglas-reid: Not entirely sure I understand the concern behind the question, so I'll try to provide more info. The Chat history (w/ users) is distinct from the potential reasoning history of the agent itself. So, we combine the short-term/long-term memory with the agent-assistant messages in a sort of mega-list of blocks. This is roughly equivalent to the two sections of history + the scratchpad in the ReAct implementation.

src/steamship/agents/schema/agent.py
douglas-reid: nothing other than saying "yup, this still needs to be implemented". I don't know what the pythonically-appropriate way of marking that is.

src/steamship/agents/schema/llm.py
douglas-reid: probably needs to be updated.

src/steamship/agents/functional/output_parser.py
douglas-reid: I think I copy/pasted this from the other bit, so I'll need to update that too then.

src/steamship/agents/functional/functions_based.py
douglas-reid: this line, in particular, is short-term memory. the line above is the long-term bit. hopefully, there isn't overlap. I may go ahead and add the de-duping here to be safe.

src/steamship/agents/llms/openai.py
douglas-reid: fwiw, this bit was copy/pasted from the other OpenAI LLM. `action_task` is somewhat appropriate for the Agent, as the generation is really action selection. we could call it `tool_selection_task` or similar to clear it up?

src/steamship/agents/llms/openai.py
douglas-reid: Interesting. The ickiness to me is the `File.create` bit (as we are taking blocks from another file and then creating a temp file with just those blocks here), solely for the purpose of calling `generate`. 

The tool conversion bit doesn't bother me as much.

src/steamship/agents/schema/agent.py
dkolas: If you mark ConversationalLLMAgent as still inheriting from ABC, can you just omit it?

src/steamship/agents/functional/functions_based.py
douglas-reid: done now. ptal.

src/steamship/agents/examples/my_functions_based_assistant.py
douglas-reid: it allows the agent to pull in history from the context. So, we still send some slices from short-term memory in.

src/steamship/agents/schema/agent.py
douglas-reid: done.

src/steamship/agents/llms/openai.py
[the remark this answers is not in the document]
dkolas: 👍 .  I wonder if there will be a way to simplify this at some point, but definitely carry on for now.

src/steamship/agents/llms/openai.py
eob: Ah, I missed that. Yeahhh. I'm OK with it -- we'll keep evolving this, and this works pretty great already having played with it a good bit

eob:
LGTM! Fine w/ a merge on my end if everyone else is!

douglas-reid:
Thanks for the thorough reviews!  Looking forward to iterating on this as we keep moving forward.

PR #456, opened 2023-07-05
Add `ship prod` and `ship local` commands to CLI 

EniasCailliau:
This PR adds 2 new CLI cmd's + ads automatic invocation of __dir__ for local deployments. 

- `ship local` is an alias for `ship serve` 
- `ship prod` is an alias for `ship deploy && ship use` 

@eob I'm now calling `__dir__` before running the local HTTP server like you did in the old `run_server.py` script. Without it, the package doesn't register the Telegram webhook as `instance_init` is not hit.


src/steamship/cli/cli.py
```diff
@@ -402,6 +412,43 @@ def logs(
     )
 
 
+@click.command(name="prod")
```
douglas-reid, line 415: nit: i think we should use a different name. `prod` implies production, which is not exactly aligned with this use case (as I understand it). i get that you want to distinguish it from `local`, but I think this will lead to confusion (staging envs, etc.).

wdyt about `ship launch` ?

or even: `ship run local` and `ship run remote` ?

eob: I really like `run local` and `run remote`, but my worry is the reverse -- that `run remote` actually requires a full deployment (though the API technically supports a deployment that wouldn't become the default for new instances).

I think my dream solution would be something like:

run/serve local
run/serve remote <- deploys without making the version the default for new instances, then creates an instnace publish <- publishes the new version

I'm pretty OK with this being an evolving thing. What it is that they say? The hardest part of computer science is naming things?

eob (dismissed):
lgtm

eob:
@douglas-reid @EniasCailliau ok to merge this?

Since remaining convo is around the words, propose to just leave door open to renaming if it feels weird to use with current set.

douglas-reid:
> @douglas-reid @EniasCailliau ok to merge this?
> 
> Since remaining convo is around the words, propose to just leave door open to renaming if it feels weird to use with current set.

I think getting the naming right on a PR that is exclusively about aligning with user's expectations is worth nailing in terms of the verbs we are exposing. I'm happy to be out-voted, but this feels like something we should be able to nail -- and explain clearly to users.

what is the current naming?  it looks like `ship serve local` and `ship serve prod`. Is that correct?

I think using `prod` anywhere is a mistake. It implies something that this isn't.

I'm also a little worried that `serve` isn't quite right either, given that we aren't providing a full serving stack, and that it doesn't really align with any other Steamship concepts ( we talk of versions, instances, and deployments, but never of serving).

To me, it _feels_ like we are really asking people to do two different things (debug, and create an instance immediately -- which can also be used for debugging, but doesn't have to be). Are the commands really: `ship debug` and `ship create`?  Or something similar?  Or maybe even `ship run local` and `ship run blue-green`? 

One concern is that the people that don't care about `ship deploy && ship use` won't care that the new *version* immediately becomes the default version -- nor will they be concerned about _others_ accidentally grabbing a bad version.

Having a two-step command for people wanting to *debug* / *test* running in the cloud while *also* wanting to preserve an existing state seems ok to me. It _feels_ like here we just want to optimize for crowd that doesn't care about that case.

I think to fix everything, we'd need something like `ship deploy && ship promote && ship use`, where `deploy` doesn't make a version default (by default, maybe you have a flag like `-f` which forces update). And you'd need a short-hand for both, like: `ship stage` and `ship run`.  Then you could use something like `ship debug`, with options like `ship debug --local` and `ship debug --staged`.  `ship debug staged` could then get "smart" and call `ship deploy` for you if no staged version newer than the selected production version was found.

In any case, I think getting this right matters, because it tells people how we want them to use us, and how we think about deployments. But I won't take the step of blocking the PR.

eob:
Update from Slack convos:

It seems like everyone's OK with the following:

```
ship run local
ship run remote
```

with the additional property that `ship run remote` **interactively** asks if you want to deploy an new version first.

This achieves the objectives of:

* Having a simple command for both situations
* Having a command which doesn't deploy a production version without telling you 
* Not equating "Run on Steamship" with the notion of "production"

## Additional Observations

I think this also involves us removing:

* `ship serve` which is now just `ship run local`
* `ship use` which is now just `ship run remote`

PR #458, opened 2023-07-06
feat: Standardize AgentREPL handling of output; provide default `prompt`

eob:
This PR attempts to simplify the complexity of developing & creating an Agent in two ways:

1) It standardizes the handling of Tool/Agent output within the Repl (previously Tool output was handled by the REPL, but Agent output was handled by the Agent).

2) It provides a default `prompt` method if none is provided during REPL instantiation. This permits the 90% case of agent development to be nothing more than the agent definition itself, without the `prompt` method which sets up the debugging harness.


src/steamship/utils/repl.py
```diff
@@ -1,16 +1,31 @@
...
+    from termcolor import colored  # noqa: F401
+except ImportError:
+
+    def colored(text: str, **kwargs):
+        print(text)
+
+
+# Use either "gpt-3.5-turbo-0613" or "gpt-4-0613" here.
+# Other versions of GPT tend not to work well with the ReAct prompt.
+MODEL_NAME = "gpt-4-0613"
```
dkolas, line 27: Why does this have to be defined in the REPL?  Is there a way to refactor around this? It feels really strange that the REPL would be selecting a default model.

src/steamship/agents/examples/document_qa_agent.py
dkolas, line 1: What's the motivation for removing this one?  did it go somewhere else?  This seems like a very relevant example; 
I've pointed people at it a couple times in the last few days

dkolas (approved):
Looks good, though I wonder about the repl model thing below.

douglas-reid (commented):
This looks reasonable to me. The only thing I think we need to do is to add some discussion of `AgentContext` and how to create one, etc., somewhere. The `prompt()` testing bits covered that (and also introduced a bunch of confusion), but it would be great if we had a way to provide the same documentation somewhere convenient.

eob:
@douglas-reid yeah, I agree that once of the side benefits to having it in the agents was that it acted as a kind of documentation. 

I'll do another pass to see if I can find a way to work that documentation in there -- or at least clear links to it -- before going for a merge. 

The REPL is also printing `Block(UUID)` instead of the actual URL, when I generate images from agents, which makes me think I've missed something I'd intended in the output refactoring too.

src/steamship/agents/service/agent_service.py
```diff
@@ -79,3 +83,51 @@ def run_agent(self, agent: Agent, context: AgentContext):
         for func in context.emit_funcs:
             logging.info(f"Emitting via function: {func.__name__}")
             func(action.output, context.metadata)
+
+    @post("prompt")
+    def prompt(self, prompt: str, **kwargs) -> List[Block]:
```
douglas-reid, line 88: I don't think this is an issue. However, this does have one slight real difference with the REPL hosting. It means that _all_ agents will have an (extra) endpoint.

We probably want to disclose this somewhere obvious. Maybe in the docs for `AgentService`.

src/steamship/agents/service/agent_service.py
```diff
@@ -79,3 +83,51 @@ def run_agent(self, agent: Agent, context: AgentContext):
...
+        elif hasattr(self, "_agent"):
+            agent = self._agent
+
+        if not agent:
+            raise SteamshipError(
+                message="No Agent object found in the Agent Service. "
+                "Please name it either self.agent or self._agent."
+            )
+
+        self.run_agent(agent, context)
```
douglas-reid, line 130: I _feel_ like this is missing the `context.chat_history.append_assistant_message(output)` bit. Without that, the chat history will be missing all AI responses.

eob:
All the tests passed; the one that failed was a ghost in the system of an unrelated test, so going to merge.

src/steamship/agents/service/agent_service.py
eob: Done!

src/steamship/agents/service/agent_service.py
eob: Done!

PR #459, opened 2023-07-07
do not submit: POC of agent context persist and async status update

douglas-reid:
DO NOT SUBMIT

This PR is an attempt to sketch a possible path forward for "async" status reporting of agents, per discussion in meeting.

Main point of focus should be: [assistant_package.py](https://github.com/steamship-core/python-client/pull/459/files#diff-981ae5377c90472af81e1f10b111b270e544662498db1773783d41f738f7b1f1R71)

It is NOT meant to be an ideal path forward, but rather an exploration of the space that sparks discussions. It consists of a few key parts:
- making `AgentContext` serializable(-ish) and persisting it using `File` + `Block` tags.
- `invoke_later()` calls on the Agent service
- polling on a `/agent_context` endpoint.

A few small modifications could be made to such an approach:
- provide a more tailored `/agent_status` endpoint that only returns (completed) steps
- provide a `/known_contexts` endpoint that allows retrieval of in-flight contexts
- add `in-process steps` to `AgentContext` to provide even more insight
- more refined serialization.

DO NOT SUBMIT

### DIFF
--- a/src/steamship/agents/schema/context.py
+++ b/src/steamship/agents/schema/context.py
@@ -1,7 +1,11 @@
-from typing import Any, Callable, Dict, List
+import json
+from typing import Any, Callable, Dict, List, Optional
 
-from steamship import Block, Steamship, Tag
+from steamship import Block, File, Steamship, SteamshipError, Tag
 from steamship.agents.schema.action import Action
+from steamship.agents.schema.tool import Tool
+from steamship.data import TagKind
+from steamship.data.tags.tag_constants import ChatTag
 
 Metadata = Dict[str, Any]
 EmitFunc = Callable[[List[Block], Metadata], None]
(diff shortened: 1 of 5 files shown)


tests/assets/packages/assistant_package.py
```diff
@@ -0,0 +1,141 @@
...
+            )
+            output += block_text
+
+        context.emit_funcs.append(sync_emit)
+        self.run_agent(self._agent, context)
+        # TODO: is this right?
+        context.chat_history.append_assistant_message(output)
+        return output
+
+    @post("/async_prompt")
```
eob, line 70: Hot take:

Something I like about this is that I can see how it could evolve very cleanly to an HTTP Steaming-style response:

1) User posts to async_prompt
2) User gets back HTTP Stream URL (in addition to / instead of file_handle + task_id)
3) User can poll on file OR connect to the HTTP Stream URL for streaming updates

At that point, we'd basically have built-in support for poll-style async AND stream-style async using the same basic structures, and backed by the same persistence (file w/ blocks) underneath. 

Feels nice.

tests/steamship_tests/agents/test_context.py
```diff
@@ -0,0 +1,26 @@
...
+    ctx.completed_steps.append(
+        Action(
+            tool=SearchTool(),
+            input=[Block(text="where's waldo?")],
+            output=[Block(text="location unknown")],
+        )
+    )
+
+    file_handle = ctx.persist_to_file(client=client)
+    retrieved_ctx = AgentContext.hydrate_from_file(client=client, file_handle=file_handle)
```
eob, line 23: This feels like something that would be great to have completely independent of streaming tbh.

src/steamship/agents/schema/context.py
```diff
@@ -54,6 +58,132 @@ def get_or_create(
...
+                if tag.kind == "AgentContext":
+                    if tag.name == "metadata":
+                        new_ctx.metadata = json.loads(block.text)
+                    elif tag.name == "context_keys":
+                        from steamship.agents.schema.chathistory import ChatHistory
+
+                        new_ctx.chat_history = ChatHistory.get_or_create(
+                            client=client, context_keys=json.loads(block.text)
+                        )
+                    elif tag.name == "completed_steps":
```
eob, line 155: I think I need to do more learning on diff between the steps & the chat history --- I have a blurry understanding of when something wants to be in one and when in the other

douglas-reid:
This should be considered OBE at this point. Closing.

PR #460, opened 2023-07-07
feat: Slack Transport for Agent <> Slack communication

eob:
@dkolas this is the Slack transport that I wrote the other day. It works, but it's not polished and needs a bit of touch-up.

Before I do that, I was hoping for a quick scan for any high level design disagreements within so that I can fix those before polishing off the edges.

My thinking was to get something relatively simple in, along with documentation, and then we'll learn where the gremlins lie (as we did with Telegram) once people start to use it.

Also hoping it'll serve useful for us to consider a broader set of "Transports" together vis-a-vis developer/user UX issues ---- e.g. is there a universal UI/UX that manages transports? Are they ad hoc? Etc.


eob:
@douglas-reid I'm pretty sure this is ready for an initial review / edit / commit. 

It implements:

1) a basic Slack Transport that can send multi-modal message back
2) is fully tested using a mock slack server

Additional edits in this pr are:

1) I slightly refactored Dave's Telegram testing code to use variable name that port more easily across different transports
2) I added a `to_public_url` method on the Block class to facilitate the fact that we have a few ways of creating public urls on blocks

(2) there also surfaced a TODO on the engine for creating a signed url in aws for blocks (currently we only support this for files)

PR #464, opened 2023-07-10
feat(history): allow message deletion

douglas-reid:
Per request in Discord, this supports the ability to delete bits from chat history, including a full reset.


src/steamship/agents/schema/chathistory.py
```diff
@@ -230,3 +230,43 @@ def search(self, text: str, k=None) -> Task[SearchResults]:
...
+            for msg in selected_messages:
+                msg.delete()
+
+            self.refresh()
+            if self.is_searchable():
+                self.embedding_index.reset()
+                for msg in self.messages:
+                    for tag in msg.tags:
+                        if tag.kind == TagKind.CHAT and tag.name == ChatTag.CHUNK:
+                            # TODO(dougreid): figure out why tag.text gets lost.
```
dkolas, line 254: Tag.text is ephemeral.  It was added as a convenience, but it's not stored in the db. It can be derived from the block text + start index + end index.  

It would likely be better as a lazy-loaded property.

src/steamship/agents/schema/chathistory.py
```diff
@@ -230,3 +230,43 @@ def search(self, text: str, k=None) -> Task[SearchResults]:
...
+    def clear(self):
+        """Deletes ALL messages from the ChatHistory (including system).
+
+        NOTE: upon deletion, refresh() is called to ensure up-to-date history refs.
+        """
+        for block in self.file.blocks:
+            block.delete()
+
+        if self.is_searchable():
+            self.embedding_index.reset()
```
dkolas, line 270: Could consider just deleting the whole backing file here and starting over

dkolas (approved):
Looks good - comments below on Tag.text.  Maybe just as well to address that separately though.

src/steamship/agents/schema/chathistory.py
douglas-reid: a potential refinement if `reset` proves too problematic. i think the comments in `reset()` point at that possibility too.

douglas-reid:
Squash merging around test flake.

PR #469, opened 2023-07-11
fix(agent): don't add blocks that contain only punctuation

douglas-reid:
Trailing punctuation blocks are of little to no value (imho). This PR ensures that we don't add those blocks.

From:

```
Here is the image you requested: 
(image/png: B1A98152-A422-4B44-8FD9-D7C954E82D2A)
.
```

To:

```
Here is the image you requested: 
(image/png: B1A98152-A422-4B44-8FD9-D7C954E82D2A)
```

### DIFF
--- a/src/steamship/agents/functional/output_parser.py
+++ b/src/steamship/agents/functional/output_parser.py
@@ -1,5 +1,6 @@
 import json
 import re
+import string
 from typing import Dict, List, Optional
 
 from steamship import Block, MimeTypes, Steamship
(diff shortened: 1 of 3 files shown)


src/steamship/agents/functional/output_parser.py
```diff
@@ -1,13 +1,17 @@
...
 from typing import Dict, List, Optional
 
 from steamship import Block, MimeTypes, Steamship
 from steamship.agents.schema import Action, AgentContext, FinishAction, OutputParser, Tool
 from steamship.data.tags.tag_constants import RoleTag
 
 
-# TODO(dougreid): extract shared bits from this and the ReACT output parser into a utility?
+def is_punctuation(text: str):
+    return text in string.punctuation
```
dkolas, line 12: Do you want this to return true if ALL the text is punctuation, or only if it is a single character of punctuation?  The current implementation does the latter:
```
>>> s = ".."
>>> import string
>>> s in string.punctuation
False
```

dkolas (approved):
Looks good. One question below.  I'm good either way, just whatever you think best.

src/steamship/agents/functional/output_parser.py
douglas-reid: good point. updated. still doesn't filter strings like `....   ....   ...`, but that is so degenerate a case that I'm not sure we need to worry about it.

douglas-reid:
PTAL.

PR #474, opened 2023-07-12
feat(prs): add auto-rebase for main branch

douglas-reid:
This action _should_ rebase all PRs against the `main` branch on any push to the `main` branch.

Only way to figure out if it works is to create a PR and see what happens (as far as I can tell).

### DIFF
--- a/.github/workflows/rebase-main.yaml
+++ b/.github/workflows/rebase-main.yaml
@@ -1,0 +1,12 @@
+name: Rebase
+on:
+  push:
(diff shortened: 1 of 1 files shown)


douglas-reid:
> I mean, if this is all there is to it, great! I assume it will just fail if the merge can't happen automatically?

That's the idea. I'll see what GH Merge Queues do first. Then we can come back to this.

douglas-reid:
Closing for now

PR #476, opened 2023-07-13
fix(test): attempt to deflake 'test_telegram' in staging

douglas-reid:
It looks like in Staging environments, the URL encoding isn't happening (or isn't always happening?). In any case, this attempts to allow flexible results.

### DIFF
--- a/tests/assets/plugins/taggers/plugin_parser.py
+++ b/tests/assets/plugins/taggers/plugin_parser.py
@@ -1,5 +1,6 @@
 import logging
 import re
+import time
 
 from steamship import Block, DocTag, Tag
 from steamship.data import TagKind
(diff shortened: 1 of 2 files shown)


dkolas (approved):
This is fine for the moment.

Staging has a build deployed that fixes the bug with URL-encoded GET params, which is why the behavior is now different in staging.  Once that build is on prod as well, we'll only need the one version of this.  Sorry I didn't catch it when I deployed the other build.

eob (approved):
LGTM

douglas-reid:
> This is fine for the moment.
> 
> Staging has a build deployed that fixes the bug with URL-encoded GET params, which is why the behavior is now different in staging. Once that build is on prod as well, we'll only need the one version of this. Sorry I didn't catch it when I deployed the other build.

SG. Filed SHIP-770 to track.

PR #477, opened 2023-07-14
fix(tools): add better description for vectorsearch tool and a test

douglas-reid:
OpenAI functions for QASearch weren't being reliably called. This creates a better tool description (the human description is used in the functions-based agent). A test is added to validate continued ability to work.

### DIFF
--- a/src/steamship/agents/tools/question_answering/vector_search_learner_tool.py
+++ b/src/steamship/agents/tools/question_answering/vector_search_learner_tool.py
@@ -15,7 +15,7 @@ class VectorSearchLearnerTool(VectorSearchTool):
     name: str = "VectorSearchLearnerTool"
     human_description: str = "Learns a new fact and puts it in the Vector Database."
     agent_description: str = (
-        "Used to remember a fact. Only use this tool if someone asks to remember or learn something. ",
+        "Used to remember a fact. Only use this tool if someone asks to remember or learn something.",
         "The input is a fact to learn. ",
         "The output is a confirmation that the fact has been learned.",
     )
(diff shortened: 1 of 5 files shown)


src/steamship/agents/tools/question_answering/vector_search_learner_tool.py
```diff
@@ -15,7 +15,7 @@ class VectorSearchLearnerTool(VectorSearchTool):
     name: str = "VectorSearchLearnerTool"
     human_description: str = "Learns a new fact and puts it in the Vector Database."
     agent_description: str = (
-        "Used to remember a fact. Only use this tool if someone asks to remember or learn something. ",
+        "Used to remember a fact. Only use this tool if someone asks to remember or learn something.",
```
eob, line 18: I think I wrote this original description with a bug.

In Python:

("S1", "S2", "S3") is a tuple of strings, but
("S1" "S2" "S3") is a concatenation of strings.

So:

* The commas in this list probably need to get removed
* The space after something. should be added back

Then the result will be the single string that was intended.

Alternative would be we use a different method of string concat that's less error prone -- my bad!

src/steamship/agents/tools/question_answering/vector_search_qa_tool.py
```diff
@@ -27,7 +27,7 @@ class VectorSearchQATool(VectorSearchTool):
     """Tool to answer questions with the assistance of a vector search plugin."""
 
     name: str = "VectorSearchQATool"
-    human_description: str = "Answers questions with help from a Vector Database."
+    human_description: str = "Answers questions about user data. Can access personal data, as it is a private store of information."
```
eob, line 30: So much better!

eob (changes requested):
Change requested related to a string bug that's my fault, but rubber stamp of approval with that change made!

Thank's for catching this and also adding the unit test!!

src/steamship/agents/tools/question_answering/vector_search_learner_tool.py
douglas-reid: Should be fixed now.

eob (approved):
LGTM!

PR #478, opened 2023-07-14
feat(agents): support persistent contexts in prompt and repl

douglas-reid:
Allow `prompt` to take a `context_id` argument that will allow frontends to maintain history/memory across calls. Then set up the REPL to exploit that when the method is `prompt`. This will allow for repl loops with memory.


eob (approved):
LGTM. I didn't see this PR and also made this one: https://github.com/steamship-core/python-client/pull/480

Your choice which to use. I think this one might also need to put the context ID in the HTTP REPL too for full coverage

douglas-reid:
@eob I can close this now, correct?

PR #479, opened 2023-07-14
feat(invoke): add invoke handlers for blocks and task

douglas-reid:
Hot take:  Getting raw dicts back from `invoke` is a usability concern. Since we are pushing `AgentService` with a builtin `prompt(): -> List[Block]`, we should have an easy way to actually get those blocks directly.

Proposed usage:

```python
answer_blocks = agent.blocks_from_invoke("prompt", prompt="What did I eat for dinner?", context_id="doug-dinner-discussion')

for block in answer_blocks:
   print(block.text)
```

### DIFF
--- a/src/steamship/data/package/package_instance.py
+++ b/src/steamship/data/package/package_instance.py
@@ -5,7 +5,7 @@ from typing import Any, Dict, List, Optional, Type
 
 from pydantic import BaseModel, Field
 
-from steamship import SteamshipError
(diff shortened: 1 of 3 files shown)


eob (dismissed):
The generalized version of this is something I really wish we had... to get some way to re-build the objects on the other side of the HTTP wire.

dkolas (dismissed):
LGTM

PR #480, opened 2023-07-14
bugfix: Bug in which context_id wasn't stable across REPL messages

eob:


src/steamship/utils/repl.py
```diff
@@ -178,7 +182,7 @@ def colored(text: str, color: str, **kwargs):
 
         while True:
             input_text = input(colored(text="Input: ", color="blue"))  # noqa: F821
-            output = responder(input_text)
+            output = responder(prompt=input_text, context_id=self.context_id)
```
douglas-reid, line 185: @eob I like this PR. And this seems totally fine. Do we need to be concerned at all with people trying to use the REPL for functions that aren't `prompt` and tripping on the `context_id` arg, or should we enforce that you must take `context_id` as an arg, or some other option.

douglas-reid (dismissed):

src/steamship/utils/repl.py
eob: Yeah that's a really good question. And I think the problem you're observing extends even to the `prompt: str` argument, right? There's an unstated requirement about the signature of `responder`.

Proposal: what if we allowed `ToolRepl` and `AgentRepl` to make these kinds of assumptions now, since that's probably the 90% case, but then we planted a stake in the ground (your enforcement suggestion above?) about the "chat I/O contract" that an AgentService must, at minimum, provide.

Here, I suppose that's implicitly:

```
prompt(prompt: str, context_id: str) => List[Block]
```

though I imagine if we were being proactive about defining it, we'd probably want to extend that a bit:

```
prompt(prompt: Optional[str, Block], context_id: str) => List[Block]
```

Or even something further..

douglas-reid: I think we should standardize. Maybe with some nasty combo of lists

PR #481, opened 2023-07-14
Simple agent example without Tools

eob:
There's a great many folks whose use case does not yet involve tools --- only a prompt and chat history.

The prompt may be:
- Static
- Parameterized (e.g. CharacterAI)
- Dynamic (e.g. based on vector lookup before LLM invocation)

This PR is a first stab at developing such a base agent. It implements a simple agent with a personality and no tools.

@douglas-reid am I right that building an `AgentWithoutTools` base class is faster here than using the function-calling base class? My thinking was that removing the reasoning loop could be meaningful.... but I might be stuck in a react-based mindset that's no longer applicable for the function-calling world

### DIFF
--- a/src/steamship/agents/examples/logs/shiplog--2023-07-14--14:38:15.log
+++ b/src/steamship/agents/examples/logs/shiplog--2023-07-14--14:38:15.log
@@ -1,0 +1,4 @@
+[Client] Creating/Fetching workspace with handle/id: default/None.
+[Client] Switched to workspace default/562A13D4-28E4-4A6D-8350-0DE11823580A
+[Client] Creating/Fetching workspace with handle/id: gutsy-river-voo95/None.
+[Client] Switched to workspace gutsy-river-voo95/7ED6EF05-4EDB-4962-8554-EAB4E9DC893A
--- a/src/steamship/agents/examples/simple_character_without_tools.py
+++ b/src/steamship/agents/examples/simple_character_without_tools.py
@@ -1,0 +1,49 @@
+from typing import Type
+
+from pydantic import Field
+
+from steamship.agents.llms.openai import ChatOpenAI
+from steamship.agents.service.agent_service import AgentService
+from steamship.agents.simple.agent_without_reasoning import AgentWithoutReasoning
+from steamship.invocable import Config
+from steamship.utils.repl import AgentREPL
+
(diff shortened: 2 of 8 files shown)


src/steamship/agents/examples/logs/shiplog--2023-07-14--14:38:15.log
```diff
@@ -0,0 +1,4 @@
+[Client] Creating/Fetching workspace with handle/id: default/None.
```
douglas-reid, line 1: @eob did you mean to commit the logs for your runs?

src/steamship/agents/examples/simple_character_without_tools.py
```diff
@@ -0,0 +1,49 @@
...
+from steamship.utils.repl import AgentREPL
+
+DEFAULT_NAME = "Harry Potter"
+DEFAULT_TAGLINE = "famous wizard, the one who lived, defeater of Voldemort"
+DEFAULT_PERSONALITY = """You chat with your fans about your adventures in the wizarding world. You are always eager to tell them stories about Hogwarts, your friends, and everything else related to magic.
+
+Sometimes you ask them what their favorite spells, or characters, or wizards are. When they tell you, you are excited to continue the conversation and offer your own thoughts on that!"""
+
+
+class SimpleCharacterWithoutTools(AgentService):
```
douglas-reid, line 18: This is totally fine. But, it opens the question of: what is the transition story?

If I make a Harry Potter agent without Tools and love it, but one day wish it could also draw pictures of castles, how do I update my Harry Potter agent to now have a Tool?

src/steamship/agents/examples/simple_character_without_tools.py
```diff
@@ -0,0 +1,49 @@
...
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+        prompt = (
+            f"""You are {self.config.name}, {self.config.tagline}.\n\n{self.config.personality}"""
+        )
+
+        # The agent's planner is responsible for making decisions about what to do for a given input.
+        self._agent = AgentWithoutReasoning(
+            llm=ChatOpenAI(self.client, model_name="gpt-4"),
```
douglas-reid, line 39: you _may_ want to give it a temperature of something more on the creative side if reasoning with functions isn't important.

PR #482, opened 2023-07-14
Two minimum viable agents

eob:
These are two agents which appear in the Agent Guidebook.

I was thinking it'd be good to make sure these are captured in the official codebase so that we could begin to write tests against them and then eventually find a way to import them **from this repo** into the website itself directly.

### DIFF
--- a/src/steamship/agents/examples/chatgpt_api.py
+++ b/src/steamship/agents/examples/chatgpt_api.py
@@ -6,9 +6,17 @@ This will result in an agent that effectively acts like ChatGPT.
 from steamship.agents.functional import FunctionsBasedAgent
 from steamship.agents.llms.openai import ChatOpenAI
 from steamship.agents.service.agent_service import AgentService
+from steamship.utils.repl import AgentREPL
 
 
 class MyAssistant(AgentService):
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
         self._agent = FunctionsBasedAgent(llm=ChatOpenAI(self.client), tools=[])
+
+
+if __name__ == "__main__":
+    AgentREPL(
+        MyAssistant,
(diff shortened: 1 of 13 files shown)


src/steamship/agents/examples/logs/shiplog--2023-07-15--00:02:11.log
```diff
@@ -0,0 +1,7 @@
+[Client] Creating/Fetching workspace with handle/id: default/None.
```
douglas-reid: did you mean to include these logs?  should we add a .gitignore ?

src/steamship/agents/examples/chatgpt_api.py
```diff
@@ -0,0 +1,22 @@
...
+class MyAssistant(AgentService):
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self._agent = FunctionsBasedAgent(llm=ChatOpenAI(self.client), tools=[])
+
+
+if __name__ == "__main__":
+    AgentREPL(
+        MyAssistant,
+        agent_package_config={"botToken": "not-a-real-token-for-local-testing"},
```
douglas-reid, line 21: do you need `botToken` config here ?

src/steamship/agents/examples/logs/shiplog--2023-07-15--00:02:11.log
eob: Lol you were so fast with your slack notifications you caught the PR before I removed the logs :)

Added a .gitignore!

src/steamship/agents/examples/chatgpt_api.py
eob: Argh. Sloppy of me. Thanks for the catch!

douglas-reid (approved):

PR #486, opened 2023-07-14
fix(tests): add retries to model checkpoint test

douglas-reid:
There is a test flake around DEFAULT checkpoints when multiple test runs happen at the same time. This PR is an attempt to solve that by attempting to rerun the checks until the hashes match.

I don't know if that's the best way to handle this issue, but it was all I could think of at the moment. Someone with a deeper understanding of the intended behavior here could possibly come up with a better solution.


tests/steamship_tests/plugin/unit/trainable/test_model_checkpoints.py
```diff
@@ -42,20 +42,41 @@ def test_model_checkpoint_save_load():
...
+        )
+        _, model_version = checkpoint_downloaded.download_model_bundle()
+        if update_version != model_version:
+            attempts -= 1
+            update_version = checkpoint_1.upload_model_bundle()
+        else:
+            version_mismatch = False
+
+    assert attempts > 0
     _test_folders_equal(
```
douglas-reid, line 62: i wonder if I can just use `_test_folders_equal` instead of hashing each upload/download.

PR #489, opened 2023-07-17
bugfix: Tell isort to behave in a black-compliant way

eob:
isort's default semantics conflict with black's default semantics. Sometimes this causes an endless loop of disagreement between them during the git precommit hooks.

This configuration line tells isort to behave in a black-compliant way.


douglas-reid (approved):
Hallelujah!

PR #493, opened 2023-07-18
feat(agents): add caching for LLM calls and Actions

douglas-reid:
This is an initial PR for adding caching functionality to Agents. It is aimed at reducing cost and latency for repeated requests.

It attempts to provide two levels of caching:
(1) prompts to the LLMs (prompt-to-Action cache) and (2) action executions (Action-to-outputs cache).

Cache behavior can be dynamically control per call to `prompt`.

NOTE: This PR is NOT introducing semantic caches (text/block uuids must match precisely).
FOLLOW-ONS: Allow Tool-default opt-in/opt-out of caching

### DIFF
--- a/src/steamship/agents/schema/cache.py
+++ b/src/steamship/agents/schema/cache.py
@@ -1,0 +1,187 @@
+import hashlib
+import json
+import logging
+from typing import Dict, List, Optional
+
+from steamship import Block, MimeTypes, Steamship
+from steamship.agents.schema.action import Action, FinishAction
+from steamship.utils.kv_store import KeyValueStore
+
+
+def _blocks_to_cache_dict(value: List[Block]) -> Dict[str, any]:
+    """Attempts to convert a list of blocks to key-store-safe dictionary.
+
+    Convention: {'blocks':[{'id': <block_id>}, {'text': <some_text>}]}
+    """
+
+    if not value:
+        return {}
+    if not isinstance(value, list):
+        return {}
+    blocks = []
+    for block in value:
+        if block.id:
+            blocks.append({"id": block.id})
+        else:
+            # TODO(dougreid): safe assumption about temporary blocks?
+            blocks.append({"text": block.text})
+    return {"blocks": blocks}
+
+
+def _blocks_from_cache_dict(client: Steamship, value: Dict[str, any]) -> List[Block]:
+    """Attempts to convert a key-store-safe dictionary to a list of blocks.
+
+    Convention: {'blocks':[{'id': <block_id>}, {'text': <some_text>}]}
+    """
+    if block_list := value.get("blocks"):
+        return_blocks = []
+        for b in block_list:
+            if block_id := b.get("id"):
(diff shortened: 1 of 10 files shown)


src/steamship/agents/schema/cache.py
```diff
@@ -0,0 +1,187 @@
...
+    for block in value:
+        if block.id:
+            blocks.append({"id": block.id})
+        else:
+            # TODO(dougreid): safe assumption about temporary blocks?
+            blocks.append({"text": block.text})
+    return {"blocks": blocks}
+
+
+def _blocks_from_cache_dict(client: Steamship, value: Dict[str, any]) -> List[Block]:
```
eob, line 31: Really curious why serializing the whole block didn't work out. You mentioned the `urls` not serializing? 

You're probably way ahead of me on this, but could we just add that into the JSON structure?

src/steamship/agents/schema/context.py
```diff
@@ -38,6 +39,14 @@ def id(self) -> str:
     """Called when an agent execution has completed. These provide a way for the AgentService
     to return the result of an agent execution to the package that requested the agent execution."""
 
+    action_cache: ActionCache
```
eob, line 42: nit: I think this and llm_cache below technically should be `Optional[]` types since they're None-able

src/steamship/agents/service/agent_service.py
```diff
@@ -15,18 +15,90 @@ class AgentService(PackageService):
...
 
     ###############################################
     # Tool selection / execution
     ###############################################
 
-    def run_action(self, action: Action, context: AgentContext):
+    def next_action(self, agent: Agent, input_blocks: List[Block], context: AgentContext) -> Action:
+        action: Action = None
+        if context.llm_cache:
+            action = context.llm_cache.lookup(key=input_blocks)
```
eob, line 41: I think really good idea to log the cache hit as well --- wearing the hat of a developer, it would feel like a positive reward to see that I hit cache (and didn't use tokens). 

Actually.. could be cool if the log had a field for tokens saved!

eob (approved):
@douglas-reid this is really solid.

LGTM and feel free to merge as-is; also left a few small comments in there that might inform a few tiny changes.

Re: the Block hydration thing -- do you want to put heads together to see if we can figure out how to enable them to be directly serializable? Unsure how big a challenge it'll end up being since I haven't tried yet like you were.

src/steamship/agents/schema/cache.py
douglas-reid: maybe. i think they are just python properties at this point. i got a little lost in the expectations of serialization and rather than try to fix that, i just tried to work around it.

worth addressing again, especially if it causes latency issues.

src/steamship/agents/service/agent_service.py
douglas-reid: added cache log line.

dkolas (approved):

PR #496, opened 2023-07-24
feat(cli): add delete command for workspaces and instances

douglas-reid:
Add a `ship delete` that will allow rapid CLI-based deletion of workspaces and instances.

### DIFF
--- a/src/steamship/cli/cli.py
+++ b/src/steamship/cli/cli.py
@@ -9,7 +9,7 @@ from typing import Optional
 import click
 
 import steamship
-from steamship import Steamship, SteamshipError
+from steamship import PackageInstance, Steamship, SteamshipError
 from steamship.base.configuration import DEFAULT_WEB_BASE, Configuration
 from steamship.cli.create_instance import (
     config_str_to_dict,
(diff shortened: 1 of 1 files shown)


src/steamship/cli/cli.py
```diff
@@ -547,6 +547,83 @@ def logs(
...
+                f"Deleting instance '{instance}' in workspace '{workspace}'... ",
+                nl=False,
+                fg="green",
+            )
+            pkg_inst.delete()
+            click.secho("Done.", fg="green")
+            return
+        except SteamshipError as e:
+            click.secho(
+                f"⚠️ Failed to delete instance '{instance}' in workspace '{workspace}': {e.code}",
```
dkolas, line 621: Should this be `e.message` instead?

dkolas (approved):
Looks very useful!  One super minor note/question.

src/steamship/cli/cli.py
douglas-reid: Maybe. I initially like `ObjectNotFound` because it was short and sweet. But updated to `e.message` now.

PR #498, opened 2023-07-25
RFC: Beginnings of "socially intelligent cron jobs"

eob:
Right now, we're capable of scheduling a package invocation for a future time. That means we technically support an agent reaching out to tell someone something, but the reality of working with that is a bit tricky:

- What if some other conversation in the interim renders the original message moot? Or socially weird?
- What if too many outreach messages get enqueued, resulting in the agent being a psycho stalker that keeps SMS'ing you "hey! hey! hey!"

ENTER:

# The Socially Intelligent CROD Job!

(sorry I'm in a silly mood..)

Here's the rough idea:

- You can enqueue message sends (to start; more complex stuff in V2) with different kinds of invalidation flags.
  - Example invalidation flags: "If another message send is sent with this key" or "if the user messages me back before the send"
- Then, we schedule a maybe_send callback for the time when the send was scheduled. But it isn't guaranteed to succeed if some kind of invalidation event occurred.

## Examples (pseudocode)

1. Are you still there?

```
agent.send(answer)
agent.social_cron("10 minutes", "Are you still there?", expire_if_user_responds=True, delete_chat_after_send=True)
```

2. Hey! What's cookin?

```
best_friend.send(response)
agent.social_cron("1 day + rand(1 day)", "What's cookin? How's your day?", expire_if_user_responds=True)
```

### DIFF
--- a/src/steamship/agents/examples/telegram_bot_with_callback.py
+++ b/src/steamship/agents/examples/telegram_bot_with_callback.py
@@ -2,13 +2,15 @@ from typing import Type
 
 from pydantic import Field
 
+from steamship import Block
 from steamship.agents.functional import FunctionsBasedAgent
 from steamship.agents.llms.openai import ChatOpenAI
 from steamship.agents.mixins.transports.steamship_widget import SteamshipWidgetTransport
 from steamship.agents.mixins.transports.telegram import TelegramTransport, TelegramTransportConfig
+from steamship.agents.schema import AgentContext
 from steamship.agents.service.agent_service import AgentService
 from steamship.agents.tools.image_generation.stable_diffusion import StableDiffusionTool
-from steamship.invocable import Config
+from steamship.invocable import Config, post
 from steamship.utils.repl import AgentREPL
 
 SYSTEM_PROMPT = """You are Picard, captain of the Starship Enterprise.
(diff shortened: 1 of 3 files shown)


douglas-reid:
this idea has tons of potential.

aside: i wonder which has more resonance with devs and users: "socially-intelligent" or "context-aware" or something less jargon-y like "conditional reminder" or even "if-not-then" (aside-aside: do people prefer to describe in the negative or positive -- if <timer x> expired, do <action y> vs. if not <action y>, do <action z>).

my sketch expanding upon it:

```python
agent.schedule(
  action=FutureAction(tool="GreetingTool", input=user.name),
  condition=NoUserMessagesCondition(),
  delay="1d", 
  recurring=True,
  context=ctx
)
```

eob:
I like this expansion a lot. Also that it's an Action your scheduling rather than a message -- fits better into the rest of the Action/Tool framing. 

This is me just piling on ideas, but a few other things that I was thinking could be interesting (though maybe indulgent for v1?) could be:

- A bit of randomness
- Some kind of time-zone awareness of the user.. combined with an enum of human-style time blocks?

E.g.:

```
send this tomorrow-morning
send this late at night
send this on the weekend
send this around dinnertime
```

PR #499, opened 2023-07-25
Tweak ImporterPipelineMixin and fix youtube importer

EniasCailliau:
This PR tweaks the IndexerPipelineMixin based on my experience with `embed-chain` and `langchain`. Very open to feedback on this one. On the one hand, the Filtype + mimeType support feels bloated, but also safer. I'm mostly concerned about situations where it's not clear from the data type which blockifier/loader we should use. 

Examples: 
* Wikipedia -> Webpage loader or Wikipedia loader 
* Youtube -> Audio + Transcription or Transcription loader 

Here's a list of concepts I liked and how I tried to introduce them: 

1. **FileType controls how the file is being loaded&parsed.** This is a nice addition since you can parse a webpage with a Wikipedia parser OR a website parser by changing the FileType. 
2. **The **document** object.** We don't need Documents as it has the same functionality as Files. However, I thought it would be nice to introduce a `_type` tag that helps us distinguish Documents (that are indexed) VS other files such as chatHistory. We can use this tag to filter for our front end. I held back on going all in and introducing a Document class that has the field pre-filled and would filter its search method on this field. That way Document.search() would only find documents and not chatfiles. 
2. **The metadata attribute on documents**. I had a ton of fun just adding information to the metadata of Documents in LangChain. Felt very Python-friendly vs storing Tags on Files or Blocks. I introduced a simple getter & helper function to emulate the same access patterns. 
3. **Fix the youtube importer** by using transcripts vs audio 

Already did some preliminary testing, will add more tomorrow morning.


src/steamship/invocable/mixins/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+
+from steamship import Tag
+
+
+class FileType(str, Enum):
+    YOUTUBE = "YOUTUBE"
+    PDF = "PDF"
+    WEB = "WEB"
+    TEXT = "TEXT"
+
```
eob, line 12: (continuing our in person chat) -- I think relying on the existing MimeType we've got is probably simplifying here.

PDF -> MimeTypes.PDF
WEB -> MimeTypes.HTML
TEXT -> MimeTypes.TXT
YOUTUBE You could have a few options and then build in the special casing just as you'd already done. E.g. could extend the MimeTypes class to include things like:

```
    YOUTUBE_TRANSCRIPT = "application/youtube-transcript"
    YOUTUBE_VIDEO = "application/youtube-video"
```

src/steamship/invocable/mixins/blockifier_mixin.py
```diff
@@ -34,11 +33,11 @@ def blockify(
         """
 
         file = File.get(self.client, _id=file_id)
-        update_file_status(self.client, file, "Blockifying")
+        file.add_or_update_metadata("status", "Blockifying")
```
eob, line 36: Maybe `add_or_update_tag`? (If I'm right above that the `metadata` object has different semantics than the `tags` array

src/steamship/invocable/mixins/blockifier_mixin.py
```diff
@@ -47,17 +46,21 @@ def blockify(
 
         if blockifier_handle:
             plugin_instance = self.client.use_plugin(blockifier_handle)
-        elif _mime_type == MimeTypes.PDF:
+        elif file_type == FileType.PDF or (file_type is None and _mime_type == MimeTypes.PDF):
```
eob, line 49: _mime_type == MimeTypes.PDF:

src/steamship/invocable/mixins/blockifier_mixin.py
```diff
@@ -47,17 +46,21 @@ def blockify(
...
             plugin_instance = self.client.use_plugin("pdf-blockifier")
+        elif file_type in (FileType.YOUTUBE, FileType.TEXT) or (
```
eob, line 51: mime_type in (MimeTypes.TXT, MimeTypes.MKD, MimeTypes.YOUTUBE_TRANSCRIPT)

src/steamship/invocable/mixins/blockifier_mixin.py
```diff
@@ -47,17 +46,21 @@ def blockify(
...
+            file_type is None and _mime_type in [MimeTypes.MKD, MimeTypes.TXT]
+        ):
+            plugin_instance = self.client.use_plugin("markdown-blockifier-default")
+        elif file_type == FileType.WEB:
```
eob, line 55: MimeTypes.HTML

eob (changes requested):
Thanks for taking a pass at this -- really appreciate the perspective from having worked with some of the other doc loading pipelines.

Most of the comments are around the `metadata` interaction with tag semantics and the introduction of a duplicate field called FileType which is basically MimeType by a different name

src/steamship/data/block.py
```diff
@@ -161,6 +161,10 @@ def query(
         )
         return res
 
+    @property
+    def metadata(self):
```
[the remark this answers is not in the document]
EniasCailliau: Hmmm. Damn completely forgot about that. In my head kind was a key. Shows how mental models shift as you're using different tooling. 

I think get_tag would help a bit.

EniasCailliau:
> 

Thanks for the review. Let me decompose some of the ideas and propose them as separate PRs. Let's drop the notion of FileType for now. You're right, most of it is covered by mimetype.

src/steamship/invocable/mixins/blockifier_mixin.py
EniasCailliau: Correct. Yes. I do think it's a nice addition to have. But since the indexer pipeline is the only one using it we can stick I'll change the update_file_status to be a bit more flexible.

PR #507, opened 2023-07-28
[SHIP-768] Receive headers in invocations from proxy

dkolas:
Receive headers in the invocation context from the proxy.

Depends on https://github.com/nludb/nludb/pull/603


douglas-reid (commented):
Do we care at all about filtering out certain headers?  Like, package owners should be able to see the Steamship creds sent by package instance users, or similar ?

douglas-reid:
> Do we care at all about filtering out certain headers? Like, package owners should be able to see the Steamship creds sent by package instance users, or similar ?

Oh, I see the filtering in the companion PR.

douglas-reid (approved):

PR #510, opened 2023-08-01
fix: fix path in sphynx config file for new subfolder location

eob:


eob:
(note: working live w/ Max as shared screen code reviewer)

PR #512, opened 2023-08-08
Support engine-registered development runs in the client

eob:
- Fix response unpacking bug in local_server -- It should have been unwrapping the Service response like the Steamship Proxy
- Fix the HttpREPL, which depended upon the improperly unwrapped response (see above)
- Upon `ship run local`, register the local running NGROK URL with the Engine so that it can make async callbacks to it.


dkolas (approved):
I don't see any issues so far!  Exciting stuff :)

dkolas (approved):
LGTM

PR #514, opened 2023-08-14
Add example slack bot

eob:
Adds an example of a Slack Bot


douglas-reid (approved):

src/steamship/agents/examples/slack_bot.py
```diff
@@ -0,0 +1,61 @@
...
+from steamship.agents.functional import FunctionsBasedAgent
+from steamship.agents.llms.openai import ChatOpenAI
+from steamship.agents.mixins.transports.slack import SlackTransport, SlackTransportConfig
+from steamship.agents.mixins.transports.steamship_widget import SteamshipWidgetTransport
+from steamship.agents.service.agent_service import AgentService
+from steamship.agents.tools.image_generation.stable_diffusion import StableDiffusionTool
+from steamship.invocable import Config
+from steamship.utils.repl import AgentREPL
+
+SYSTEM_PROMPT = """You are Assistant - a Slack bot that generates Stable Diffusion images."""
```
douglas-reid, line 12: Dude, without the chunk of prompt about dealing with Blocks, the LLM won't handle them correctly.

You don't need to set a SYSTEM_PROMPT like this, really.  The bit about generating images is supplied by the functions in the prompt. And the bit about being Slack _may_ cause the LLM to generate slack output that might mess with raw text you want to send to Slack.

douglas-reid (changes requested):
Fix issues with the prompt. I missed this in the first review.

src/steamship/agents/examples/slack_bot.py
eob: 🙏 💯 this is the fix for the weirdness I was getting in my other bot -- thank you!

PR #516, opened 2023-08-15
fix(agents): update telegram bot example config

douglas-reid:
When using Mixins, Mixin config should be a parent class for `AgentService` Config types.

### DIFF
--- a/src/steamship/agents/examples/telegram_bot.py
+++ b/src/steamship/agents/examples/telegram_bot.py
@@ -49,7 +49,7 @@ class TelegramBot(AgentService):
 
     """
 
-    class TelegramBotConfig(Config):
+    class TelegramBotConfig(TelegramTransportConfig):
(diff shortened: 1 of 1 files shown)


dkolas (approved):

src/steamship/agents/examples/telegram_bot.py
```diff
@@ -49,7 +49,7 @@ class TelegramBot(AgentService):
 
     """
 
-    class TelegramBotConfig(Config):
+    class TelegramBotConfig(TelegramTransportConfig):
```
eob, line 52:
```suggestion
    class TelegramBotConfig(TelegramTransportConfig):
        """"Configuration object for new TelegramBot instances.
        
        The config object for an AgentService must be, at minimum, the union of the config
        objects of its mixins. In this case, that means it inherits from TelegramTransportConfig.
        """
        
```

eob (dismissed):

PR #517, opened 2023-08-15
Handle None for parse_inbound

GitOnUp:
parse_inbound takes the `message` from `self._parse_inbound`, which can in some cases be None, causing issues when we try to access `.url` and `.text` off it.


douglas-reid (approved):
Thanks!

PR #518, opened 2023-08-16
bugfix: Agents should only respond to unresponded-to user messages

eob:
When `run_agent` is invoked, it locates the last user message in `context.chat_history` and responds to it regardless of whether that message has already been responded to.

This PR causes `run_agent` to exit early, doing nothing, if the last user message had already been marked as having been responded to.

### DIFF
--- a/src/steamship/agents/service/agent_service.py
+++ b/src/steamship/agents/service/agent_service.py
@@ -131,15 +131,48 @@ class AgentService(PackageService):
                 context.action_cache.update(key=action, value=action.output)
 
     def run_agent(self, agent: Agent, context: AgentContext):
-        # first, clear any prior agent steps from set of completed steps
-        # this will allow the agent to select tools/dispatch actions based on a new context
+        """Run the agent.
+
+        This is the root method that runs when the agent awakes to do work. Any chat messages to response to should
+        have already been appended to the ChatHistory associated with the AgentContext.
+        """
+
+        # Clear any prior agent steps from set of completed steps.
+        # This will allow the agent to select tools/dispatch actions based on a new context
         context.completed_steps = []
 
-        action = self.next_action(
-            agent=agent, input_blocks=[context.chat_history.last_user_message], context=context
-        )
+        last_user_message = context.chat_history.last_user_message
+
+        if last_user_message is None:
+            # There is nothing to respond to.
+            # TODO: Consider the situations in which an agent awakes for reasons other than message response.
+            logging.info(
+                "AgentService.run_agent invoked but no last user message existed.",
+                extra={
+                    AgentLogging.IS_MESSAGE: False,
+                    AgentLogging.MESSAGE_AUTHOR: AgentLogging.AGENT,
+                },
+            )
+            return
+
+        if last_user_message.response_sent:
+            # We've already sent a response to this message, so exit.
+            # TODO: Consider the situations in which an agent awakes for reasons other than message response.
+            logging.info(
+                "AgentService.run_agent invoked but the last user message was already responded to.",
+                extra={
+                    AgentLogging.IS_MESSAGE: False,
+                    AgentLogging.MESSAGE_AUTHOR: AgentLogging.AGENT,
+                },
+            )
+            return
+
+        # Select the first action to perform, based on the user message.
+        action = self.next_action(agent=agent, input_blocks=[last_user_message], context=context)
 
         while not isinstance(action, FinishAction):
+            # As long as the Agent hasn't decided to complete an action sequence, continue performing actions and then
+            # selecting a potential next action.
             self.run_action(agent=agent, action=action, context=context)
             action = self.next_action(agent=agent, input_blocks=action.output, context=context)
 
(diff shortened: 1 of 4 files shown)


src/steamship/agents/service/agent_service.py
```diff
@@ -131,15 +131,48 @@ def run_action(self, agent: Agent, action: Action, context: AgentContext):
                 context.action_cache.update(key=action, value=action.output)
 
     def run_agent(self, agent: Agent, context: AgentContext):
-        # first, clear any prior agent steps from set of completed steps
-        # this will allow the agent to select tools/dispatch actions based on a new context
+        """Run the agent.
+
+        This is the root method that runs when the agent awakes to do work. Any chat messages to response to should
```
douglas-reid, line 136:
```suggestion
        This is the root method that runs when the agent awakes to do work. Any chat messages to respond to should
```

src/steamship/agents/service/agent_service.py
```diff
@@ -131,15 +131,48 @@ def run_action(self, agent: Agent, action: Action, context: AgentContext):
...
+            logging.info(
+                "AgentService.run_agent invoked but no last user message existed.",
+                extra={
+                    AgentLogging.IS_MESSAGE: False,
+                    AgentLogging.MESSAGE_AUTHOR: AgentLogging.AGENT,
+                },
+            )
+            return
+
+        if last_user_message.response_sent:
```
douglas-reid, line 158: does this imply that, from an agent perspective, a single user message can ONLY have a singular response (perhaps consisting of multiple blocks)?  There is no response-in-progress type state?

src/steamship/agents/service/agent_service.py
```diff
@@ -165,6 +202,9 @@ def run_agent(self, agent: Agent, context: AgentContext):
             logging.info(f"Emitting via function: {func.__name__}")
             func(action.output, context.metadata)
 
+        # Finally, mark this block as responded to.
+        last_user_message.set_response_sent(True)
```
douglas-reid, line 206: i wonder what the implications are for transports?  If the agent think that a message has been successfully sent, but the transport fails to deliver, will we have issues?  Does the `emit_func` need a different signature to allow for that possibility?

src/steamship/data/block.py
```diff
@@ -250,6 +250,17 @@ def raw_data_url(self) -> Optional[str]:
         else:
             return None
 
+    @property
+    def response_sent(self) -> bool:
+        return get_tag_value_key(
+            self.tags, TagValueKey.BOOL_VALUE, kind=DocTag.CHAT, name=ChatTag.RESPONSE_SENT
+        )
+
+    def set_response_sent(self, is_sent: bool):
```
douglas-reid, line 259: is there any value in recording any of the following in the tag:
- recording date/time of response
- recording block IDs / hash of response 
- context ID of response
- request ID that led to response

?

src/steamship/agents/service/agent_service.py
```diff
@@ -165,6 +202,9 @@ def run_agent(self, agent: Agent, context: AgentContext):
             logging.info(f"Emitting via function: {func.__name__}")
             func(action.output, context.metadata)
 
+        # Finally, mark this block as responded to.
```
douglas-reid, line 205: wdyt: should we record a log event related to marking that a response has been sent?

src/steamship/agents/service/agent_service.py
eob:
```
"does this imply that, from an agent perspective, a single user message can ONLY have a singular response (perhaps consisting of multiple blocks)?"
```

Yeah that is the semantics this implements, except for this edge case below. Does that block a usage scenario we've got?

```
There is no response-in-progress type state?
```

Yeah there's a `delta-t` in between `webhook-received` and `last-message-marked-as-responded-to` in which the agent could always double respond. As implemented here, that `doublt-t` spans the duration of the entire response generation. We could also shorten it by having an `response-in-progress` state that is more quickly added to the message.

Do you think that's important to add here?

src/steamship/agents/service/agent_service.py
eob: Since our PackageService failure model is `throw on fail`, then a failed transport  would result in the message not getting stamped as having been delivered.

I think Step 1 for defining something more nuanced would probably be to define a different set of semantics around `emit_fn`, yeah.

src/steamship/data/block.py
eob: This is a great idea -- and trying to add these in led to a worthwhile question to bounce back:

* Block IDs aren't yet available from within `run_agent` since we request that the webhook implementor be the one to decide whether to commit the response to the ChatHistory. That means the following two scenarios are possible:
  * An input has been responded to, and that response is in the chat history
  * An input has been responded to, and that response isn't in the chat history

Would it make sense to similarly permit the webhook implementor to decide whether to mark a message as having been responded to? Basically  make calling `set_response_set` something that wasn't automatic, from within `run_agent`?. 

That feels consistent with the current semantics of `run_agent` not mutating `ChatHistory`

What do you think?

src/steamship/agents/service/agent_service.py
douglas-reid: I don't know that it is a blocker for any particular use case we have. I'd have to invent a scenario in which having multiple agents respond to a single user input is the "right" thing.

In general, though, I'm wondering if this solves the core issue that was being observed, and how that manifests.

Is the issue that the agent is being triggered to run on the _same_ block/message?  Or that a new, *identical* block/message is being added because a client is *resending* the same content?

Because if it is the first scenario, I want to understand _why_ that is happening and what the implications are for latency and in-flight state.

If it is the second, I'm not sure `response_sent` solves the issue.

eob:
Closing w/o merge because:

1) a slack-specific fix was smaller and better to achieve the goal of this pr, and
2) the comment thread above reveals that there's a good bit of nuance as to how we'd want to go about this

PR #519, opened 2023-08-16
fix(react): update code to match caching refactor of Action

douglas-reid:
`Action::tool` is now a `str` and not an obj of type `Tool`.


dkolas (approved):
👍

PR #522, opened 2023-08-18
[Proposal / RFC] Align Mixin instance_init signature with Invocable instance_init signature

eob:
## Context

We've been discussing ways to simplify the `Config` handling requirements across mixins and packages. 

One complexity identified is that configs of different types were being passed around automatically at different times.

## This PR

This PR removes all arguments from the Mixin's `instance_init` method, bringing it into alignment with PackageService.

This simplifies the contract of a Mixin's state without impacting possibility: the state would not be provided exclusively upon instantiation, by the PackageService author. 100% of the state it needs must be provided at that time.

## Mixins affected

This only affects the Telegram mixin, but it turns out that that Mixin already had -- via its `__init__` method -- all of the information that it just lost by simplifying the `instance_init` method.

## Implied "Best Practice" Mixin Configuration Pattern

This results in a situation where **we do not need to commingle MixinConfig and PackageConfig**.

The following manner of configuration can now keep the two configuration layers completely separate, ducking around multi-inheritance issues with Pydantic:

```
            TelegramTransport(
                client=self.client,
                config=TelegramTransportConfig(bot_token=self.config.bot_token),
            )
```


dkolas (approved):
Thanks!  Looks good to me

dkolas (dismissed):
🚢 :shipit:

PR #524, opened 2023-08-21
RFC: Slack thread support

GitOnUp:
Adds knowledge of threading to the Slack Transport.

In its current form, the transport will respond to the channel ID only, which means that if you respond from a thread the transport will respond outside of it.  This adds knowledge of the `thread_ts` field to blocks, which is the `ts` of the parent message in the thread, and is specified in the `chat.postMessage` method as a way to reply to it.

Some open questions about whether context ID for the bot context should be channel or (channel,thread) tuple.

I've also noticed that the Tags representing slack metadata on those blocks seems to be getting stripped on its way down to the tool, which I'm assuming is defined behavior but I haven't traced it yet.

### DIFF
--- a/src/steamship/agents/mixins/transports/slack.py
+++ b/src/steamship/agents/mixins/transports/slack.py
@@ -5,18 +5,28 @@ from typing import List, Optional
 
 import requests
 from pydantic import BaseModel, Field
-
 from steamship import Block, Steamship
+from steamship import DocTag
 from steamship.agents.llms import OpenAI
 from steamship.agents.mixins.transports.transport import Transport
 from steamship.agents.schema import EmitFunc, Metadata
 from steamship.agents.service.agent_service import AgentService
 from steamship.agents.utils import with_llm
+from steamship.data import TagValueKey
+from steamship.data.block import get_tag_value_key
 from steamship.invocable import Config, InvocableResponse, get, post
 from steamship.utils.kv_store import KeyValueStore
 
 SLACK_API_BASE = "https://slack.com/api/"
 SETTINGS_KVSTORE_KEY = "slack-transport"
+
+
+def get_block_thread_ts(block: Block) -> Optional[str]:
+    return get_tag_value_key(block.tags, TagValueKey.STRING_VALUE, kind=DocTag.CHAT, name="slack-threadts")
+
+
+def set_block_thread_ts(block: Block, thread_ts) -> None:
+    block._one_time_set_tag(tag_kind=DocTag.CHAT, tag_name="slack-threadid", string_value=thread_ts)
 
 
 class SlackElement(BaseModel):
@@ -95,6 +105,9 @@ class SlackEvent(BaseModel):
     ts: Optional[str] = Field(
         description="Timestamp of the message. A string, but is a floating point number within that."
     )
+    thread_ts: Optional[str] = Field(
+        description="Timestamp of the thread this message is a part of, if any.  Same format as `ts`."
+    )
     item: Optional[str] = Field(
         description="Data specific to the underlying object type being described."
     )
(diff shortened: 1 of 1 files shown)


src/steamship/agents/mixins/transports/slack.py
```diff
@@ -369,14 +397,22 @@ def _respond_to_block(self, incoming_message: Block):
         """Respond to a single inbound message from Slack, posting the response back to Slack."""
         try:
             chat_id = incoming_message.chat_id
-            context = self.agent_service.build_default_context(context_id=chat_id)
+            thread_ts = get_block_thread_ts(incoming_message)
+            # TODO (george) not sure this is always what we want, unfortunately.  Some bots may want to keep a
+            #  conversation going through threads, others may not?
+            context_id = chat_id if not thread_ts else f"{chat_id}-{thread_ts}"
```
GitOnUp, line 403: Possibly an advanced configuration option for this class.

src/steamship/agents/mixins/transports/slack.py
```diff
@@ -5,20 +5,30 @@
...
+from steamship.data.block import get_tag_value_key
 from steamship.invocable import Config, InvocableResponse, get, post
 from steamship.utils.kv_store import KeyValueStore
 
 SLACK_API_BASE = "https://slack.com/api/"
 SETTINGS_KVSTORE_KEY = "slack-transport"
 
 
+def get_block_thread_ts(block: Block) -> Optional[str]:
+    return get_tag_value_key(block.tags, TagValueKey.STRING_VALUE, kind=DocTag.CHAT, name="slack-threadts")
```
eob, line 25: Since a `thread-id` is something in common to many chat clients, I think it might be nice to add a `THREAD_ID = 'thread-id'` constant to `ChatTags` here:

https://github.com/steamship-core/python-client/blob/4ecd7f15a7f2ca6c5390a5d01123b484d433f797/src/steamship/data/tags/tag_constants.py#L261

Then we can just record it as a transport-agnostic `thread-id` and the `slack.py` transport is "that which converts to our global schema"

src/steamship/agents/mixins/transports/slack.py
```diff
@@ -120,8 +133,16 @@ def to_blocks(self) -> Optional[List[Block]]:
...
         for slack_block in self.blocks or []:
             for block in slack_block.to_blocks() or []:
+                # TODO (george): Having some sort of wrapper for Slack-specific Blocks might be nice
+                #   Or I guess the metadata block?
                 if self.channel:
                     block.set_chat_id(str(self.channel))
+                if self.ts:
+                    block.set_message_id(str(self.ts))
+                if self.user:
+                    block._one_time_set_tag(tag_kind=DocTag.CHAT, tag_name="userid", string_value=self.user)
```
eob, line 143: Ditto here w/r/t creating a `USER_ID` on `ChatTag`.  We might even include the transport (`slack`) for completeness while we're at it, too, so that reconstructing the data will make it clear where this message came from at the top level

a la:

```python
class ChatTag(str, Enum):
    # ship

    USER_ID = "user-id"
    USER_TRANSPORT = "user-transport"

```

src/steamship/agents/mixins/transports/slack.py
eob: Agree that this could be a nice config option. 

Just thinking out:

- If the inbound message was already in a thread, it seems Always Right to respond in-thread.
- If the inbound message was not already in a thread, maybe that's when some kind of `always_thread_response` option would dictate?

eob (changes requested):
This looks awesome! A few small requests w/r/t where you can stash some constants.

Excited to start using this in our own bots :)

src/steamship/agents/mixins/transports/slack.py
GitOnUp: Oh, that's not even what I was meaning to discuss here but that's a good point and I will chase that.

The thing I was getting at here is that `build_default_context` IIRC will be used for recalling previous messages from a file using that ID, and I wanted to consider whether, if you're talking to the agent from within a thread, the requests to the bot from the outer channel should be considered.

GitOnUp:
Still going to give this a bit of ad hoc testing before merge

eob (dismissed):
LGTM! 🚢

GitOnUp:
Sorry for churn, found a couple of issues (didn't realize config values stripped Enum-ness so comparisons failed)

PR #525, opened 2023-08-21
WIP - Stream Consumption API 

eob:
This PR explores some ways to think about stream consumption from the Python point of view.


docs/nextra/data/blocks.rst
```diff
@@ -39,3 +39,27 @@ If you want the raw data bytes of a :py:class:`Block<steamship.data.block.Block>
...
+* ``CREATED`` - The :py:class:`Block<steamship.data.block.Block>` has been created on a :py:class:`File<steamship.data.file.File>`
+* ``DATA_APPENDED`` - New data has been appended to the :py:class:`Block<steamship.data.block.Block>`
+* ``FINISHED`` - The :py:class:`Block<steamship.data.block.Block>` is fully formed and now immutable.
+
+For ``CREATED`` and ``FINISHED`` events, the payload will be the :py:class:`Block<steamship.data.block.Block>` object
+itself. For ``DATA_APPENDED`` events, the data will be the bytes of the data that has just been appended.
+
+Once the ``FINISHED`` event event has been fired, the block is considered immutable.
+Subsequent requests to stream the block will succeed, but the only event sent on the stream will be the ``FINISHED``
+event with the fully copy of the completed block.
```
dkolas, line 65: Currently at least, there's no SSE stream for a Block, it's simply a stream of the bytes of the content.  This is what I meant by it being the same as consuming `raw` but with potential waiting on the stream.

I think this is preferable, because it does not require any special handling for Blocks to consume the byte stream.

src/steamship/data/block.py
```diff
@@ -322,6 +378,14 @@ def as_llm_input(self, exclude_block_wrapper: Optional[bool] = False) -> str:
...
             return f"Block({self.id})"
 
+    def stream(self):
+        """Return a stream of BlockEvent events for this block."""
+        return self.client.stream(
+            "file/stream",
+            EventStreamRequest(),
+            expect=BlockEvent,
+        )
+
```
dkolas, line 388: Per above, I don't think this is necessary for Blocks

dkolas (commented):
I think we're well aligned on file, but may need to talk about Blocks?

eob:
Note that I haven't updated the doc RST files yet :)

dkolas (commented):
more comments

eob:
@dkolas @douglas-reid I'm going to close this PR because I think it's overcome by events. 🚢

PR #527, opened 2023-08-23
Ignore build directory

GitOnUp:
This can be created when specifying python-client as a local dependency to another local package in requirements.txt for testing.


eob:
LGTM! The failure can be ignored -- I'll push a PR to to address.

PR #528, opened 2023-08-23
Add tests & slight refactor to TelegramTransport key handling

eob:
This PR:

1. Adds a few additional tests to TelegramTransport
2. Improves TelegramTransport's handling of dynamically set keys (it already had this; this PR makes it a bit more robust)

This will enable a few nice things form a user perspective:

1. Users can re-bind an Agent via the Web UI to a new Telegram Bot
2. Users can defer binding their Agent to a Telegram Bot until after they've created an instance

Follow-on work is a PR to do for Telegram what we just did for Slack in the web UI:

image

### DIFF
--- a/src/steamship/agents/mixins/transports/slack.py
+++ b/src/steamship/agents/mixins/transports/slack.py
@@ -556,7 +556,9 @@ class SlackTransport(Transport):
     @post("set_slack_access_token")
     def set_slack_access_token(self, token: str) -> InvocableResponse[str]:
         """Set the slack access token."""
-        kv = KeyValueStore(client=self.agent_service.client, store_identifier=SETTINGS_KVSTORE_KEY)
+        kv = KeyValueStore(
+            client=self.agent_service.client, store_identifier=self.setting_store_key()
+        )
         kv.set("slack_token", {"token": token})
         return InvocableResponse(string="OK")
 
@@ -564,7 +566,9 @@ class SlackTransport(Transport):
         """Return the Slack Access token, which permits the agent to post to Slack."""
         if self.bot_token:
             return self.bot_token
-        kv = KeyValueStore(client=self.agent_service.client, store_identifier=SETTINGS_KVSTORE_KEY)
+        kv = KeyValueStore(
+            client=self.agent_service.client, store_identifier=self.setting_store_key()
+        )
         v = kv.get("slack_token")
         if not v:
             return None
@@ -578,3 +582,6 @@ class SlackTransport(Transport):
         if token is None:
             return InvocableResponse(json=False)
         return InvocableResponse(json=True)
+
+    def setting_store_key(self):
+        return f"{SETTINGS_KVSTORE_KEY}-{self.agent_service.context.invocable_instance_handle}"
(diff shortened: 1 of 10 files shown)


src/steamship/agents/mixins/transports/telegram.py
```diff
@@ -74,31 +82,38 @@ def _instance_init(self):
...
 
     @post("telegram_disconnect_webhook")
     def telegram_disconnect_webhook(self, *args, **kwargs):
         """Unsubscribe from Telegram updates."""
-        requests.get(f"{self.api_root}/deleteWebhook")
+        api_root = self.get_api_root()
+        if not api_root:
+            raise SteamshipError(
+                message="Unable to disconnect from Telegram -- perhaps your bot token isn't set?"
+            )
```
dkolas, line 100: Obviously a non-blocker, but if you put this chunk in `get_api_root()` you wouldn't have to repeat it so much.

tests/assets/packages/demo_package.py
```diff
@@ -48,6 +48,14 @@ def resp_dict(self) -> InvocableResponse[dict]:
...
         return InvocableResponse(json=TestObj(name="Foo"))
 
+    @post("resp_true")
+    def resp_true(self) -> InvocableResponse[dict]:
+        return InvocableResponse(json=True)
+
+    @post("resp_false")
+    def resp_false(self) -> InvocableResponse[dict]:
+        return InvocableResponse(json=False)
+
```
dkolas, line 58: Was this commit intended for this PR?

dkolas (approved):
Looks good; I think this will help users be able to fix their existing bots without creating 1000 instances. 

Minor request for a rename if you're up for it.

tests/assets/packages/demo_package.py
eob: Yeah I added it while trying to debug, and then left it because `false` tends to be a really hard thing to represent in a multi-platform system.. so as silly as it sounds, it feels very comforting to know we have a test that tests for our ability to "return false" across Python -> Swift -> HTTP

src/steamship/agents/mixins/transports/slack.py
```diff
@@ -578,3 +582,6 @@ def is_slack_token_set(self) -> InvocableResponse[bool]:
         if token is None:
             return InvocableResponse(json=False)
         return InvocableResponse(json=True)
+
+    def setting_store_key(self):
+        return f"{SETTINGS_KVSTORE_KEY}-{self.agent_service.context.invocable_instance_handle}"
```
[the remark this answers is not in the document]
eob: Yeah it's intentional.. but I wasn't 100% confident about it. What do you think?

Here's a bug I ran into while testing:

- The prior naming strategy meant two agents in the same workspace would share the same settings file.

This modification ensures that the transport settings are specific to the agent, not the workspace.


As a complete aside: I'm starting to winder if some sort of "DynamicSettings" object might be a powerful compliment to the immutable settings object, which is also valuable.

E.g. a single thing that contains transport settings, preferred default LLM, temprerature, etc.

PR #529, opened 2023-08-24
[SHIP-832] Request ID python plumbing

dkolas:
Receive a requestId in lambda handler and attach it to the client object.

This requestId is then passed in all engine api calls to create traceability throughout the system.

Relies on PR: https://github.com/nludb/nludb/pull/621, so will not pass tests until that is deployed.


tests/assets/packages/request_id_plumbing_test_package.py
```diff
@@ -0,0 +1,37 @@
...
+
+        self_task_fetched = Task.get(
+            self.client, _id=self_task.task_id
+        )  # to undo the wrong generic Task output
+        invoke_later_request_id_dict = json.loads(
+            str(base64.b64decode(self_task_fetched.wait()), encoding="UTF-8")
+        )  # invoke_later tasks get stored b64 encoded
+        return InvocableResponse(
+            json={
+                "packageRequestId": self.client.config.request_id,
```
eob, line 29: This is a clever testing strategy

tests/steamship_tests/app/integration/test_request_id.py
```diff
@@ -0,0 +1,25 @@
...
+    request_id_plugin_path = PLUGINS_PATH / "generators" / "request_id_generator.py"
+
+    with deploy_package(client, request_id_package_path) as (package, version, instance):
+        with deploy_plugin(client, request_id_plugin_path, plugin_type="generator") as (
+            plugin,
+            _,
+            _,
+        ):
+            result = instance.invoke("requestids", plugin_handle=plugin.handle)
+            package_request_id = result.get("packageRequestId")
```
eob, line 20: Not necessary in this PR, but an awesome addition would be to have the Proxy stamp a header on all requests that contained the requestId.

Then in addition to `assert package_requestId is not None` we could check against the Proxy's opinion of events.

I guess that would require us to use a raw HTTP post instead of `instance.invoke` to test though

eob (approved):
LGTM!

One small idea for a powerup, and it looks like a test is failing, but the design & testing strategy are awesome

src/steamship/base/configuration.py
```diff
@@ -45,6 +45,9 @@ class Configuration(CamelModel):
     workspace_handle: str = None
     profile: Optional[str] = None
 
+    # For use in deployed packages and plugins for tracing. Do not set manually
+    request_id: Optional[str] = None
```
GitOnUp, line 49: Not blocking: `Do not set manually` is a surprising thing to read without being as ingrained in the context here yet; it also feels at first blush to me like this is Request-Scoped rather than Client-Scoped.

e.g. in the `RequestIdGenerator` in tests the PluginRequest seems like a natural place this would be, but I'm blanking on a request-scoped place to do this in e.g. a PackageService.

GitOnUp (approved):
One point of discussion here but LGTM

src/steamship/base/configuration.py
dkolas: Yeah, you're right of course...

The reason I tucked it into the client config object specifically and not the `InvocationContext` was that the client object would need it to send back with every single request to the Engine.  So putting it there allowed me to do it without the package or plugin author ever having to know it existed.

The comment, then, is to say "hey, don't mess with this" since AFAIK there isn't a particularly good Pythonic way of actively preventing someone from changing 'internal' member vars.

dkolas:
> LGTM!
> 
> One small idea for a powerup, and it looks like a test is failing, but the design & testing strategy are awesome

Good idea! 

Test won't pass until the engine PR is deployed :)

tests/steamship_tests/app/integration/test_request_id.py
dkolas: This is a good idea; should be very easy to add to the engine PR

PR #530, opened 2023-08-24
Fail tests if plugin/package deployment fails init

eob:
`wait_for_init` only checks to see if the initializing phase completes -- not that it succeeded.

This adds an assertion to `deploy_package` and `deploy_plugin` that will more explicitly check that the initialization succeeded to help us weed through some of the random deployment errors that happen in test batteries.


GitOnUp (commented):
🚢

PR #531, opened 2023-08-26
Tag ChatHistory blocks and files 

eob:
We don't currently Tag ChatHistory files as being ChatHistory files. That means that the Web UI can't show a list of prior chats and let the user resume or browse one.

This PR adds a tag constant for ChatHistory and tags new ChatHistory files with it when they are created.

### Along for the ride on this PR

- Remove unused code from `experimental`
- Consolidate duplicated `get_tag_value_key` methods into a single `tag_util.py` file.
- Extract `build_chat_history_for_tool` function from `next_action` to enable testing on its own and possible resuse


src/steamship/agents/llms/openai.py
```diff
@@ -80,7 +82,14 @@ def chat(self, messages: List[Block], tools: Optional[List[Tool]], **kwargs) ->
...
 
-        temp_file = File.create(client=self.client, blocks=messages)
+        temp_file = File.create(
+            client=self.client,
+            blocks=messages,
+            tags=[
+                # Mark that this is a CHAT COMPLETION file
+                Tag(kind=TagKind.CHAT, name=ChatTag.COMPLETION)
+            ],
+        )
```
dkolas, line 93: Why mark this 'temp_file' as a chat completion?  If anything is a completion, shouldn't it be the generated block? 

Seems like maybe you want a new tag kind for this file.

dkolas (dismissed):
Looks good - one suggestion for a new tag kind

src/steamship/agents/llms/openai.py
eob: Just switched to:

```
  Tag(kind=TagKind.GENERATION, name=GenerationTag.PROMPT_COMPLETION)
```

PR #532, opened 2023-08-27
RFC: Add `return_direct` to Tools, Actions, and FunctionCalling Agents

eob:
This PR adds a `return_direct` flag to `Tool` [the style of LangChain](https://github.com/langchain-ai/langchain/blob/0d01cede03fbd1acc7fd3f31870ffec0b7880e3d/libs/langchain/langchain/tools/base.py#L137C13-L137C13) that permits a tool to signal that its response should be delivered as the "final answer" to the user.

This prevents a common error condition in which a reasoning agent runs a "terminal" tool, but then instead of returning its output, it either:

- Summarizes for the user what it did
- Runs other tools unnecessarily

Here's an example of that in action with the Prompt Golf bot:

image

What happened here is:

- The Agent correctly ran the `GameTurnTool` which generated a new image
- The Agent then replied with a summary of what it did rather than replying with the output of the tool

image


eob:
@douglas-reid does this seem like a reasonable way to achieve this? It would result in:

- FinishAction being the way the Agent signals completion.
- Tool.return_direct being a way a Tool can signal that it always should end reasoning

douglas-reid:
@eob my first reaction is that this is fine, as it seems to address an immediate need. it _feels_ like "llm smell" in the sense that a summarizing action seems like something that should be ideally smoothed out of workflows at a different level. and i worry that there are likely tools that don't always want to return `FinishAction`s. but both of those concerns seem minimal in the face of our existing reality. i'll take a closer look at the code impl now.

src/steamship/agents/schema/action.py
```diff
@@ -22,6 +22,12 @@ class Action(BaseModel):
     output: Optional[List[Block]]
     """Any direct output produced by the Tool."""
 
+    return_direct: bool = False
```
douglas-reid, line 25: curious: should we label this more like  `is_final` or `tool_is_direct` similar to track that state (and update the comment -- as i'm not expecting users to set `return_direct` on `Action`s) ?

aside: Would this be "simpler" if the `Action` returned by an `Agent`/`AgentExecutor` when  `Tool.return_direct == True` was a `FinishAction` instead?  Or even a new type of `Action` (ex: `DirectReturnAction`) ?  Maybe that would be more complex (as the `AgentExecutor` would have to be involved) ??

src/steamship/agents/schema/tool.py
```diff
@@ -40,6 +40,12 @@ class Tool(BaseModel):
     """Human-friendly description.
     Used for logging, tool indices, etc."""
 
+    return_direct: bool = False
+    """Whether to return the tool's output directly.
+
+    Setting this to True means that after the tool is called, the AgentExecutor will stop looping.
```
douglas-reid, line 46:
```suggestion
    Setting this to True means that after the Tool is run, the Tool's output will be returned without further Agent interactions. This will preempt the execution of other tools, as well as any subsequent calls to LLMs, etc., to interpret the Tool output.
```

src/steamship/agents/service/agent_service.py
```diff
@@ -146,6 +148,20 @@ def run_agent(self, agent: Agent, context: AgentContext):
 
         while not isinstance(action, FinishAction):
             self.run_action(agent=agent, action=action, context=context)
+
+            # If this action wishes to end the loop, comply with that wish
+            if action.return_direct:
```
douglas-reid, line 153: aside (take it or leave it): my api spidey-sense is tingling seeing this code block. i think seeing this, i'd prefer something like `action.is_final` (or even `if isinstance(action, DirectReturnAction)`) for clarity (as `Action`s aren't really "returning" anything per se).

src/steamship/agents/service/agent_service.py
```diff
@@ -146,6 +148,20 @@ def run_agent(self, agent: Agent, context: AgentContext):
...
+                logging.info(
```
douglas-reid, line 154: curious: do you think we need this log statement for full coverage?  WDYT about just adding "is_final" or similar to the existing "next tool" log?  that might lead to more readable code (no floating `break`) and fewer lines of logging code?

src/steamship/agents/service/agent_service.py
```diff
@@ -146,6 +148,20 @@ def run_agent(self, agent: Agent, context: AgentContext):
...
+                    f"Exciting reasoning loop by request of: {action.tool}",
```
douglas-reid, line 155:
```suggestion
                    f"Exiting Agent execution by request of: {action.tool}",
```

douglas-reid (commented):
Some initial reactions.

eob:
@douglas-reid thanks a ton -- will do some tinkering down these threads. 

A few thoughts I had that are in the same vein as yours:

- I'm not sure if it should feel weird that a the `is_final` property of `Tool` is static as opposed to part of its response.
- I like the idea of `Action.is_final`... that feels cleaner and also aligns it with `FinalAction`

PR #535, opened 2023-08-29
Add a `max_actions_per_run` variable on AgentService as a backstop

eob:
An observed failure mode of a reasoning Agent is when it spirals into endlessly calling tools. 

In a development loop, the programmer can quickly break the loop, but in a hosted situation this can result in surprising costs if each tool involves LLM usage.

This PR adds a simple backstop in which the `max_actions_per_run` variable on AgentService (default = 5) acts as a limit on how many actions the agent is permitted to run before being stopped.

If the AgentService attempts to run an action after `max_actions_per_run` has been reached, it is treated as an error which will be returned to the chat invoker.


eob:
Note: 

On staging, the failed telegram test is fixed by https://github.com/steamship-core/python-client/pull/531
And the failed OpenAI test appears to be an error on their side of things.

douglas-reid (approved):
This seems like a fine solution. I wonder if we'll see 5 be considered too few however. We should at least two for the simplest Tool invocation (tool + finish action).

PR #537, opened 2023-08-29
feat(streaming): add logger to include status-based messages in chat history file

douglas-reid:
This PR provides an approach for providing "status"-like events inline in the same "stream" as other File-based (or in the case of agents, History-based) events. 

This PR uses a `StreamHandler` to capture existing logs and redirect them into the `ChatHistory` as blocks tagged with either a role of `Agent` or `Tool` or `LLM`. Those messages are then explicitly ignored when providing blocks for conversation history in LLM interactions (in an attempt to prevent contamination).

The `StreamHandler` is added/removed on a request-by-request basis by adding context manager capabilities to `AgentContext`.

Here is a **BEFORE** snapshot of an agent history (without status-like messages in stream):

```
----- Agent Chat History 6CEBF384-381D-40DE-9746-9A47D7F52138 -----

[user] draw a cat with a funny hat
[assistant] Here is the image you requested: 
[assistant] FD36377D-8C86-4CEC-A259-40FEAD220847 (image/png)
[user] who is the president of Taiwan?
[assistant] The current president of Taiwan is Tsai Ing-wen.
[user] draw them at the podium
[assistant] Here is the image you requested: 
[assistant] 7BEE023D-112F-4FBE-AF43-A6440A0A0FCB (image/png)
```

Here is the **AFTER**:

```
----- Agent Chat History 51A59B73-A398-4D53-8CEC-FE9280D5C043 -----

[user] draw a cat with a funny hat
[agent] Selecting next action...
[llm] OpenAI ChatComplete (draw a cat with a funny hat)
[agent] Selected next action: DalleTool
[agent] Running Tool DalleTool (a cat with a funny hat)
[agent] Tool DalleTool: (Block(8691E180-8216-46DE-B1B4-997429236BDA))
[agent] Selecting next action...
[llm] OpenAI ChatComplete (Block(8691E180-8216-46DE-B1B4-997429236BDA))
[agent] Selected next action: Agent-FinishAction
[agent] Next Tool: Agent-FinishAction
[assistant] Here is the image you requested: 
[assistant] 4D557AD0-FA0E-40F9-BB90-6EF95B898CB8 (image/png)
[user] who is the president of Taiwan?
[agent] Selecting next action...
[llm] OpenAI ChatComplete (who is the president of Taiwan?)
[agent] Selected next action: SearchTool
[agent] Running Tool SearchTool (current president of Taiwan)
[tool] Executing search: current president of Taiwan
[agent] Tool SearchTool: (Tsai Ing-wen)
[agent] Selecting next action...
[llm] OpenAI ChatComplete (Tsai Ing-wen)
[agent] Selected next action: Agent-FinishAction
[agent] Next Tool: Agent-FinishAction
[assistant] The current president of Taiwan is Tsai Ing-wen.
[user] draw them at the podium
[agent] Selecting next action...
[llm] OpenAI ChatComplete (draw them at the podium)
[agent] Selected next action: DalleTool
[agent] Running Tool DalleTool (Tsai Ing-wen, the president of Taiwan, standing at the podium)
[agent] Tool DalleTool: (Block(E0BA53EA-09E8-41F0-8248-380799BD95BD))
[agent] Selecting next action...
[llm] OpenAI ChatComplete (Block(E0BA53EA-09E8-41F0-8248-380799BD95BD))
[agent] Selected next action: Agent-FinishAction
[agent] Next Tool: Agent-FinishAction
[assistant] Here is the image you requested: 
[assistant] EC6B831C-754A-4DEC-9E63-03FD7BC9A4F9 (image/png)

```


douglas-reid:
There are some considerations here to add:
- this could dramatically increase the number of blocks in the history file (and require more time to scan/skip)
- this duplicates logs stream information.

eob:
Comments just based on the description (which is really helpful, thanks) before diving into code:
 
**Persona: Front End**

* This checks the box of what would produce a Chat-GPT style interface. The main thing here being that the front-end agent is provided with information to meaningfully keep the user engaged while long-latency delays prevent streaming the actual response ("Running tool..", "Ran tool...")

* I don't think, from the front-end perspective, it's 100% necessary to know the input/output of intermediate actions. It would be probably be fine just to know that they were being run and what their eventual status was. This isn't a change request, just an observation about what could be trimmed if the size of ChatHistory is a concern.

**Persona: Performance Engineer**

* Very much read you on the concern for ChatHistory bloat. Thinking out loud:
  * If we trimmed down the ChatHistory metadata blocks to just which actions ran and their status, and eliminated the FinishAction reporting as sort of a given, that might be a meaningful diet to put the reporting on.
  * If we had a AgentContextHistory file that sat alongside ChatHistory that contained a fuller picture of the ChatHistory, that could be a way to keep a parallel set of books that was more detailed, but by default not consumed. ChatHistory & ContextHistory could be kept 1:1
  
**Persona: Agent SDK Design**

* The logging StreamHandler trick feels really clever, but also maybe if we went that route, we'd probably want to make a point of being incredibly explicit (as you allude to above) with rigid expectations upon logging style. Maybe helper functions that pass down to logging operations?
* The logging duplication problem does seem challenging.. and the AgentContextHistory idea above exacerbates it. Is AgentContextHistory really just a stream over ElasticSearch with a filter applied?


Will keep thinking... let me know what's useful/not useful feedback wise, but this is really awesome to read through so far 🚀🚀🚀🚀🚢🚢🚢🚢🚢

dkolas:
I am a big fan of this approach generally. I think having all of the inner 'reasoning' steps in the history of the chat makes total sense, and then having them appropriately filtered back out based on tags etc should work really well.

I am not particularly concerned about chat history bloat, personally.  I think that can be handled via nice filter options.

dkolas (dismissed):
This looks good to me! The approach makes sense, and nothing jumped out at me in the code.

src/steamship/agents/logging.py
```diff
@@ -37,3 +42,34 @@ class AgentLogging:
...
+
+LOGGING_FORMAT = {
+    "level": "%(levelname)s",
+    "host": "%(hostname)s",
+    "where": "%(module)s.%(filename)s.%(funcName)s:%(lineno)s",
+    "type": "%(levelname)s",
+    "stack_trace": "%(exc_text)s",
+    "message_type": "%(message_type)s",
+    "component_name": "%(component_name)s",
+    AgentLogging.IS_MESSAGE: f"%({AgentLogging.IS_MESSAGE})s",  # b doesn't work. Unsure how to make a bool
```
eob, line 56: Naieve response is the way you probably were avoiding because it felt silly: `'true' if AgentLogging.IS_MESSAGE else 'false'`

src/steamship/agents/logging.py
```diff
@@ -37,3 +42,34 @@ class AgentLogging:
...
+    AgentLogging.AGENT_NAME: f"%({AgentLogging.AGENT_NAME})s",
+    AgentLogging.MESSAGE_AUTHOR: f"%({AgentLogging.MESSAGE_AUTHOR})s",
+    AgentLogging.MESSAGE_TYPE: f"%({AgentLogging.MESSAGE_TYPE})s",
+    AgentLogging.TOOL_NAME: f"%({AgentLogging.TOOL_NAME})s",
+}
+
+
+class StreamingOpts(BaseModel):
```
eob, line 64: I like how you've started with the idea that what gets added to the stream can be selected.

One thing that would help me (and therefor I think others?) is to actively add comments here about what distinguishes an `agent` and `llm` and `tool` message.

For me: I find myself wanting to be reassured about my guess about `agent` versus `llm` specifically.

For ex:
- Is `agent_message` something the agent says that gets added to the chat history?
- Is `llm_message` something that the LLM emits to a tool or agent, but doesn't tend to get added to the chat history?

src/steamship/agents/schema/agent.py
```diff
@@ -49,14 +50,13 @@ def messages_to_prompt_history(messages: List[Block]) -> str:
         as_strings = []
         for block in messages:
             role = block.chat_role
+            # DON'T RETURN AGENT MESSAGES -- THOSE ARE INTERNAL STATUS MESSAGES
```
eob, line 53: Oh interesting. It's not what I guessed above. Yeah little doc notes would def be helpful then :)

src/steamship/agents/schema/chathistory.py
```diff
@@ -272,3 +278,120 @@ def clear(self):
...
+        tags: List[Tag] = None,
+        content: Union[str, bytes] = None,
+        url: Optional[str] = None,
+        mime_type: Optional[MimeTypes] = None,
+    ) -> Block:
+        """Append a new block to this with status update messages from the Agent."""
+        return self.append_message_with_role(text, RoleTag.LLM, tags, content, url, mime_type)
+
+
+class ChatHistoryLoggingHandler(StreamHandler):
```
eob, line 316: I think it's really cool that we're hooking on to the logging system in this way.

src/steamship/agents/schema/chathistory.py
```diff
@@ -272,3 +278,120 @@ def clear(self):
...
+            self.streaming_opts = streaming_opts
+        else:
+            self.streaming_opts = StreamingOpts()
+
+    def emit(self, record):
+        if record.levelno < self.log_level:
+            # don't bother doing anything if level is below logging level
+            return
+
+        message_dict = cast(dict, self.format(record))
```
eob, line 347: [not high pri; feel free to ignore] 

below feels like an opportunity to collapse into one block of code below since so much is repeated.

src/steamship/agents/service/agent_service.py
```diff
@@ -272,11 +268,28 @@ def build_default_context(self, context_id: Optional[str] = None, **kwargs) -> A
         if runtime_use_action_cache := kwargs.get("use_action_cache"):
             use_action_cache = runtime_use_action_cache
 
+        include_agent_messages = True
```
eob, line 271: I think, but am not 100% confident, that you could alternatively say:

```
include_agent_messages = kwargs.get("include_agent_messages", True)
```

src/steamship/agents/service/agent_service.py
```diff
@@ -288,54 +301,48 @@ def build_default_context(self, context_id: Optional[str] = None, **kwargs) -> A
...
-            # Need to make the output blocks public here so that they can be copied to the chat history.
-            # They generally need to be public anyway for the REPL to be able to show a clickable link.
-            output_block.set_public_data(True)
-            context.chat_history.append_assistant_message(
-                text=output_block.text,
-                tags=output_block.tags,
-                url=output_block.raw_data_url or output_block.url or output_block.content_url,
-                mime_type=output_block.mime_type,
-            )
+        with self.build_default_context(context_id, **kwargs) as context:
```
eob, line 311: +1. nice

eob (approved):
Sprinkled comments but mostly just talking out loud to myself while reading. This looks great!!

LGTM

src/steamship/agents/logging.py
douglas-reid: Added some docs. One thing that I wanted to make more distinct in this PR is the boundary between `Agent` and `LLM`. There is nothing in `Agent` that requires `LLM` integration or backing. You could have a purely random tool selection agent, for instance. So, I wanted to allow for `LLM`s to be responsible for `LLM`-scoped messages and `Agent`s to focus on the level above in logs. This might come at a cost of a bit more of verbosity, however.

src/steamship/agents/schema/chathistory.py
douglas-reid: refactored a bit. the slight diffs between the various modes makes a full collapse uglier than a small bit of near repetition (imho), but the refactor hopefully makes things clearer/more readable.

PR #413, opened 2023-06-05
feat(agents): Chat Memory (riff 2)

dkolas:
Based on #404 - with a couple differences:

- Renamed filtering as strategies rather than memory so it is clear it is not where the data is stored.
- Inverted the calling relationship between ChatHistory and (ConversationMemory -> MemoryWindowStrategy)

PR #538, opened 2023-08-29
fix(agents): use proper init names for message selectors

douglas-reid:
`conversation_memory` was renamed/used in https://github.com/steamship-core/python-client/pull/413, but the examples were never updated.


eob (approved):
Nice catch. (Raises fist and says "kwargs!!" in George Costanza voice)

LGTM!

PR #540, opened 2023-08-30
fix: CI/CD tests sometimes fail if deployment takes longer than default Task timeout

eob:
This PR introduces a `wait_forever` method on Task, and then uses it for Package & Plugin deployments within tests.

The hope is that some of the occasional deployment timeout failures can be avoided by simply extending the wait time.

(In response to a PR earlier today in which a single failure related to deployment timeouts happened.)

### DIFF
--- a/src/steamship/base/tasks.py
+++ b/src/steamship/base/tasks.py
@@ -244,6 +244,7 @@ class Task(GenericCamelModel, Generic[T]):
         ----------
         max_timeout_s : int
             Max timeout in seconds. Default: 180s. After this timeout, an exception will be thrown.
+            A timeout of -1 is equivalent to no timeout.
         retry_delay_s : float
             Delay between status checks. Default: 1s.
         on_each_refresh : Optional[Callable[[int, float, Task], None]]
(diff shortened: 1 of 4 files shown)


douglas-reid (approved):

douglas-reid:
@eob do you think this is an indication that we should add something like either: `task.wait_forever()` or `task.until_succeed_or_failed()` ?

eob:
@douglas-reid I sat on it for the day to see if I had an opinion, and I think I'm completely agnostic.

It's just a few LoC change -- LMK if you think that's a better route and I'll update this PR to be that instead.

If yes, my vote between the two would be  `wait_forever` with the implicit semantics of throwing on failure or returning on success.

douglas-reid:
> @douglas-reid I sat on it for the day to see if I had an opinion, and I think I'm completely agnostic.
> 
> It's just a few LoC change -- LMK if you think that's a better route and I'll update this PR to be that instead.
> 
> If yes, my vote between the two would be `wait_forever` with the implicit semantics of throwing on failure or returning on success.

No worries. I've just run across this before personally, and always end up writing something like:

```python
   state = task.state
   while state not in [TaskState.succeeded, TaskState.failed]:
        try:
            task.wait()
            state = task.state
        except:
           ...
```

or whatever and have _thought_ about adding a shorthand for that if others started stumbling too.

douglas-reid (approved):

PR #541, opened 2023-08-31
feat(tools): support non-caching tools

douglas-reid:
When turning on an action cache for an `AgentService`, there may be certain tools that developers may never want to cache the results of and skip execution for. This PR introduces a mechanism for opting Tools out of caching, when desired.


eob (approved):
💪 LGTM! Nice

PR #544, opened 2023-09-07
feat: Adds Lamma/Replicate to our Agent LLM stack

eob:
This PR wraps the soon-to-launch Replicate Generator Plugin with an LLM abstraction for Agent & Tool use.


src/steamship/agents/llms/replicate.py
```diff
@@ -0,0 +1,79 @@
...
+      https://github.com/steamship-plugins/replicate-llms
+    """
+
+    generator: PluginInstance
+    client: Steamship
+
+    def __init__(
+        self,
+        client,
+        model_name: str = LLAMA_2_70B,
```
GitOnUp, line 30: The plugin provides its own default here, so it's not strictly necessary to provide one (although whether the plugin should be opinionated about defaults rather than the agent code is up for debate)

ETA: replicate LLMs provide their own defaults across the board for parameters like max_tokens and the like as well, though if we've previously been providing e.g. 0.4 for temperatures that could result in a change in behavior for people moving over -- replicate's default for that field is .75.

src/steamship/agents/llms/replicate.py
```diff
@@ -0,0 +1,79 @@
...
+         - replicate/dolly-v2-12b
+        """
+
+        # TODO: We've got whatever the opposite of an encapsulation leak here, in that this is all strongly typed
+        # and defined in the Replicate plugin, yet we can't refer to it here.
+        #
+        # One side-step approach might be to strongly type LLM configuration in the python-client repository, since
+        # it tends to be mostly stable across LLM providers, and then we can refer to it from plugins, inverting the
+        # location of definition.
+        replicate_config = {
```
GitOnUp, line 53: Related to the above, if you leave out values for defaultable entries here they'll use model defaults.

src/steamship/agents/llms/replicate.py
```diff
@@ -0,0 +1,79 @@
...
+            "stop_sequences": stop_sequences,
+        }
+
+        generator = client.use_plugin(PLUGIN_HANDLE, replicate_config)
+        super().__init__(client=client, generator=generator, **kwargs)
+
+    def complete(self, prompt: str, **kwargs) -> List[Block]:
+        """Completes the prompt.
+
+        At present, no generate-time configuration overrides are supported.
```
GitOnUp, line 69: I believe they are, via the `options` kwarg to `generate()`.

https://github.com/steamship-core/python-client/blob/main/src/steamship/data/plugin/plugin_instance.py#L132
https://github.com/steamship-plugins/replicate-llms/blob/impl/src/versions/common.py#L22C23-L22C23

Doesn't `options` become `PluginRequest.data.options`?

src/steamship/agents/llms/replicate.py
```diff
@@ -0,0 +1,79 @@
...
+
+        Valid model names are presently:
+         - replicate/llama-2-70b-chat
+         - replicate/dolly-v2-12b
+        """
+
+        # TODO: We've got whatever the opposite of an encapsulation leak here, in that this is all strongly typed
+        # and defined in the Replicate plugin, yet we can't refer to it here.
+        #
+        # One side-step approach might be to strongly type LLM configuration in the python-client repository, since
```
GitOnUp, line 50: If configuration moves out of this, I'm thinking `__init__` should just take that.

I also don't know if I would call this a strictly side-step approach rather than a normal pattern, publishing an 'interfaces' lib (or bundling common interfaces in a base lib) is an established flow in many cases.

In the case that those configs are just passed into the `__init__`, it might even be possible to just have one general LLM and you provide a config for the one you want to use, since the code otherwise is pretty generic.

GitOnUp (commented):
Would like to discuss theory around defaults quickly

src/steamship/agents/llms/replicate.py
eob: @GitOnUp re: our conversation about {"key": None} versus {}, I'll update this to only include the non-none values as a double-check.

src/steamship/agents/llms/replicate.py
eob: Oh! Fantastic! I'll add them then. When I read your PR I didn't realize you'd supported it yet. Will add.

src/steamship/agents/examples/example_chatbot_with_replicate.py
```diff
@@ -0,0 +1,36 @@
+from steamship.agents.basic_chat import BasicChatAgent
+from steamship.agents.llms.replicate import ChatReplicate
+from steamship.agents.schema.message_selectors import MessageWindowMessageSelector
+from steamship.agents.service.agent_service import AgentService
+from steamship.utils.repl import AgentREPL
+
+
+class MyAssistant(AgentService):
```
GitOnUp, line 8: I don't want to derail this PR but would this fit better in the MMAS?

I think there was a recent PR where there was discussion about where examples should live in general but I'd have to dig it up.

src/steamship/agents/llms/replicate.py
```diff
@@ -0,0 +1,150 @@
...
+        stop_sequences: Optional[str] = None,  # Should be a CSV string
+        *args,
+        **kwargs
+    ):
+        """Create a new instance.
+
+        Valid model names are presently:
+         - replicate/llama-2-70b-chat
+         - replicate/dolly-v2-12b
+        """
```
GitOnUp, line 47:
```suggestion
        a16z-infra/llama-2-13b-chat
        """
```

GitOnUp: It probably would make sense to have an endpoint on the plugin that could report these rather than having to remember to come back here to update these, come to think of it.

src/steamship/agents/llms/replicate.py
```diff
@@ -0,0 +1,150 @@
...
+
+        replicate_config = remove_none(
```
GitOnUp, line 105: I don't think you need to do the interim config here; passing the params along to `super().__init__` should be enough since it also removes none there.

src/steamship/agents/llms/replicate.py
```diff
@@ -0,0 +1,150 @@
...
+        stop_sequences: Optional[str] = None,  # Should be a CSV string
+        *args,
+        **kwargs
+    ):
+        """Create a new instance.
+
+        Valid model names are presently:
+         - replicate/llama-2-70b-chat
+         - replicate/dolly-v2-12b
+        """
```
GitOnUp, line 103:
```suggestion
        a16z-infra/llama-2-13b-chat
        """
```

GitOnUp (approved):
LGTM with only minor suggestions :shipit:

src/steamship/agents/llms/replicate.py
douglas-reid: I haven't been following the evolution of that plugin, but wasn't the idea that *any* replicate LLM could be used? What is the reasoning behind the limited selection here?

src/steamship/agents/llms/replicate.py
```diff
@@ -0,0 +1,150 @@
...
+            blocks=messages,
+            tags=[Tag(kind=TagKind.GENERATION, name=GenerationTag.PROMPT_COMPLETION)],
+        )
+
+        if len(tools) > 0:
+            functions = []
+            for tool in tools:
+                functions.append(tool.as_openai_function())
+            options["functions"] = functions
+
```
douglas-reid, line 146: This whole block seems suspect to me. Do these models even support openai functions?

src/steamship/agents/llms/replicate.py
```diff
@@ -0,0 +1,150 @@
...
+                "stop_sequences": kwargs.get("stop_sequences", None),
+            }
+        )
+        action_task = self.generator.generate(text=prompt, options=options)
+        action_task.wait()
+        return action_task.output.blocks
+
+
+class ChatReplicate(ChatLLM, Replicate):
+    """ChatLLM that uses Steamship's Replicate plugin to generate chat completions."""
```
douglas-reid, line 83: Do any of the replicate LLMs follow the `Chat` model?  It looks like the llama API does, but with special formatting within a single prompt, but Dolly does not.

Is that difference handled on the plugin side somehow?  Or does that need to be handled in the `chat()` code here?

Or should Replicate LLMs not be "ChatLLM"s in our existing hierarchy?

src/steamship/agents/llms/replicate.py
```diff
@@ -0,0 +1,150 @@
...
+from steamship.data import GenerationTag, TagKind
+from steamship.utils.dict_utils import remove_none
+
+PLUGIN_HANDLE = "replicate-llm"
+DEFAULT_MAX_TOKENS = 256
+
+LLAMA_2_70B = "replicate/llama-2-70b-chat"
+DOLLY_2_12B = "replicate/dolly-v2-12b"
+
+DEFAULT_MODEL = LLAMA_2_70B
```
douglas-reid, line 14: Are these used anywhere?  Should they even exist in the wrapper code?

src/steamship/agents/examples/example_chatbot_with_replicate.py
douglas-reid: We, at the very least, need an example `AgentService` that we are testing in this codebase.

Which raises the question of "where are the tests" for this code?

src/steamship/agents/basic_chat/basic_chat_agent.py
```diff
@@ -0,0 +1,94 @@
...
+from steamship.agents.schema import Action, AgentContext, ChatAgent, ChatLLM
+from steamship.data.tags.tag_constants import RoleTag
+
+DEFAULT_PROMPT = """You are a helpful AI assistant.
+
+You chat with users are eager to engage on a variety of topics. You answer their questions, provide help thinking
+through challenges they have, and engage as a trusted assistant."""
+
+
+class BasicChatAgent(ChatAgent):
```
douglas-reid, line 14: some small concerns / things to ponder:

- is `ChatAgent` vs. `LLMAgent` (still) a useful distinction?  The lines seem incredibly blurred, and that may impact our agility here.
- is `Chat` overloaded?  the word, as used in our SDK has an backend api-specific meaning. if you want to use `Chat` to signal to our users that they can have a conversation with this LLM, then I think we need a different set of naming conventions, etc.

src/steamship/agents/basic_chat/basic_chat_agent.py
```diff
@@ -0,0 +1,94 @@
...
+    """BasicChatAgent implements a conversational agent with no function calling or tool reasoning.
```
douglas-reid, line 15: I think this documentation is misleading and will get us into trouble.  `ChatAgent`s in our codebase are for selection actions. They are not for carrying on conversations (directly).

I fear that if what you want is *direct* LLM responses, trying to bend `AgentService` to fit that need _may_ be problematic.

Should we have a `ConversationService` as a parallel concept?  Something that doesn't have the indirection of `run_agent` and `run_action` just to issue a call to an LLM and stream back a response?

douglas-reid (changes requested):

eob:
@douglas-reid @GitOnUp thanks a ton for all the conversation internally about the best way to do this. On the plane today I had a few realizations that I coded up... and I **think** this is actually in a really good state now.

Here were the realizations:

* This can't be a **Replicate** PR. This is a **Llama** PR -- or perhaps a **Llama+Replicate** PR. The reason is that every LLM is meaningfully different from the perspective of the application developer. Some support functions and some don't. Some support chat and some don't. Those that support chat require their own chat formatting.
* The `LLM` and `ChatLLM` actually seem just fine -- even though I know that @douglas-reid had concern that those were written specifically for OpenAI. While we might refactor later to remove them, they fit quite nicely into `Llama` and `ChatLlama` exactly as the Llama builders intended.
* The fact that some LLMs can't support tools is inevitable -- we can't get around it. So we'll just raise an error if you try to shove tools into an LLM that doesn't work well with them. 

With those realizations in mind, I did a simplification of this PR that amounts to basically copy-pasting the `openai.py` file and customizing it for `llama + replicate`. There are a few differences:

- It will throw if you give the chat variant tools
- It uses the LLama specific chat formatting (instead of structured configuration)

But there are also some great similarities:

- The chat variant works with a system prompt
- That system prompt, if you wanted it to be, could be the React system prompt for use with a react agent.

And this is a minimum-code addition that adds value without preventing a refactor:

- If we refactor in the style in the Notion doc, it would just be to remove base classes and wrap the prompt formatting function in a common interface.

Anyway... this feels pretty solid to me so wanted to bounce it back to you all :)

src/steamship/agents/basic_chat/basic_chat_agent.py
eob: I think the `ConversationService` route is a perfectly fine idea, but since the rest of our codebase (e.g. transports -- which the UI is also built around) depends upon AgentService so it would involve a larger rewrite.

I don't understand the concern about an Agent that always selects the `FinishAction`. It seems to me that an agent which behaves more narrowly is no different than an agent who behaves more broadly.

The current description of agent in our codebase is:

    """Agent is responsible for choosing the next action to take for an AgentService.

    It uses the provided context, and a set of Tools, to decide on an action that will
    be executed by the AgentService.
    """

If we just add the word `possibly` before `a set of Tools` then I'm pretty an agent that chit chats fits that description perfectly fine.

src/steamship/agents/examples/example_chatbot_with_replicate.py
eob: Now that the MMAS has a pretty different structure, there are a few moves we could choose to make if we wanted:

- Relocate the agents for testing into the `tests/assets` folder, which is where some other things like those are stored.
- Keep them here, but remove the ones that weren't used for testing.

Seems like the second one is probably worthwhile in a new PR. The first one I'm unsure if it's a win

GitOnUp:
@eob I had some discussion with other folks at sync this morning that would change quite a bit about this, which I'll be chasing up on rather than getting this to an inbetween state in the meantime

eob:
@GitOnUp @douglas-reid 👍 . I'm going to close this PR given the above comment; but it's always around if we need to chop it for parts!

PR #545, opened 2023-09-07
Add support for optional fields in the config template

eob:


dkolas (approved):
Wow, that turned out to be super straightforward.  Good work!

PR #548, opened 2023-09-11
[SHIP-834] [SHIP-818] Streaming Generators

dkolas:
This PR includes:

- Streaming Generator interface
- Additional methods and status on `Block` for streaming

Depends on: https://github.com/nludb/nludb/pull/631

(tests will fail until deploy of above)


src/steamship/data/block.py
```diff
@@ -27,6 +27,12 @@ class BlockUploadType(str, Enum):
     NONE = "none"  # No upload; plain text only.
 
 
+class StreamState(str, Enum):
+    STARTED = "started"  # A producer has begun streaming to this block.
+    COMPLETE = "complete"  # The producer has finished streaming to this block successfully.
+    ABORTED = "aborted"  # The producer finished streaming to the block, but there was an error.
```
eob, line 33: nit: I don't think this is a huge deal, but if you wanted to align with the enums we use in the task system I think these would be: `running`, `succeeded`, `failed`

eob (approved):
LGTM! This is really exciting

PR #549, opened 2023-09-12
feat(streaming): proof of concept for streaming agentservice execution

douglas-reid:
This PR presents a series of changes that should support a way to stream response information back to a client via an AgentService.

In order to achieve the streaming result, a new method on the AgentService is exposed: `async_prompt`. This new method returns a new `StreamingResponse` that has two fields: `task` and `file`. These fields provide access to (a) the async task that will be streaming results and (b) a file (here the `ChatHistory` file) to which all status messages and assistant interactions will be saved.

This PR relies on a full deployment of https://github.com/steamship-plugins/gpt4/pull/10 to the target environment for testing / validation.

### DIFF
--- a/src/steamship/agents/functional/functions_based.py
+++ b/src/steamship/agents/functional/functions_based.py
@@ -1,9 +1,11 @@
+import json
 from typing import List
 
-from steamship import Block
+from steamship import Block, MimeTypes, Tag
 from steamship.agents.functional.output_parser import FunctionsBasedOutputParser
-from steamship.agents.schema import Action, AgentContext, ChatAgent, ChatLLM, Tool
-from steamship.data.tags.tag_constants import RoleTag
+from steamship.agents.schema import Action, AgentContext, ChatAgent, ChatLLM, FinishAction, Tool
+from steamship.data.tags.tag_constants import RoleTag, TagKind, TagValueKey
+from steamship.data.tags.tag_utils import get_tag
 
 
 class FunctionsBasedAgent(ChatAgent):
@@ -41,21 +43,15 @@ Only use the functions you have been provided with."""
                 .wait()
                 .to_ranked_blocks()
             )
-
             # TODO(dougreid): we need a way to threshold message inclusion, especially for small contexts
-
-            # remove the actual prompt from the semantic search (it will be an exact match)
-            messages_from_memory = [
-                msg
-                for msg in messages_from_memory
-                if msg.id != context.chat_history.last_user_message.id
-            ]
 
         # get most recent context
         messages_from_memory.extend(context.chat_history.select_messages(self.message_selector))
 
         # de-dupe the messages from memory
-        ids = []
+        ids = [
+            context.chat_history.last_user_message.id
+        ]  # filter out last user message, it is appended afterwards
         for msg in messages_from_memory:
             if msg.id not in ids:
                 messages.append(msg)
@@ -67,10 +63,8 @@ Only use the functions you have been provided with."""
         # this should happen BEFORE any agent/assistant messages related to tool selection
         messages.append(context.chat_history.last_user_message)
 
-        # get completed steps
-        actions = context.completed_steps
-        for action in actions:
-            messages.extend(action.to_chat_messages())
+        # get working history (completed actions)
+        messages.extend(self._function_calls_since_last_user_message(context))
 
         return messages
 
@@ -81,4 +75,58 @@ Only use the functions you have been provided with."""
         # Run the default LLM on those messages
         output_blocks = self.llm.chat(messages=messages, tools=self.tools)
 
-        return self.output_parser.parse(output_blocks[0].text, context)
+        future_action = self.output_parser.parse(output_blocks[0].text, context)
+        if not isinstance(future_action, FinishAction):
+            # record the LLM's function response in history
+            self._record_action_selection(future_action, context)
+        return future_action
+
+    def _function_calls_since_last_user_message(self, context: AgentContext) -> List[Block]:
+        function_calls = []
+        for block in context.chat_history.messages[::-1]:  # is this too inefficient at scale?
+            if block.chat_role == RoleTag.USER:
+                return reversed(function_calls)
+            if get_tag(block.tags, kind=TagKind.ROLE, name=RoleTag.FUNCTION):
+                function_calls.append(block)
+            elif get_tag(block.tags, kind=TagKind.FUNCTION_SELECTION):
+                function_calls.append(block)
+        return reversed(function_calls)
+
+    def _to_openai_function_selection(self, action: Action) -> str:
+        fc = {"name": action.tool}
+        args = {}
+        for block in action.input:
+            for t in block.tags:
+                if t.kind == TagKind.FUNCTION_ARG:
+                    args[t.name] = block.as_llm_input(exclude_block_wrapper=True)
+
+        fc["arguments"] = json.dumps(args)  # the arguments must be a string value NOT a dict
+        return json.dumps(fc)
+
+    def _record_action_selection(self, action: Action, context: AgentContext):
+        tags = [
+            Tag(kind=TagKind.ROLE, name=RoleTag.ASSISTANT),
+            Tag(kind=TagKind.FUNCTION_SELECTION, name=action.tool),
+        ]
+        context.chat_history.file.append_block(
+            text=self._to_openai_function_selection(action), tags=tags, mime_type=MimeTypes.TXT
+        )
+
+    def record_action_run(self, action: Action, context: AgentContext):
+        super().record_action_run(action, context)
+
+        tags = [
+            Tag(
+                kind=TagKind.ROLE,
+                name=RoleTag.FUNCTION,
+                value={TagValueKey.STRING_VALUE: action.tool},
+            ),
+        ]
+        # TODO(dougreid): I'm not convinced this is correct for tools that return multiple values.
+        #                 It _feels_ like these should be named and inlined as a single message in history, etc.
+        for block in action.output:
+            context.chat_history.file.append_block(
+                text=block.as_llm_input(exclude_block_wrapper=True),
+                tags=tags,
+                mime_type=block.mime_type,
+            )
--- a/src/steamship/agents/llms/openai.py
+++ b/src/steamship/agents/llms/openai.py
@@ -58,6 +58,7 @@ class OpenAI(LLM):
         if "max_tokens" in kwargs:
             options["max_tokens"] = kwargs["max_tokens"]
 
+        # TODO(dougreid): do we care about streaming here? should we take a kwarg that is file_id ?
         action_task = self.generator.generate(text=prompt, options=options)
         action_task.wait()
         return action_task.output.blocks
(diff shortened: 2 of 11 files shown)


src/steamship/agents/service/agent_service.py
```diff
@@ -295,6 +306,33 @@ def build_default_context(self, context_id: Optional[str] = None, **kwargs) -> A
         context = with_llm(context=context, llm=llm)
         return context
 
+    @post("async_prompt")
+    def async_prompt(
+        self, prompt: Optional[str] = None, context_id: Optional[str] = None, **kwargs
+    ) -> StreamingResponse:
```
douglas-reid, line 312: should we just use `InvocableResponse(status=task, data=ctx.chat_history.file)` instead?

src/steamship/agents/schema/message_selectors.py
```diff
@@ -29,20 +30,34 @@ def is_assistant_message(block: Block) -> bool:
...
+        if len(history_msgs) <= (self.k * 2):
+            return history_msgs
 
         selected_msgs = []
         limit = self.k * 2
-        scope = msgs[len(messages) - limit :]
+        scope = history_msgs[len(history_msgs) - limit :]
         for block in scope:
-            if is_user_message(block) or is_assistant_message(block):
+            if is_user_history_message(block):
```
douglas-reid, line 60: probably don't need this check anymore.

src/steamship/agents/functional/functions_based.py
```diff
@@ -81,4 +75,58 @@ def next_action(self, context: AgentContext) -> Action:
...
-        return self.output_parser.parse(output_blocks[0].text, context)
+        future_action = self.output_parser.parse(output_blocks[0].text, context)
+        if not isinstance(future_action, FinishAction):
+            # record the LLM's function response in history
+            self._record_action_selection(future_action, context)
+        return future_action
+
+    def _function_calls_since_last_user_message(self, context: AgentContext) -> List[Block]:
+        function_calls = []
+        for block in context.chat_history.messages[::-1]:  # is this too inefficient at scale?
```
eob, line 86: This is exactly the line of code I've been wondering about too.

You think if we put some kind of optimized query into the engine/database that returns either the right subset of history or on-the-fly tags demarcate certain things like this, it could be a way to do it that was both faster and also let the python half be simpler?

eob: (fwiw I think it's fine to pursue the slower-but-simple solution now and then we'll optimize it after proving to ourselves it works)

src/steamship/agents/functional/functions_based.py
```diff
@@ -81,4 +75,58 @@ def next_action(self, context: AgentContext) -> Action:
...
+        super().record_action_run(action, context)
+
+        tags = [
+            Tag(
+                kind=TagKind.ROLE,
+                name=RoleTag.FUNCTION,
+                value={TagValueKey.STRING_VALUE: action.tool},
+            ),
+        ]
+        # TODO(dougreid): I'm not convinced this is correct for tools that return multiple values.
```
eob, line 125: Slightly different thought, but the `N -> N` thing has proven to be a real bear to wrestle with. I find myself writing `output[0].text` all over the place, which doesn't smell right at the application-code level.

src/steamship/agents/service/agent_service.py
```diff
@@ -295,6 +306,33 @@ def build_default_context(self, context_id: Optional[str] = None, **kwargs) -> A
...
+
+            # if you can't find a consistent context_id, then there is no way to provide an accurate
+            # streaming endpoint.
+            if not ctx_id:
+                # TODO(dougreid): this points to a slight flaw in the context_keys vs. context_id
+                raise SteamshipError("Error setting up context: no id found for context.")
+
+            task = self.invoke_later(
+                "/prompt", arguments={"prompt": prompt, "context_id": ctx_id, **kwargs}
+            )
```
dkolas, line 333: Forgive me if I'm being dense, but what do we gain out of making this `invoke_later`?

src/steamship/agents/service/agent_service.py
dkolas: Data = the file makes sense to me, rather than having a new type.  See below, I'm not sure I yet understand the task bit.

src/steamship/agents/llms/openai.py
```diff
@@ -119,7 +116,31 @@ def chat(self, messages: List[Block], tools: Optional[List[Tool]], **kwargs) ->
...
+        # the way to detect this would be if all messages were from the same file
+        if self._from_same_file(blocks=messages):
+            file_id = messages[0].id
+            block_ids = [b.id for b in messages]
+            generate_task = self.generator.generate(
+                input_file_id=file_id,
+                input_file_block_index_list=block_ids,
+                options=options,
+                append_output_to_file=True,
+            )
```
dkolas, line 129: If we're going to generate with streaming _blocks_, we also need to pass `streaming=True` here.  That will result in the generate_task.output.blocks being allocated but not yet completed.  This means that the caller then has to be aware of whether or not they're streaming.

See: https://github.com/steamship-core/python-client/pull/548

src/steamship/agents/llms/openai.py
```diff
@@ -119,7 +116,31 @@ def chat(self, messages: List[Block], tools: Optional[List[Tool]], **kwargs) ->
...
+        else:
+            tags = [Tag(kind=TagKind.GENERATION, name=GenerationTag.PROMPT_COMPLETION)]
+            temp_file = File.create(client=self.client, blocks=messages, tags=tags)
+            generate_task = self.generator.generate(input_file_id=temp_file.id, options=options)
+
+        generate_task.wait()
+
+        return generate_task.output.blocks
+
+    def _from_same_file(self, blocks: List[Block]) -> bool:
```
eob, line 139: nit: A comment here about strategy would be super useful.

Above, it seems like this is being used like this:

```
if the blocks we're working with are all from the same file:
  keep it going
else:
  use a temp file
```

src/steamship/agents/service/agent_service.py
dkolas: It's also possible the caller might want a more direct reference to the new blocks that are being streamed as a result of this request?

src/steamship/agents/schema/message_selectors.py
eob: I think it depends on /why/ someone is calling get_messages?

E.g. if I just want the chat transcript to prepare a chat prompt, I'd still want these kinds of things filtered out, but if I want things like status messages or function calls, then I'd want it.

src/steamship/agents/service/agent_service.py
eob: That feels like a really clean return type to me

src/steamship/agents/service/agent_service.py
```diff
@@ -295,6 +306,33 @@ def build_default_context(self, context_id: Optional[str] = None, **kwargs) -> A
...
+    def async_prompt(
+        self, prompt: Optional[str] = None, context_id: Optional[str] = None, **kwargs
+    ) -> StreamingResponse:
+        with self.build_default_context(context_id, **kwargs) as context:
+            ctx_id = context_id
+
+            # if no context ID is provided, we need to make sure that the streaming context ID
+            # is the same one as the non-streaming.
+            if not ctx_id:
+                ctx_file = context.chat_history.file
```
eob, line 319: Maybe we could push this into AgentContext as a `get_default_id` method

src/steamship/agents/service/agent_service.py
```diff
@@ -295,6 +306,33 @@ def build_default_context(self, context_id: Optional[str] = None, **kwargs) -> A
...
+                for tag in ctx_file.tags:
+                    if tag.kind == TagKind.CHAT and tag.name == ChatTag.CONTEXT_KEYS:
+                        if value := tag.value:
+                            ctx_id = value.get("id", None)
+
+            # if you can't find a consistent context_id, then there is no way to provide an accurate
+            # streaming endpoint.
+            if not ctx_id:
+                # TODO(dougreid): this points to a slight flaw in the context_keys vs. context_id
```
eob, line 328: I've gotten the same feeling in use too, since it almost always is the case that just a single ID is used rather than a dict to distinguish.

But the concept of a multi-valued identifier still feels like it makes sense.. maybe the right thing to do is see if we can all put heads together and come up with something that would ONLY work with a full dict as its identifier rather than a string?

tests/steamship_tests/agents/test_agent_service.py
```diff
@@ -248,3 +248,38 @@ def test_context_logging_to_chat_history_everything(client: Steamship):
...
+        assert streaming_resp is not None
+        assert streaming_resp["file"] is not None
+        assert streaming_resp["task"] is not None
+
+        file = File(client=client, **(streaming_resp["file"]))
+        task = Task(client=client, **(streaming_resp["task"]))
+
+        original_len = len(file.blocks)
+
+        task.wait()
```
eob, line 279: The fact that we can still wait on the task is 💯 .

eob (approved):
A few minor comments, but this looks awesome to me.

- It's amazing that streaming has been done in a way where you can't actually see any explicit place where we say we're streaming. I thinks that's a sign that this design will "Just Work" without streaming be a special case folks tiptoe around 🚀🚀🚀
- +1 to a few of your "huh.. do we need to change this?" comments within.

dkolas (commented):
Code looks good.  We may want a sync discussion of how the client might expect to consume the stream to make sure we're sending the things that make it as easy as possible.

src/steamship/agents/service/agent_service.py
douglas-reid: We use `invoke_later` to schedule the work and allow an "early" return with the bits required to monitor that work.

 We call `/prompt` but have passed back the file that all the updates (status, and otherwise) will be posted to, but also have a handle on the `Task` that is doing the actual work.

I _thought_ this was the approach discussed earlier. Are you imagining something different?

src/steamship/agents/llms/openai.py
douglas-reid: Walking through this together is probably the best way to answer the confusion here. I _think_ we want `streaming=True` to *always* be the case really, but I don't think the caller *here* cares, because there is a `wait()` on this task anyway. It is more that the downstream consumer that has access to the `File` cares, right?

src/steamship/agents/service/agent_service.py
douglas-reid: @dkolas we don't have any way to cascade pre-allocated blocks **within** a `File` back -- and I'm not entirely sure that we can (given the current `Agent` + `AgentService` architecture). For instance, the status message blocks, by their very nature, cannot be pre-allocated. Any scheme of pre-allocation would then result in out-of-order messaging, etc.

I _think_ `File` may be the "best" we can do easily. For things that don't require an `Agent` and `Action` and status messages, that may not be the case.

Let's talk this through in person.

tests/steamship_tests/agents/test_agent_service.py
douglas-reid: @dkolas it is complicated. The _existing_ LLM `chat()` call waits on the full `generate()` at the moment. This may change when we add `streaming=True`, but I _think_ we'll want to make that have a way to wait on all the returned blocks to finish streaming (at least, something in the `Agent` must, as the next action must be selected).

So, at the moment, that `task` is waiting on everything. And, it _feels_ like it will always do that. To stream, the client would ignore it and watch the `File` events (which I thought was the way we wanted it).

PR #550, opened 2023-09-15
feat(tasks): expose request ID field on Tasks

douglas-reid:
This PR is part of the work to enable streaming use cases for the Agent SDK. With this change, it will be possible to extract the request ID from a `Task` obj. This request ID can then be used, in concert with a `File`, to stream updates (blocks added).


dkolas (approved):
Nice!

PR #551, opened 2023-09-15
fix(empty messages): don't embed in chat history

douglas-reid:


dkolas (approved):
This works for me... we may want to throw an error at a lower level in the stack

PR #554, opened 2023-09-25
feat(streaming): initial implementation of streaming agentservice execution

douglas-reid:
This PR establishes an initial implementation of streaming utilities for `AgentService` endpoints, based on `FunctionsBasedAgent` runs. This will allow clients to invoke an agent run through an async endpoint and stream, via SSE, block creation events related to the agent's execution. These events can include status messages (from Agents and Tools) as well as generated blocks (from LLMs and Tools). 

With this code, new async endpoints can be exposed with code like the following:

```python
    @post("async_prompt")
    def async_prompt(
        self, prompt: Optional[str] = None, context_id: Optional[str] = None, **kwargs
    ) -> StreamingResponse:
        ctx_id, history_file = self._streaming_context_id_and_file(context_id=context_id, **kwargs)
        task = self.invoke_later(
            "/prompt", arguments={"prompt": prompt, "context_id": ctx_id, **kwargs}
        )
        return StreamingResponse(task=task, file=history_file)
```

The PR ensures all `Block`s created in the service of a `run_agent` call are tagged with the proper `request-id`, and that all calls to the LLM are streaming-compatible.

A test is provided that demonstrates an approach to consuming a stream of events, based on the sseclient-py library. A Generator is constructed that emits `Block`s until a terminal block for a request is found in the stream.

### DIFF
--- a/src/steamship/agents/service/agent_service.py
+++ b/src/steamship/agents/service/agent_service.py
@@ -3,12 +3,16 @@ import uuid
 from collections import defaultdict
 from typing import Dict, List, Optional
 
-from steamship import Block, SteamshipError, Task
+from pydantic.main import BaseModel
+
+from steamship import Block, File, SteamshipError, Task
 from steamship.agents.llms.openai import OpenAI
 from steamship.agents.logging import AgentLogging, StreamingOpts
 from steamship.agents.schema import Action, Agent, FinishAction
 from steamship.agents.schema.context import AgentContext, EmitFunc, Metadata
 from steamship.agents.utils import with_llm
+from steamship.data import TagKind
+from steamship.data.tags.tag_constants import ChatTag
 from steamship.invocable import PackageService, post
 
 
@@ -32,6 +36,11 @@ def build_context_appending_emit_func(
             )
 
     return chat_history_append_func
+
+
+class StreamingResponse(BaseModel):
+    task: Task
+    file: File
 
 
 class AgentService(PackageService):
@@ -116,6 +125,7 @@ class AgentService(PackageService):
             },
         )
 
+        # save action selection to history...
         return action
 
     def run_action(self, agent: Agent, action: Action, context: AgentContext):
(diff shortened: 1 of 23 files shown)


src/steamship/agents/service/agent_service.py
```diff
@@ -2,15 +2,24 @@
...
 from steamship.agents.logging import AgentLogging, StreamingOpts
 from steamship.agents.schema import Action, Agent, FinishAction
 from steamship.agents.schema.context import AgentContext, Metadata
 from steamship.agents.utils import with_llm
+from steamship.data import TagKind
+from steamship.data.tags.tag_constants import ChatTag
 from steamship.invocable import PackageService, post
 
 
+class StreamingResponse(BaseModel):
```
eob, line 18: @douglas-reid if you extend `BaseModel` instead of `CamelModel` then the serialization is underscored rather than the camelcase that we use for engine comms.

It sneaks through in the case since it's internal to an object's response, but it breaks the parsing of the task in the Typescript client.

Would it be possible to add a test that, e.g., `response.task.requestId` is serialized as `requestId` instead of `request_id` just to make sure that the clients demarshalling it won't explode?

douglas-reid: sure

tests/steamship_tests/agents/test_agent_service.py
```diff
@@ -248,3 +252,147 @@ def test_context_logging_to_chat_history_everything(client: Steamship):
...
+                            assistant_chat_response_event = True
+
+        file.refresh()
+        assert (
+            len(file.blocks) > original_len
+        ), "File should have increased in size during AgentService execution"
+
+        assert num_blocks > 0, "Blocks should have been streamed during execution"
+        assert llm_prompt_event_count == 2, (
+            "At least 2 llm prompts should have happened (first for tool selection, "
```
eob, line 337: These error messages are awesome -- thank you for adding them

src/steamship/agents/service/agent_service.py
```diff
@@ -338,6 +356,42 @@ def build_default_context(self, context_id: Optional[str] = None, **kwargs) -> A
...
+        history_file = self._history_file_for_context(context_id=context_id)
+        ctx_id = context_id
+
+        # if no context ID is provided, try to get a consistent ID from history_file
+        if not ctx_id:
+            ctx_id = _context_key_from_file(key="id", file=history_file)
+
+        # if you can't find a consistent context_id, then something has gone wrong, preventing streaming
+        if not ctx_id:
+            # TODO(dougreid): this points to a slight flaw in the context_keys vs. context_id
```
eob, line 380: Is this the situation where in practice we've just been using `context_id` whereas in theory we're hashing the whole `context_keys` object?

src/steamship/agents/service/agent_service.py
```diff
@@ -338,6 +356,42 @@ def build_default_context(self, context_id: Optional[str] = None, **kwargs) -> A
...
+            raise SteamshipError("Error setting up context: no id found for context.")
+
+        return ctx_id, history_file
+
+    @post("async_prompt")
+    def async_prompt(
+        self, prompt: Optional[str] = None, context_id: Optional[str] = None, **kwargs
+    ) -> StreamingResponse:
+        ctx_id, history_file = self._streaming_context_id_and_file(context_id=context_id, **kwargs)
+        task = self.invoke_later(
```
eob, line 390: I think in a V2 we'll want to follow up and see if we can have the things that are later-invoked auto-block this task from completing (thinking about the comment string about the semantics of the Task below.. we could maybe try to make conformance to that automatic)

eob (approved):
A few non-blocking thoughts in there, but this looks fantastic. Excited for it to be merged!!

src/steamship/agents/service/agent_service.py
douglas-reid: yeah. we have a mismatch in our exposed args in `/prompt` vs. our actual capability. not sure if that matter practically yet.

PR #556, opened 2023-09-26
fix(agents): append assistant msg in answer endpoint

douglas-reid:


douglas-reid:
I think this is likely better fixed by #558 .

dkolas (commented):
Agree that this is fine but #558 is better.

douglas-reid:
Closing in favor of #558

PR #559, opened 2023-09-26
Add bot handle reporting to Telegram Transport

eob:
It looks like back in June, the reporting method which returned information about the bot itself was removed from the transport -- probably accidentally.

This is used to generate a link in the web UI that tells you which bot you've connected your agent to.

### DIFF
--- a/src/steamship/agents/mixins/transports/telegram.py
+++ b/src/steamship/agents/mixins/transports/telegram.py
@@ -80,12 +80,22 @@ class TelegramTransport(Transport):
                 f"Could not set webhook for bot. Webhook URL was {webhook_url}. Telegram response message: {response.text}"
             )
 
+    @post("telegram_bot_info")
+    def telegram_bot_info(self) -> dict:
+        api_root = self.get_api_root()
+        if not api_root:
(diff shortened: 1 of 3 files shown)


src/steamship/agents/mixins/transports/telegram.py
```diff
@@ -294,7 +304,7 @@ def get_api_root(self) -> Optional[str]:
         if bot_token:
             if ".steamship.run/" in api_base or ".apps.staging.steamship.com" in api_base:
                 # This is a special case for our testing pipeline -- it contains a mock Telegram server.
-                return api_base
+                return api_base[:-1]
```
eob, line 307: It turns out that even the testing one shouldn't have a / character..

tests/assets/packages/transports/mock_telegram_api.py
```diff
@@ -84,3 +84,9 @@ def send_video(
             ],
         )
         return "OK"
+
+    @get("getMe", public=True)
```
eob, line 88: Simulation of the telegram API endpoint.

douglas-reid (approved):

PR #560, opened 2023-09-27
New runs should exist in their own workspace

eob:
@GitOnUp mentioned this was really surprising to him, and it does feel like bug:

Unless a workspace is specified, `ship run local` will always run the agent in the `default` workspace which means the data and chat histories from multiple runs will be commingled.

This PR:

- Creates a new workspace if one was not specified.
- Reports the workspace in the logging output so you know where you're running
- Alters the semantics if a workspace IS specified to create it if it doesn't already exist.

### DIFF
--- a/src/steamship/cli/cli.py
+++ b/src/steamship/cli/cli.py
@@ -272,7 +272,8 @@ def serve_local(  # noqa: C901
     client, user, manifest = initialize_and_get_client_and_prep_project()
 
     if workspace:
-        workspace_obj = Workspace.get(client, handle=workspace)
+        # Fetch/create a workspace if one was specified.
+        workspace_obj = Workspace.create(client, handle=workspace, fetch_if_exists=True)
     else:
         # Create a new workspace if none was specified.
         # Otherwise multiple runs co-mingle data in the `default` workspace.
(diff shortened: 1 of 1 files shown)


src/steamship/cli/cli.py
```diff
@@ -272,11 +272,17 @@ def serve_local(  # noqa: C901
...
 
     if workspace:
-        workspace_obj = Workspace.get(client, handle=workspace)
+        # Fetch/create a workspace if one was specified.
+        workspace_obj = Workspace.create(client, handle=workspace, fetch_if_exists=True)
     else:
-        workspace_obj = Workspace.get(client)
+        # Create a new workspace if none was specified.
+        # Otherwise multiple runs co-mingle data in the `default` workspace.
+        workspace_obj = Workspace.create(client)
```
GitOnUp, line 280: Is this namespaced at all to the package/instance that's being used?  It just might be harder to clean up your steamship workspace later without that mapping.

Something like appending the unix ts could help (or a random string suffix that also checks against collision).

GitOnUp (approved):
LGTM with one QOL suggestion

douglas-reid (approved):

PR #561, opened 2023-09-27
feat: `ship run local` should direct to, and work with, the live Agent Management Console

eob:
Before, the `ship run local` command directed to a limited debugging page which had just a bare chat interface that hit a local endpoint provided by URL query parameter.

But this left several important things unaddressed:
- How should the user connect Telegram or Slack to their locally running agent?
- How could the user see the files, tasks, and logs being created by their locally running agent?
- How could the user try out multiple chat sessions?

This PR connects the locally running agent to the live console. It removes the code that simulated the "proxy responses" from the local server so that it can communicate correctly via the Steamship Proxy in the wild.

https://github.com/steamship-core/python-client/assets/63262/87e9aa50-44b5-4f7b-be3e-3cb1fc67e576

### DIFF
--- a/.github/workflows/test-main.yml
+++ b/.github/workflows/test-main.yml
@@ -70,7 +70,7 @@ jobs:
         run: python -m pip install -e .
 
       - name: Run Tests
-        run: pytest --junitxml=tests.xml
+        run: pytest --junitxml=tests.xml -n 5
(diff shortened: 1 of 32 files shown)


dkolas (approved):
Looks good!

PR #562, opened 2023-09-27
Limit specific tools by name

dkolas:
Per request: https://discord.com/channels/927675537390968882/1156666953440698498


douglas-reid (approved):
Looks reasonable. I've wondered if we should make the `max_actions` stuff runtime accessible too.  maybe with some sort of `AgentOptions` that can be parsed out of kwargs.

but that's for later.

PR #564, opened 2023-09-28
Agents Refactor 1/2: Plugin Capabilities and SteamshipLLM prototype 

GitOnUp:
## PR Structure Notes
I've got a large number of `# TODO (PR Callout)` or `# TODO (PR)` form comments in this PR, which are generally discussion topics rather than actual TODOs.  I've preseeded comment threads on those for easy discovery, in the case that there's not more to say other than the comment I've just quoted the comment.

I've also prototyped what I think SteamshipLLM looks like here, and given the number of open questions I wanted to clear those before finalizing the more formulaic pieces.

## Overview

This PR provides the ability for Plugins to assert their capabilities and for users to request that they be fulfilled, along with parameters for those capabilities if applicable.  Those parameters are most relevant in the case of functions for GPT-4, but could be extended in the general case for other capabilities.

Step 2 is to adapt the plugin arch to use this and have Agent determine that certain capabilities should be used based on e.g. providing tools or a ChatHistory.

### Why blocks and temp files instead of on Options?
- The response contains only blocks.  Without engine changes, this means we can communicate back and forth via the same channel consistently.  
- I had considered options as a venue for communicating requirements, but I also want to be able to communicate that yes, we did fulfill these requirements, and here's how.
- OpenAI's ChatLLM does this already

### Why MIME type instead of Tags?
Blocks have one MIME type, and in the case that it should only be parsed one way with one format, it makes sense to me to clamp that.

---
Behavior overview lifted from the docstring on capabilities.py:

```
Capabilities are a concept communicated back and forth between plugins and client code via blocks, and they are meant to
indicate client code's need for certain levels of support of a range of features.

Clients can request NATIVE, BEST_EFFORT, or OPTIONAL support for features that a plugin may or may not support.  Plugins
are expected to parse this and fail-fast if the user has requested support for a feature that the plugin does not
support, so that users are not e.g. billed for usage they can't incorporate.

Capability requests can include other information on the request itself, but oftentimes indicate that certain blocks
will be tagged in Steamship-native ways as part of the rest of the payload.  For example, ConversationSupport is a
capability that indicates the CHAT TagKind will be included in blocks that are part of the input, and the plugin is
expected to incorporate these with a model that supports them.

In the case that a Plugin does not support behavior indicated by the Capability request, it will throw, listing the
models that it could not support at the levels requested.  Otherwise, when Plugins respond, they'll include another
block indicating at which level they served the requested capabilities.
```


src/steamship/agents/llms/steamship_llm.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+class SteamshipLLM:
+    """A class wrapping LLM plugins."""
+
+    def __init__(self, plugin_instance: PluginInstance):
+        # TODO (PR callout): In the spirit of getting folks to pin to specific instances more often, take the route of
+        #  using specific PluginInstances rather than client.use_plugin all the time with the params that we determine?
+        self.client = plugin_instance.client
+        self.plugin_instance = plugin_instance
```
GitOnUp, line 24: Open design question: take plugin instance for SteamshipLLM's constructor?

src/steamship/agents/llms/steamship_llm.py
```diff
@@ -0,0 +1,110 @@
...
+
+    @staticmethod
+    def with_gpt4(client: Steamship, temperature: float = 0.4, max_tokens: int = 256):
+        # TODO (PR callout): This is an example of helpers that can initialize plugins that we know are solid / are
+        #  steamship-published.  These don't need to live in this class either, to keep the interface more pure.
+        gpt4 = client.use_plugin(
+            "gpt-4", config={"model": "gpt-4", "temperature": temperature, "max_tokens": max_tokens}
+        )
+        return SteamshipLLM(gpt4)
```
GitOnUp, line 33: Calling out ability to convenience-function these, whether they're `@staticmethod` or external

src/steamship/agents/llms/steamship_llm.py
```diff
@@ -0,0 +1,110 @@
...
+          ConversationSupport capability.
+        :param tools: A list of Tools, representing functions which can be called with this LLM. Providing this requests
+          the FunctionCallingSupport capability, and provides the function definitions for it.
+        :param assert_capabilities: If True (default), raise a SteamshipError if the LLM plugin did not respond with a
+          block that asserts what level capabilities were fulfilled at.
+        :param kwargs: Options that can be passed to the LLM model as other parameters.
+        :return: a List of Blocks that are returned from the plugin.
+        """
+
+        # TODO (PR callout): I'm not certain this class needs to be abstract?  This seems pretty soup-to-nuts.
```
GitOnUp, line 60: Discussion: doesn't seem like SteamshipLLM needs to be an ABC.

src/steamship/agents/llms/steamship_llm.py
```diff
@@ -0,0 +1,110 @@
...
+            blocks += history
+            capabilities.append(ConversationSupport())
+
+        if tools:
+            capabilities.append(
+                FunctionCallingSupport(functions=[tool.as_openai_function() for tool in tools])
+            )
+
+        blocks.append(CapabilityPluginRequest(requested_capabilities=capabilities).to_block())
+        # TODO (PR callout): I have maybe ideas for not needing temp files here in all cases? GH comment
```
GitOnUp, line 77: Probably outside the scope of this PR, but if we're already posting files, just post blocks to the client as part of the request eventually?

src/steamship/agents/llms/steamship_llm.py
```diff
@@ -0,0 +1,110 @@
...
+                    break
+            else:
+                if assert_capabilities:
+                    version_string = f"{self.plugin_instance.plugin_handle}, v.{self.plugin_instance.plugin_version_handle}"
+                    raise SteamshipError(
+                        f"Asserting capabilities are used, but capability response was not returned by plugin ({version_string})"
+                    )
+        finally:
+            # TODO (PR callout): It looks like current OpenAI impl might not do this?
+            temp_file.delete()
```
GitOnUp, line 106: This seemed like a resource leak to me.

src/steamship/agents/llms/steamship_llm.py
```diff
@@ -0,0 +1,110 @@
...
+
+        # TODO (PR callout): Use special output parser in the meantime to extract the tool to call from output block
+        #  tagged with new tags, _easy_ slot in with existing FunctionsBasedLLMAgent.
+        return generation_task.output.blocks
```
GitOnUp, line 110: In TagConstants, we have a new CAPABILITY_RESPONSE kind, name would be the name of the capability, and value would be per-capability.

e.g. in the case of Functions, return steamship.function-calling as the name, and "tool_name" as the value for which tool to call, and the block(s) that is(/are) tagged are the input to the tool.

src/steamship/plugin/capabilities.py
```diff
@@ -0,0 +1,291 @@
...
+
+        name: str = Field(init_var=False, default=None)
+        fulfilled_at: RequestLevel
+
+
+class CapabilityImpl(Capability, extra=Extra.forbid):
+    # TODO (PR): The Extra.forbid here is to enforce clamping of deserialization, but it may just be simpler to leave
+    #  that up to individual capabilities?  My goal here is to prevent accidental breakage of contract when someone
+    #  provides specific metadata that makes it so e.g. a plugin with an older view of the world thinks it has Best
+    #  Effort support but can't because of those extra requests.
```
GitOnUp, line 121: Discussion: adding extra details to a capability seems dangerous because it adds extra things to support that a plugin may not know about.

Also note to self, the extra as a metaclass argument here should get moved into a Config inner class for consistency.

src/steamship/plugin/capabilities.py
```diff
@@ -0,0 +1,291 @@
...
+        super().__init__(
+            f"The following features were requested but are not supported by this plugin: {names}"
+        )
+        self.unsupported_capabilities = unsupported_capabilities
+
+
+class CapabilityPluginRequest(BaseModel):
+    """Model representing the text in a STEAMSHIP_PLUGIN_CAPABILITIES block when it is used as a request."""
+
+    # TODO (PR): Might be cleaner to have separate MIMEs for response/request.
```
GitOnUp, line 149: Assuming the MIME/block approach I have here makes sense, I'm just going to do this for clarity.

src/steamship/plugin/capabilities.py
```diff
@@ -0,0 +1,291 @@
...
+class FunctionCallingSupport(CapabilityImpl):
+    """This plugin supports function calling.
+
+    Function definitions come across as a list of function definitions.
+    """
+
+    name = "steamship.function_calling_support"
+
+    functions: List[OpenAIFunction]
+    # TODO (PR): Worth generalizing this?
```
GitOnUp, line 291: Discussion: at very least it seems like generalizing the name here makes sense.

GitOnUp (commented):
Initial comment chain seeding

src/steamship/agents/llms/steamship_llm.py
```diff
@@ -0,0 +1,110 @@
...
+    """A class wrapping LLM plugins."""
+
+    def __init__(self, plugin_instance: PluginInstance):
+        # TODO (PR callout): In the spirit of getting folks to pin to specific instances more often, take the route of
+        #  using specific PluginInstances rather than client.use_plugin all the time with the params that we determine?
+        self.client = plugin_instance.client
+        self.plugin_instance = plugin_instance
+
+    @staticmethod
+    def with_gpt4(client: Steamship, temperature: float = 0.4, max_tokens: int = 256):
```
eob, line 27: **[Style, Nonblocking]**

I really like what this implies.. sort of:

```
SteamshipLLM.with_gpt4
SteamshipLLM.with_llama2
SteamshipLLM.with_vicuna

etc
```

A few thoughts:

* To the extent that there are common, known params, I think this is an opportunity to officially type and standardize them in the SDK (temperature, etc)
* The `with_` feels a bit weird to me.. I think because it's a Python reserved word used in a very specific way that isn't the way this is being used here. What if it was just: `LLM.gpt4()` or even `LLM.gpt()`
* I'm not sure if we need the `Steamship` prefix since that's already the scope of this entire library

src/steamship/agents/llms/steamship_llm.py
eob: Feels fine to me.

src/steamship/agents/llms/steamship_llm.py
eob: At a high level, if it's possible to have a single LLM class that really does check the boxes we need, that feels like a great win.

eob (commented):
First pass of comments - will take a second pass & want to make sure @douglas-reid gets a chance to weigh in too. this is looking pretty great though

src/steamship/agents/llms/steamship_llm.py
GitOnUp: Re: `with_`, I'm emulating `steamship.agents.utils.with_llm`.  Prefixing these in *some* way is nice because:
- IDE users can type `.with_` and get a list of things they can use
- It avoids collisions with method names we might otherwise want to use in the case of someone naming their model in a generic way

Using `Steamship` as a prefix was a suggestion in the design proposal:
> Consolidate LLM and ChatLLM into single abstract base class (proposed name: SteamshipLLM to avoid collisions on updates).

src/steamship/agents/llms/steamship_llm.py
GitOnUp: There might be an opportunity for naming clarification here if:
- `messages` --> `current_prompt`
- `history` --> `previous_chat_history`

Alternatively, `history` could go away and be replaced with `use_chat_completion: bool`?

src/steamship/agents/llms/steamship_llm.py
douglas-reid: Do you have a clear code example of this somewhere?

src/steamship/plugin/capabilities.py
douglas-reid: I think this should probably be `Tool`-specific naming, just to keep things clearly in Steamship-land and not OpenAI land.

I was hoping we'd have some serialization of `Tool`, such that this could be a `List[Tool]`.

One thing we'll want is some way to specify the serialization of params for a `Tool`. Right now, we built it assuming everything is blocks-in, blocks-out. But our users don't want to wrap the parsing internally.

So, having some meta wrapper that allows `List[Block] --> params` would be extra credit.

tests/steamship_tests/plugin/unit/test_capabilities.py
```diff
@@ -0,0 +1,114 @@
...
+    block = original.to_block()
+    roundtripped = CapabilityPluginResponse.from_block(block)
+    assert original == roundtripped
+
+
+def _make_input_blocks() -> List[Block]:
+    return [
+        Block(
+            text="Can you make me an image of a whale?",
+            tags=[Tag(kind=TagKind.CHAT, name=ChatTag.HISTORY)],
```
douglas-reid, line 76: FWIW, so far, we only use `ChatTag.HISTORY` on the `File`-level tags for the `ChatHistory` file. There are several reasons for this, including that a message is only `HISTORY` based on other context. For instance, at first, it is just a `MESSAGE`.

src/steamship/agents/llms/steamship_llm.py
douglas-reid, line 27: I feel like we've been burned before by not allowing `kwargs` on the end of these types of methods. I suggest allowing the `config` to include `**kwargs` style expansion for forward-compat with the plugin.

douglas-reid, line 27: This is a style choice, but our `gpt-4` plugin allows other models. I think we either need to make a different builder-style method, expose `model_name` as an explicit parameter, or something equivalent.

This might also be a good time to think about renaming this to reflect provider vs. model type (or some other aspect).  `SteamshipLLM.using_openai_chatcomplete` perhaps, or (less-clear) `SteamshipLLM.using_gpt` ?

douglas-reid: Instead of `with` and the python association, we could use: `using_` or `from_` or even `wrapping_` and get the same benefits. WDYT?

- `SteamshipLLM.using_openai_chatcomplete` or
- `SteamshipLMM.using_replicate_model` or
- `SteamshipLLM.from_openai` or
- `SteamshipLLM.wrapping_openai` or ...

src/steamship/plugin/capabilities.py
douglas-reid: I'd much prefer that we serialize `Tool` and pass that around. `List[Tool]`. That would prevent any sort of intermediate step in the SDK that is backend specific (until OpenAIFunction becomes a _de facto_ standard, at least).

Bonus points if we can figure out a way to include parameterization of `Tool::run` in a way that will auto-handle `Block --> param` translation, so that devs can write code like `def run(self, query: str, location: str)` and not have to worry about extracting those values from `List[Block]`.

src/steamship/agents/llms/steamship_llm.py
dkolas: @GitOnUp I agree that burying calls to use_plugin is probably bad form.

src/steamship/agents/llms/steamship_llm.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def __init__(self, plugin_instance: PluginInstance):
+        # TODO (PR callout): In the spirit of getting folks to pin to specific instances more often, take the route of
+        #  using specific PluginInstances rather than client.use_plugin all the time with the params that we determine?
+        self.client = plugin_instance.client
+        self.plugin_instance = plugin_instance
+
+    class Impls:
+        @staticmethod
+        def gpt(
```
dkolas, line 28: May want a `version_handle` arg for these to pass through to use_plugin, otherwise there's no way to control what version you get.

dkolas (dismissed):
I think this all tracks for me! Seems like a great way to sort out the differences between the LLM plugins.

A few comments / suggestions / possible minor changes below.

src/steamship/agents/llms/steamship_llm.py
GitOnUp: `model` handles that for this specific plugin usage, but yes, e.g. Replicate would want that.

GitOnUp: I just realized this probably means version of the plugin, rather than version of the model.

GitOnUp:
Merging as discussed at sync

PR #568, opened 2023-09-29
fix(agents): ensure proper messaging and tool serialization

douglas-reid:
In an investigation into somewhat random duplicate message issues, we uncovered a few small issues that this PR cleans up:
- don't repeat the user prompt
- use proper JSON serialization of tools
- catch improper JSON serialization coming back from OpenAI.


eob (approved):
LGTM!

PR #570, opened 2023-10-03
fix(agents): generate proper history blocks in agent

douglas-reid:
With the inflight changes for streaming, we unfortunately merged a commit that left message selection in a broken state. The existing testing was not enough to capture the issue. This PR attempts to restore proper message selection functionality.

Follow on PRs will be necessary to clean up and streamline the selection bits added in this PR.

### DIFF
--- a/src/steamship/agents/functional/functions_based.py
+++ b/src/steamship/agents/functional/functions_based.py
@@ -1,9 +1,12 @@
+import json
+from operator import attrgetter
 from typing import List
 
-from steamship import Block
+from steamship import Block, MimeTypes, Tag
 from steamship.agents.functional.output_parser import FunctionsBasedOutputParser
-from steamship.agents.schema import Action, AgentContext, ChatAgent, ChatLLM, Tool
-from steamship.data.tags.tag_constants import RoleTag
+from steamship.agents.schema import Action, AgentContext, ChatAgent, ChatLLM, FinishAction, Tool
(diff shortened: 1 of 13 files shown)


src/steamship/agents/functional/functions_based.py
```diff
@@ -67,10 +72,8 @@ def build_chat_history_for_tool(self, context: AgentContext) -> List[Block]:
         # this should happen BEFORE any agent/assistant messages related to tool selection
         messages.append(context.chat_history.last_user_message)
```
eob, line 73: Am I right that this will result in the following message list:

- System Message
- Topic-Selected Ancient Messages In Order
- Last User Message
- Last Action

I feel like if I was the LLM, I would want to see at the very least the last K messages from the history, in order, and the topic-selected list shouldn't include any that were included in the Last K messages list.

eob (approved):

src/steamship/agents/functional/functions_based.py
douglas-reid: We have:

1. System message
2. Semantic match historical messages
3. Context-window messages (user <--> assistant and currently assistant<-->tool)
4. Last User message
5. Scratchpad messages.

(2) and (3) should be de-duped and sorted by file index (which is a current placeholder for temporal order).

So, I think that we are meeting your notional LLM requirements.

PR #571, opened 2023-10-04
docs: add notice about ngrok acc

maxwfreu:
Adds a disclaimer that an ngrok acc is required for running `ship run local`


eob (approved):
LGTM

PR #575, opened 2023-10-06
Warn against unsupported Python version when running CLI

GitOnUp:
During the initialization of the CLI, check the Python version and warn if it's not supported.

I took this approach to attempt to be nice, rather than immediately explode in the SDK itself, though the SDK could certainly leverage the `__init__.py` code here.

-------------

When running on 3.8:

Screenshot 2023-10-05 at 11 10 45 PM

When running on 3.11:

Screenshot 2023-10-05 at 11 10 59 PM

When running on 3.10:

Screenshot 2023-10-05 at 11 14 09 PM

### DIFF
--- a/src/steamship/__init__.py
+++ b/src/steamship/__init__.py
@@ -34,21 +34,17 @@ from .data import (
 
 from .client import Steamship  # isort:skip
 
-# None acts as wildcard here.  Major version number is always assumed to be non-None.
-SUPPORTED_PYTHON_VERSIONS = [(3, 10, None)]
+SUPPORTED_PYTHON_VERSIONS = [(3, 10)]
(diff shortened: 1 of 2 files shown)


dkolas (approved):
I really like where / how you've injected this.  It feels like it's at a place where you have the developer's attention but warning them about something won't feel overly annoying or hostile.

Only q: Do we want to specifically call out the issues around 3.11 in the warning message? To potentially make it less likely to be hand waved away in that case?

GitOnUp:
> Only q: Do we want to specifically call out the issues around 3.11 in the warning message? To potentially make it less likely to be hand waved away in that case?

I think that could be misleading, since we *know of* issues with enums, but there could be more issues that we haven't found yet.

PR #580, opened 2023-10-07
bugfix: Avoid null reference to last_user_message in ChatHistory

eob:
@douglas-reid I'm not 100% sure if I'm fixing an error or enabling odd uses of ChatHistory, but some experiments I was doing in the game resulted in attempting to do a completion on a history where the user hadn't yet chimed in, and that caused None-deferences in the code touched in this PR.

Does this look OK to you? I gave it an eyeball scan, but I know this code is very delicate..

### DIFF
--- a/src/steamship/agents/functional/functions_based.py
+++ b/src/steamship/agents/functional/functions_based.py
@@ -45,11 +45,15 @@ Only use the functions you have been provided with."""
 
         messages_from_memory = []
         # get prior conversations
-        if context.chat_history.is_searchable():
+        _last_user_message = context.chat_history.last_user_message
+        if (
+            context.chat_history.is_searchable()
+            and _last_user_message
+            and _last_user_message.text
+            and len(_last_user_message.text.strip())
+        ):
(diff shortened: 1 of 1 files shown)


src/steamship/agents/functional/functions_based.py
```diff
@@ -45,11 +45,15 @@ def build_chat_history_for_tool(self, context: AgentContext) -> List[Block]:
...
+            context.chat_history.is_searchable()
+            and _last_user_message
+            and _last_user_message.text
+            and len(_last_user_message.text.strip())
+        ):
             messages_from_memory.extend(
-                context.chat_history.search(context.chat_history.last_user_message.text, k=3)
-                .wait()
-                .to_ranked_blocks()
+                context.chat_history.search(_last_user_message, k=3).wait().to_ranked_blocks()
```
douglas-reid, line 56:
```suggestion
                context.chat_history.search(_last_user_message.text, k=3).wait().to_ranked_blocks()
```

PR #582, opened 2023-10-11
Generator Request Tags

dkolas:
This PR provides the structure for passing Tags into Generation requests.  The Tags that are passed will be applied by the Engine to all Blocks which are created as part of the generate request.


eob (dismissed):
LGTM

PR #583, opened 2023-10-12
Agents Refactor 2/2 (client side): Use capabilities for LLMs, agent impl.

GitOnUp:
Implement SteamshipLLM and Plugin Capabilities via Agents.

Highlights:
- FunctionsBasedAgent in steamship.agents.functions_based, mostly copied from the existing one, but using plugin capabilities.
- Example Tool for CustomLLMPrompt takes Steamship LLM, which assumes you're providing a PluginInstance.  That said, the PluginInstance might be more direct.
- Factored chat history building out into shared code.
- This change needs plugins to adapt to new behavior, coming in another PR.

### DIFF
--- a/src/steamship/agents/basic_chat.py
+++ b/src/steamship/agents/basic_chat.py
@@ -1,0 +1,49 @@
+from steamship import SteamshipError
+from steamship.agents.llms.steamship_llm import SteamshipLLM
+from steamship.agents.schema import Action, Agent, AgentContext, FinishAction
+from steamship.agents.utils import build_chat_history
+from steamship.data.tags.tag_constants import RoleTag
+from steamship.plugin.capabilities import ConversationSupport, SystemPromptSupport
+
+DEFAULT_PROMPT = """You are a helpful AI assistant.
+You chat with users are eager to engage on a variety of topics. You answer their questions, provide help thinking
+through challenges they have, and engage as a trusted assistant."""
+
+
+class BasicChatAgent(Agent):
+    """BasicChatAgent implements a conversational agent with no function calling or tool reasoning.
+
+    This Agent class is useful in a number of situations:
+    1) You are looking for an agent with nothing more than open-ended chat, possibly with personality and backstory.
+    2) You are using an LLM that does not support Function-calling natively
+    3) You are using an LLM that, in practice, performs poorly with ReACT-style prompts
+    In these cases, an agent whose only loop does not include an attempt to reason about tool invocation is ideal.
+    A secondary side effect of eliminating tooling is that the less processing must occur before a response begins.
+    """
+
+    PROMPT = DEFAULT_PROMPT
+
+    def __init__(self, llm: SteamshipLLM, **kwargs):
+        # Throw if the user has provided tools as a way to ensure the user understands the limitation of this agent.
+        if "tools" in kwargs:
+            raise SteamshipError(
+                "BasicChatAgent does not support tools. For tool-based agents, please use a base class such as FunctionsBasedAgent."
+            )
+
+        super().__init__(llm=llm, tools=[], **kwargs)
+        self.capabilities = [
+            SystemPromptSupport(),
+            ConversationSupport(),
+        ]
+        self.llm = llm
+
+    def next_action(self, context: AgentContext) -> Action:
+        # Build the Chat History that we'll provide as input to the action
+        messages = build_chat_history(self.PROMPT, self.message_selector, context)
+
+        # call chat() with a hard-coded absense of tools
+        output_blocks = self.llm.generate(messages=messages, tools=[])
+
+        for block in output_blocks:
+            block.set_chat_role(RoleTag.ASSISTANT)
+        return FinishAction(output=output_blocks, context=context)
--- a/src/steamship/agents/basic_chat.py
+++ b/src/steamship/agents/basic_chat.py
@@ -41,7 +41,7 @@ class BasicChatAgent(Agent):
         # Build the Chat History that we'll provide as input to the action
         messages = build_chat_history(self.PROMPT, self.message_selector, context)
 
-        # call chat() with a hard-coded absense of tools
+        # call chat() with a hard-coded absence of tools
         output_blocks = self.llm.generate(messages=messages, tools=[])
 
         for block in output_blocks:
--- a/src/steamship/agents/basic_chat.py
+++ b/src/steamship/agents/basic_chat.py
@@ -3,15 +3,17 @@ from steamship.agents.llms.steamship_llm import SteamshipLLM
 from steamship.agents.schema import Action, Agent, AgentContext, FinishAction
 from steamship.agents.utils import build_chat_history
 from steamship.data.tags.tag_constants import RoleTag
-from steamship.plugin.capabilities import ConversationSupport, SystemPromptSupport
+from steamship.plugin.capabilities import ConversationSupport, RequestLevel, SystemPromptSupport
 
 DEFAULT_PROMPT = """You are a helpful AI assistant.
 You chat with users are eager to engage on a variety of topics. You answer their questions, provide help thinking
 through challenges they have, and engage as a trusted assistant."""
 
 
-class BasicChatAgent(Agent):
+class _BasicChatAgent(Agent):
     """BasicChatAgent implements a conversational agent with no function calling or tool reasoning.
+
+    This class is under active development.
 
     This Agent class is useful in a number of situations:
     1) You are looking for an agent with nothing more than open-ended chat, possibly with personality and backstory.
@@ -33,7 +35,7 @@ class BasicChatAgent(Agent):
         super().__init__(llm=llm, tools=[], **kwargs)
         self.capabilities = [
             SystemPromptSupport(),
-            ConversationSupport(),
+            ConversationSupport(request_level=RequestLevel.BEST_EFFORT),
         ]
         self.llm = llm
 
@@ -41,8 +43,7 @@ class BasicChatAgent(Agent):
         # Build the Chat History that we'll provide as input to the action
         messages = build_chat_history(self.PROMPT, self.message_selector, context)
 
-        # call chat() with a hard-coded absence of tools
-        output_blocks = self.llm.generate(messages=messages, tools=[])
+        output_blocks = self.llm.generate(messages=messages, capabilities=self.capabilities)
 
         for block in output_blocks:
             block.set_chat_role(RoleTag.ASSISTANT)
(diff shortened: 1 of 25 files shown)


src/steamship/agents/functions_based.py
```diff
@@ -0,0 +1,127 @@
...
+                if tool is None:
+                    raise SteamshipError(
+                        f"LLM attempted to invoke tool {invocation.tool_name}, but {self.__class__.__name__} does not have a tool with that name."
+                    )
+                # TODO (PR): Could overload Action for this case to piggy back Invocations on it, but ehhhhhhhhh...
+                # TODO Block parse for input.  text is the default argument that we currently pass in for Tools.  As
+                #  part of a refactor to allow for other parameters, this would need to change.
+                future_action = Action(
+                    tool=tool.name, input=[Block(text=invocation.args.get("text"))], output=None
+                )
```
GitOnUp, line 65: This is a tad awkward at the moment without doing an args / tools refactor as well.

src/steamship/agents/functions_based.py
```diff
@@ -0,0 +1,127 @@
...
+
+        for block in output_blocks:
+            if block.mime_type == MimeTypes.STEAMSHIP_PLUGIN_FUNCTION_CALL_INVOCATION:
+                invocation = FunctionCallingSupport.FunctionCallInvocation.from_block(block)
+                tool = self.tools_map.get(invocation.tool_name)
+                if tool is None:
+                    raise SteamshipError(
+                        f"LLM attempted to invoke tool {invocation.tool_name}, but {self.__class__.__name__} does not have a tool with that name."
+                    )
+                # TODO (PR): Could overload Action for this case to piggy back Invocations on it, but ehhhhhhhhh...
```
GitOnUp, line 60: Callout is mainly for consistency with Action, here.

src/steamship/agents/utils.py
```diff
@@ -18,3 +21,49 @@ def with_llm(llm: LLM, context: Optional[AgentContext] = None) -> AgentContext:
 def get_llm(context: AgentContext, default: Optional[LLM] = None) -> Optional[LLM]:
     """Retrieves the LLM from the provided AgentContext (if it exists)."""
     return context.metadata.get(_LLM_KEY, default)
+
+
+def build_chat_history(
```
GitOnUp, line 26: Pulled generally applicable parts out of functions based agent.

src/steamship/agents/functions_based.py
```diff
@@ -0,0 +1,127 @@
...
+            return
+
+        tags = [
+            Tag(
+                kind=TagKind.ROLE,
+                name=RoleTag.FUNCTION,
+                value={TagValueKey.STRING_VALUE: action.tool},
+            ),
+            # TODO (PR): we're asserting capabilities support in next_action so the "name" tag is no longer needed for
+            #  backcompat as we won't be able to run against older versions anyway.
```
GitOnUp, line 118: Callout for removal

src/steamship/agents/utils.py
```diff
@@ -18,3 +21,49 @@ def with_llm(llm: LLM, context: Optional[AgentContext] = None) -> AgentContext:
...
+    default_system_message: str, message_selector: MessageSelector, context: AgentContext
+) -> List[Block]:
+    # system message should have already been created in context, but we double-check for safety
+    if context.chat_history.last_system_message:
```
GitOnUp, line 30: I'm also noticing while doing plugin implementations that we tend to not dedupe between this and "default_system_prompt" in configs, which seems like it could confuse some models.

src/steamship/agents/basic_chat.py
```diff
@@ -0,0 +1,49 @@
...
+            ConversationSupport(),
+        ]
+        self.llm = llm
+
+    def next_action(self, context: AgentContext) -> Action:
+        # Build the Chat History that we'll provide as input to the action
+        messages = build_chat_history(self.PROMPT, self.message_selector, context)
+
+        # call chat() with a hard-coded absense of tools
+        output_blocks = self.llm.generate(messages=messages, tools=[])
```
eob, line 45: This is so clean!!

src/steamship/agents/basic_chat.py
```diff
@@ -0,0 +1,49 @@
...
+            SystemPromptSupport(),
+            ConversationSupport(),
+        ]
+        self.llm = llm
+
+    def next_action(self, context: AgentContext) -> Action:
+        # Build the Chat History that we'll provide as input to the action
+        messages = build_chat_history(self.PROMPT, self.message_selector, context)
+
+        # call chat() with a hard-coded absense of tools
```
eob, line 44:
```suggestion
        # call chat() with a hard-coded absence of tools
```

eob (approved):
woohoooo!!!!!!!

This looks super clean to me. I'm not the authority @douglas-reid is on the function calling semantics though.

If there are tests that use the (to be deprecated) functional/functions_calling agent, maybe we could copy-paste them and replace that agent with this one?

And if not, we can just follow up with a unit testing PR where a few of us pile in and add some to feel out the new semantics.

🚀 🚀

src/steamship/agents/basic_chat.py
GitOnUp: Note to self to update this to not say `chat`

src/steamship/agents/functional/functions_based.py
```diff
@@ -10,7 +10,10 @@
 
 
 class FunctionsBasedAgent(ChatAgent):
-    """Selects actions for AgentService based on OpenAI Function style LLM Prompting."""
+    """Selects actions for AgentService based on OpenAI Function style LLM Prompting.
+
+    This class will be deprecated in the future, in favor of steamship.agents.functions_based.FunctionsBasedAgent.
```
douglas-reid, line 15: my suggestion here is not to add this until we actually deprecate.

src/steamship/agents/functions_based.py
```diff
@@ -0,0 +1,127 @@
...
+from steamship.data.tags.tag_utils import get_tag
+from steamship.plugin.capabilities import (
+    ConversationSupport,
+    FunctionCallingSupport,
+    SystemPromptSupport,
+)
+
+
+class FunctionsBasedAgent(Agent):
+    """Selects actions for AgentService based on a set of Tools."""
```
douglas-reid, line 17: from today's chat, I would add something like:

```
NOTE: Not ready for usage yet.  
```

src/steamship/agents/functions_based.py
```diff
@@ -0,0 +1,127 @@
...
+NOTE: Some functions return images, video, and audio files. These multimedia files will be represented in messages as
+UUIDs for Steamship Blocks. When responding directly to a user, you SHOULD print the Steamship Blocks for the images,
+video, or audio as follows: `Block(UUID for the block)`.
+
+Example response for a request that generated an image:
+Here is the image you requested: Block(288A2CA1-4753-4298-9716-53C1E42B726B).
+
+Only use the functions you have been provided with."""
+
+    def __init__(self, llm: SteamshipLLM, tools: List[Tool], **kwargs):
```
douglas-reid, line 30: should we use something like `warnings.warn("not ready", RuntimeWarning, ...)` or whatever is appropriate here, until we have tests and examples converted?

src/steamship/agents/functions_based.py
```diff
@@ -0,0 +1,127 @@
...
+                tool = self.tools_map.get(invocation.tool_name)
+                if tool is None:
+                    raise SteamshipError(
+                        f"LLM attempted to invoke tool {invocation.tool_name}, but {self.__class__.__name__} does not have a tool with that name."
+                    )
+                # TODO (PR): Could overload Action for this case to piggy back Invocations on it, but ehhhhhhhhh...
+                # TODO Block parse for input.  text is the default argument that we currently pass in for Tools.  As
+                #  part of a refactor to allow for other parameters, this would need to change.
+                future_action = Action(
+                    tool=tool.name, input=[Block(text=invocation.args.get("text"))], output=None
```
douglas-reid, line 64: is this assuming the argument is "text" ?  how does this deal with multi-media blocks that have args of "uuid" ?

src/steamship/agents/functions_based.py
```diff
@@ -0,0 +1,127 @@
...
+        self, invocation: FunctionCallingSupport.FunctionCallInvocation, context: AgentContext
+    ):
+        tags = [
+            Tag(
+                kind=TagKind.CHAT,
+                name=ChatTag.ROLE,
+                value={TagValueKey.STRING_VALUE: RoleTag.ASSISTANT},
+            ),
+            Tag(kind=TagKind.FUNCTION_SELECTION, name=invocation.tool_name),
+            Tag(
```
douglas-reid, line 97: we can drop this tag now, as `request-id` tagging should be automatic in next release of Engine.

src/steamship/agents/functions_based.py
```diff
@@ -0,0 +1,127 @@
...
+
+        tags = [
+            Tag(
+                kind=TagKind.ROLE,
+                name=RoleTag.FUNCTION,
+                value={TagValueKey.STRING_VALUE: action.tool},
+            ),
+            # TODO (PR): we're asserting capabilities support in next_action so the "name" tag is no longer needed for
+            #  backcompat as we won't be able to run against older versions anyway.
+            Tag(
```
douglas-reid, line 119: same as above.

src/steamship/agents/basic_chat.py
```diff
@@ -0,0 +1,49 @@
...
+            ConversationSupport(),
+        ]
+        self.llm = llm
+
+    def next_action(self, context: AgentContext) -> Action:
+        # Build the Chat History that we'll provide as input to the action
+        messages = build_chat_history(self.PROMPT, self.message_selector, context)
+
+        # call chat() with a hard-coded absence of tools
+        output_blocks = self.llm.generate(messages=messages, tools=[])
```
douglas-reid, line 45: @GitOnUp question while you are also looking at `gpt-4` plugin:  when there are no functions in the call, can we update our call to OpenAI to turn off function-selection? I think by default we use `auto` as the setting, but with the tags / your capability work, we should be able to decide up-front whether or not we even want functions as a possibility in the answer, right?

GitOnUp: The way that I'd default to doing that is to only turning them on if you get the capability defining functions, yes, and moreso that you'd _disable_ it in the capability if you didn't want them (though the distinction in that specific capability may be moot since if you don't provide functions to call it doesn't make sense to try calling any)

src/steamship/agents/functions_based.py
GitOnUp: What do you think about just `_` prefixing these?

douglas-reid (approved):
this all seems fine. thanks for adding the prefixes. will be cool when we can transition. thanks for your continued work here.

PR #585, opened 2023-10-13
No File.refresh() during File.append_block

dkolas:
The current code, upon appending a `Block` to a `File`, immediately calls `File.refresh()` to update the state of the `File` object.  This PR adds the newly created `Block` directly to the `File` without making a second API call.

The risk / downside to this is that any blocks that were created on the File in between are now not loaded as part of this call.


douglas-reid (dismissed):
🤏 🔨 ▶️  🚤

